PR #300, opened 2021-02-02
Keep chart state in single model view

romanlutz:
This came up in a recent bug bash. The single model view charts always reset to the default (`outcomePlot`) rather than keeping state. This is undesirable if all you care about is the `performancePlot`, so this change adds state to keep track of the selection.

Additionally, this PR removes the `expandAttributes` property that wasn't used anymore, and renames `displayPlotKey` for consistency.

### DIFF
--- a/libs/fairness/src/lib/v2/WizardReport.tsx
+++ b/libs/fairness/src/lib/v2/WizardReport.tsx
@@ -32,7 +32,7 @@ export interface IState {
   fairnessKey?: string;
   performanceKey?: string;
   showModalHelp?: boolean;
-  expandAttributes: boolean;
+  chartKey?: string;
 }
 
 export interface IReportProps extends IModelComparisonProps {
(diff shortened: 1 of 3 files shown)


riedgar-ms (approved):

imatiach-msft (approved):
LGTM

libs/fairness/src/lib/v2/WizardReport.tsx
```diff
@@ -208,6 +207,8 @@ export class WizardReport extends React.PureComponent<IReportProps, IState> {
                 metrics={this.state.metrics}
                 fairnessPickerProps={this.props.fairnessPickerProps}
                 performancePickerProps={this.props.performancePickerProps}
+                chartKey={this.state.chartKey}
+                onUpdateChartKey={this.updateChartKey}
```
xuke444, line 211: seems it is unnecessary to keep another state here.
Can we simply pass the prop? instead of another copy of state?

PR #301, opened 2021-02-02
fix error cohort stats in big data case

imatiach-msft:
fix error cohort stats in big data case
- when we have large data (like categorical notebook example), we should use the info from python backend for the legends instead of the downsampled explanation data


imatiach-msft:
closing this as duplicate of #303 until we resolve issues with my forks

PR #310, opened 2021-02-05
convert flask comms class to function

romanlutz:
As recommended by @xuke444 


apps/widget/src/app/FlaskCommunication.ts
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import axios from "axios";
+
+import { config } from "./config";
+
+export async function callFlaskService<TRequest, TResponse>(
```
xuke444, line 8: Do we want to match file name with function name inside?
That should be easier to maintain.

xuke444 (approved):

imatiach-msft (approved):

PR #316, opened 2021-02-06
Remove circular dependency

romanlutz:
Removing this since it makes debugging harder due to the verbose warnings.


imatiach-msft (approved):
LGTM!

PR #322, opened 2021-02-08
fix categorical labels handling in what if panel

imatiach-msft:
Fixes issue:
https://github.com/microsoft/responsible-ai-widgets/issues/314

By passing indexes instead of raw values for true_y and predicted_y columns when needed.

The what if panel now looks like:

image
image


romanlutz (approved):

romanlutz:
Would be nice to have tests for this. :-)

PR #324, opened 2021-02-08
Move dataset utilities to core-ui

romanlutz:
This PR moves JointDataset, DatasetUtils, and a few explanation interfaces interfaces that they depend on to core-ui to prevent error-analysis from depending on internals of interpret. This is the first of several such PRs to move common functionality into a common location to make dependencies between components easier to handle. Note that no logic changed here. Files were simply moved and references to those files updated. [There is one exception, a typo fix in a comment]


libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/TabularDataView/TabularDataView.tsx
```diff
@@ -1,7 +1,12 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import { JointDataset } from "@responsible-ai/interpret";
+import {
+  JointDataset,
+  constructRows,
+  constructCols,
+  rowsFromCustomPoints
```
imatiach-msft, line 8: do you really need to move these to core ui?
 constructRows,
  constructCols,
  rowsFromCustomPoints

imatiach-msft (approved):

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/TabularDataView/TabularDataView.tsx
romanlutz: Perhaps not these particular ones, but to be honest I'm struggling to draw the line between "This is a utility function that's more generally useful" and "This is a utility function that's very very specific to the particular project" (in this case error-analysis). From a cursory glance it seems like JointDataset and DatasetUtils were all dataset utilities so I just moved it all together under core-ui. For a second I considered even putting them in the same file, but didn't want to change anything in such a large PR.

If you think it's better off living in error-analysis that's fine by me. I can move this file back.

imatiach-msft: uh, I thought these were only used by error analysis just for the instance view?  is this used in interpret as well?

imatiach-msft: Maybe the file name needs to be changed to something like InstanceViewUtils or something like that.  These functions aren't used in interpret dashboard right now.

imatiach-msft: "This is a utility function that's more generally useful"
usually you would know if it's used in multiple places across multiple projects.  These functions are very error analysis specific right now.

romanlutz: Great. I'll keep the general ones (JointDataset) in core-ui then, and move the DatasetUtils back to error-analysis

PR #332, opened 2021-02-11
fix error message for global explanation

xuke444:

### DIFF
--- a/libs/interpret/src/lib/MLIDashboard/ValidateProperties.ts
+++ b/libs/interpret/src/lib/MLIDashboard/ValidateProperties.ts
@@ -155,7 +155,7 @@ export class ValidateProperties {
         !this.validate2D(
           globalExp,
           [this.featureLength, this.classLength],
-          localization.Interpret.ValidationErrors.localFeatureImportance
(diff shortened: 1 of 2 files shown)


romanlutz (approved):

libs/interpret/src/lib/MLIDashboard/ValidateProperties.ts
```diff
@@ -155,7 +155,7 @@ export class ValidateProperties {
         !this.validate2D(
           globalExp,
           [this.featureLength, this.classLength],
-          localization.Interpret.ValidationErrors.localFeatureImportance
+          localization.Interpret.ValidationErrors.globalFeatureImportance
         )
```
gaugup, line 159: not sure I understand this, but isn't the function validating local importance? Then change the error from local to global?

imatiach-msft (approved):

libs/interpret/src/lib/MLIDashboard/ValidateProperties.ts
xuke444: it is validation global(line 156)
but it was using error message for local importance.
So I correct it by using global message

PR #336, opened 2021-02-13
add notebook tests

romanlutz:
Adds validation for all notebooks that's run upon each PR submission


imatiach-msft (changes requested):
don't merge the downloaded CSV and .zip file changes, otherwise looks good

.github/workflows/CI-notebook.yml
```diff
@@ -0,0 +1,36 @@
...
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  ci-notebook:
+    strategy:
+      matrix:
+        operatingSystem: [ubuntu-latest, windows-latest]
+        pythonVersion: [3.6, 3.7, 3.8]
```
imatiach-msft, line 14: I think you may need to limit this to 3.6, 3.7 since shap 0.34.0 which we depend on only supports 3.5-3.7.  We can't drop 3.5 support and move to latest shap because we are required to support it still.

romanlutz: Who's required to support 3.5 with raiwidgets?

romanlutz: If 3.8 really isn't supported we should remove it from setup.py. We're currently claiming that this works

imatiach-msft (approved):

imatiach-msft (approved):

PR #338, opened 2021-02-16
improve node text color contrast on darker background

imatiach-msft:
- changes node text color to white on dark color fill for better contrast (resolves https://github.com/microsoft/responsible-ai-widgets/issues/222, https://github.com/microsoft/responsible-ai-widgets/issues/173)
image

### DIFF
--- a/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/MatrixArea/MatrixArea.tsx
+++ b/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/MatrixArea/MatrixArea.tsx
@@ -9,7 +9,6 @@ import {
   JointDataset
 } from "@responsible-ai/core-ui";
 import { localization } from "@responsible-ai/localization";
-import { lab as Lab } from "d3-color";
 import { interpolateHcl as d3interpolateHcl } from "d3-interpolate";
 import { scaleLinear as d3scaleLinear } from "d3-scale";
 import {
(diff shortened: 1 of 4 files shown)


libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/MatrixArea/MatrixArea.tsx
```diff
@@ -658,11 +657,6 @@ export class MatrixArea extends React.PureComponent<
   }
 
   private textColorForBackground(colorStr: string): string {
-    return this.isColorDark(colorStr) ? "white" : "rgba(0,0,0,0.8)";
-  }
-
-  private isColorDark(colorStr: string): boolean {
-    const val = Lab(colorStr).l;
-    return val <= 65;
+    return isColorDark(colorStr) ? "white" : "rgba(0,0,0,0.8)";
```
romanlutz, line 660: It would be nice if we didn't hard code colors because this will come back to bite us when supporting dark mode. Just saying... I don't particularly care if you address it now or later, but this will likely come up again.

romanlutz (approved):

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/MatrixArea/MatrixArea.tsx
imatiach-msft: good point, will move it to constants in next PR

PR #344, opened 2021-02-17
Set up API docs through GitHub Pages

romanlutz:
This includes the basic docs setup & a GitHub workflow to deploy it to GitHub Pages after PRs are merged.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -74,13 +74,11 @@ html_theme = 'pydata_sphinx_theme'
 # further.  For a list of options available for each theme, see the
 # documentation.
 html_theme_options = {
-    # TODO: fork the pydata-sphinx-theme to integrate these with logo
     "external_links": [
         # {"name": "StackOverflow",
-        #  "url": "https://stackoverflow.com/questions/tagged/fairlearn"}
+        #  "url": "https://stackoverflow.com/questions/tagged/raiwidgets"}
     ],
     "github_url": "https://github.com/microsoft/responsible-ai-widgets",
-    # "twitter_url": "https://twitter.com/fairlearn" TODO: start using this
     "show_prev_next": False
 }
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,3 +1,6 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+#
 # Configuration file for the Sphinx documentation builder.
 #
 # This file only contains a selection of the most common options. For a full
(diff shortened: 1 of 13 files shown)


docs/source/conf.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'Responsible AI Widgets'
+copyright = 'Microsoft Corporation'
+author = 'Roman Lutz, Ilya Matiach, Ke Xu'
+
+# The full version, including alpha/beta/rc tags
+release = 'v0.2.0'
```
imatiach-msft, line 30: this is another place where we need to keep track of version and could reuse the same constant.  actually this should be 0.2.1 now.

docs/source/conf.py
```diff
@@ -0,0 +1,179 @@
...
+html_theme = 'pydata_sphinx_theme'
+
+# Theme options are theme-specific and customize the look and feel of a theme
+# further.  For a list of options available for each theme, see the
+# documentation.
+html_theme_options = {
+    # TODO: fork the pydata-sphinx-theme to integrate these with logo
+    "external_links": [
+        # {"name": "StackOverflow",
+        #  "url": "https://stackoverflow.com/questions/tagged/fairlearn"}
```
imatiach-msft, line 80: remove fairlearn commented out stuff here?

raiwidgets/raiwidgets/explanation_dashboard.py
```diff
@@ -45,7 +45,6 @@ class and for the regression case a method of predict()
     def __init__(self, explanation, model=None, dataset=None,
                  true_y=None, classes=None, features=None,
                  public_ip=None, port=None, locale=None):
-        """Initialize the fairness Dashboard."""
```
imatiach-msft, line 48: are you sure we should remove such comments?  maybe just update to say explanation dashboard?

imatiach-msft (approved):

docs/source/conf.py
romanlutz: Ah, here I should be able to just reference the other one.

docs/source/conf.py
romanlutz: I want to eventually add StackOverflow, but currently there's no tag for raiwidgets and I don't have permissions to add such a tag on StackOverflow :-) I'll change the tag to raiwidgets in the text, though.

raiwidgets/raiwidgets/explanation_dashboard.py
romanlutz: Do they add anything? The method is called `__init__` in a class called `ExplanationDashboard`. I'll add it back, but I'm typically in favor of useful comments only.

PR #345, opened 2021-02-18
fix categrical feature in ice plot

xuke444:


libs/interpret/src/lib/MLIDashboard/Controls/MultiICEPlot/MultiICEPlot.tsx
```diff
@@ -549,7 +549,7 @@ export class MultiICEPlot extends React.PureComponent<
 
   private buildRangeView(featureKey: string): IRangeView | undefined {
     const summary = this.props.jointDataset.metaDict[featureKey];
```
imatiach-msft, line 551: I thought there is another issue where we are not passing the categorical mapping for features from python side to UI side?

imatiach-msft (approved):

PR #351, opened 2021-02-19
one version for js and python

xuke444:


raiwidgets/raiwidgets/__version__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+version = "0.3.0"
```
romanlutz, line 4: If we release v0.3.0 we should really have a description of what's new. Releasing for the sake of automating versioning isn't great :-)

xuke444: I am not planning to release yet.
Change to 3.0 because that is the next version good for all components.
Once we are ready to release we can update changes.md

romanlutz: Clarified this in person - no python release planned for right now

romanlutz (approved):

PR #352, opened 2021-02-19
fix indexing issue in heatmap

imatiach-msft:
fix indexing being out of order in certain cases (eg all cells selected in heatmap) - this issue seems similar to the indexing problem in https://github.com/microsoft/responsible-ai-widgets/pull/333 , except there we needed to make a copy of jointdataset because order on jointdataset was broken by aggregate/dataset tabs.  The jointdataset order should remain sorted by index always however.  In this case the order of the data in the cohort's filteredData is broken (not the jointdataset shared by all), but actually that is expected to happen in cases like aggregate/dataset tabs, so the best solution is to just re-sort the cohort filtered data again when it's expected to be in index order.

### DIFF
--- a/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InspectionView/InspectionView.tsx
+++ b/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InspectionView/InspectionView.tsx
@@ -96,6 +96,7 @@ export class InspectionView extends React.PureComponent<
       sortArray: [],
       sortingSeriesIndex: undefined
     };
+    this.props.selectedCohort.cohort.sort();
     const cohortData = this.props.selectedCohort.cohort.filteredData;
     const numRows: number = cohortData.length;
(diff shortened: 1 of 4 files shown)


xuke444 (approved):

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InspectionView/InspectionView.tsx
```diff
@@ -96,6 +96,7 @@ export class InspectionView extends React.PureComponent<
       sortArray: [],
       sortingSeriesIndex: undefined
     };
+    this.props.selectedCohort.cohort.sort();
```
xuke444, line 99: this may have side effect
if props used in sibling components

imatiach-msft: ok this may need to be re-examined in the other components

PR #358, opened 2021-02-22
add num selected text under local explanation radio buttons

imatiach-msft:
- add number of selected rows text under local explanation radio buttons
resolves issue: https://github.com/microsoft/responsible-ai-widgets/issues/251
image

### DIFF
--- a/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InstanceView/InstanceView.styles.ts
+++ b/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InstanceView/InstanceView.styles.ts
@@ -12,6 +12,8 @@ export interface IInstanceViewStyles {
   choiceGroupContainerStyle: IStyle;
   choiceItemRootStyle: IStyle;
   page: IStyle;
+  selectedTextStyle: IStyle;
+  stackItemsStyle: IStyle;
 }
 
 export const InstanceViewStyles: () => IProcessedStyleSet<
@@ -30,6 +32,10 @@ export const InstanceViewStyles: () => IProcessedStyleSet<
     page: {
       backgroundColor: theme.semanticColors.bodyBackground,
       color: theme.semanticColors.bodyText
+    },
+    selectedTextStyle: { color: getTheme().palette.neutralPrimaryAlt },
+    stackItemsStyle: {
+      padding: "0px 0px 0px 26px"
     }
   });
 };
--- a/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InstanceView/InstanceView.tsx
+++ b/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InstanceView/InstanceView.tsx
@@ -13,7 +13,8 @@ import {
   ITheme,
   PrimaryButton,
   ChoiceGroup,
-  Stack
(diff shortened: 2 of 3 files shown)


libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InstanceView/InstanceView.styles.ts
```diff
@@ -30,6 +32,10 @@ export const InstanceViewStyles: () => IProcessedStyleSet<
     page: {
       backgroundColor: theme.semanticColors.bodyBackground,
       color: theme.semanticColors.bodyText
+    },
+    selectedTextStyle: { color: getTheme().palette.neutralPrimaryAlt },
+    stackItemsStyle: {
+      padding: "0px 0px 0px 26px"
```
xuke444, line 38: padding: 0

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InstanceView/InstanceView.tsx
```diff
@@ -357,4 +367,52 @@ export class InstanceView extends React.Component<
...
+  private onRenderLabel = (type: SelectionType) => (
+    p: IChoiceGroupOption | undefined
+  ) => {
+    const classNames = InstanceViewStyles();
+    let selectionText = "";
+    switch (type) {
+      case SelectionType.AllSelectionType:
+        selectionText =
+          this.state.selectionDetails.selectedAllSelectedIndexes.length +
+          " selected";
```
xuke444, line 380: we need localize all the string.
since the sequence in each language will be different, it is better to use place holder to localize whole string including number.

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InstanceView/InstanceView.tsx
```diff
@@ -357,4 +367,52 @@ export class InstanceView extends React.Component<
...
+        break;
+    }
+    const stackItemStyles: IStackItemStyles = {
+      root: classNames.stackItemsStyle
+    };
+    const textStyle = {
+      root: classNames.selectedTextStyle
+    };
+    return (
+      <div style={{ display: "flex", flexDirection: "column" }}>
```
xuke444, line 402: this can be replaced by a horizontal stack

xuke444: seems not needed at all though.

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InstanceView/InstanceView.tsx
imatiach-msft: done, localized

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InstanceView/InstanceView.tsx
imatiach-msft: removed

xuke444 (approved):

PR #363, opened 2021-02-26
auto publish python

xuke444:


.vscode/settings.json
```diff
@@ -55,14 +55,19 @@
     "Plotly",
     "Unmount",
     "automl",
+    "bdist",
```
imatiach-msft, line 58: what is this file used for?

imatiach-msft (approved):

.vscode/settings.json
xuke444: for spell checking, in extension.json, we have an extension for spell checking

PR #368, opened 2021-03-03
Xuke/fixglobalimportance

xuke444:
1. use global names for global feature importance
2. missingParametersPlaceholder styles are copied all over the places, I replace it by a dedicated component 
3. precompute global importance are transformed and copied in many places yet only used in one component.  Instead I use the raw data from context.
4. private field in react is totally wrong, should always use state.
5. having raw html tag "<br>" in string is a really bad practice, since tool tip support string array for line wrap we should use string array instead of raw tag.

### DIFF
--- a/libs/error-analysis/src/lib/ErrorAnalysisDashboard/ErrorAnalysisDashboard.tsx
+++ b/libs/error-analysis/src/lib/ErrorAnalysisDashboard/ErrorAnalysisDashboard.tsx
@@ -624,10 +624,6 @@ export class ErrorAnalysisDashboard extends React.PureComponent<
                     <GlobalExplanationTab
                       jointDataset={this.state.jointDataset}
                       metadata={this.state.modelMetadata}
-                      globalImportance={this.state.globalImportance}
-                      isGlobalDerivedFromLocal={
-                        this.state.isGlobalImportanceDerivedFromLocal
-                      }
                       cohorts={this.state.cohorts.map(
                         (errorCohort) => errorCohort.cohort
                       )}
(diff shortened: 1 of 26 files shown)


libs/error-analysis/src/lib/ErrorAnalysisDashboard/ErrorAnalysisDashboard.tsx
```diff
@@ -624,10 +624,6 @@ export class ErrorAnalysisDashboard extends React.PureComponent<
                     <GlobalExplanationTab
                       jointDataset={this.state.jointDataset}
                       metadata={this.state.modelMetadata}
-                      globalImportance={this.state.globalImportance}
```
imatiach-msft, line 627: can this.state.globalImportance just be removed now?  I'm a bit confused how GlobalExplanationTab gets this info now.

libs/interpret/src/lib/MLIDashboard/Controls/GlobalExplanationTab/GlobalExplanationTab.tsx
```diff
@@ -134,23 +133,7 @@ export class GlobalExplanationTab extends React.PureComponent<
...
-      return (
-        <div className={classNames.missingParametersPlaceholder}>
-          <div className={classNames.missingParametersPlaceholderSpacer}>
-            <Text variant="large" className={classNames.faintText}>
-              {localization.Interpret.GlobalTab.missingParameters}
-            </Text>
-          </div>
-        </div>
-      );
+      return <GlobalOnlyChart />;
```
imatiach-msft, line 136: So for the global only chart, is it just showing top most important features for all data cohort, and it can't show feature importances for individual cohorts?  Interesting.

imatiach-msft (approved):
LGTM

libs/interpret/src/lib/MLIDashboard/Controls/GlobalExplanationTab/GlobalExplanationTab.tsx
xuke444: right, we cannot filter data.

PR #370, opened 2021-03-04
Add fairness to model assessment dashboard

romanlutz:
This PR contains
- fairness tab added 
image

- unified inputs to avoid duplication: `IDataset` and `IModelExplanationData` provide the foundation for all inputs that are passed to the individual components. Over time I'd like to use these inputs for existing dashboards as well, but that's too much to change in a single PR. That's why you find some // TODOs all over the place.
- renamed `FairLearnSetting` to `FairnessSetting`
- moved test case specific functions into a general shared place to be used by both error analysis and model assessment tests (`error-analysis/utils.ts`); same for fairness utils that are now in `fairness/utils.ts`
- updated census dataset in model assessment folder to include fairness data
- moved the following to `core-ui` since they're more general than any individual component:
  - `IFairnessBaseData`
  - `IRunTimeFairnessData`
  - `IPreComputedFairnessData`
  - `PredictionType`
  - `PredictionTypes`
  - `IPreComputedData`
  - `IDatasetSummary`
  - `IMetricRequest`
  - `IMetricResponse`

This PR does not yet contain:
- tests
- moving tabs bar to the left side (vertically rather than horizontally)
- generally some parts of how the explanation dashboard is structured are preferable, so in the next PRs you'll see some of the existing parts disappear in favor of that (as requested by @xuke444 )
- persisted state for fairness tab to be able to navigate to another tab and back and still see the previous selection (this will require a way to navigate all the way back which doesn't exist yet, plus state for 3-4 selections)


xuke444 (approved):

apps/dashboard/src/fairness/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export const messages = {
+  LocalExpAndTestReq: [{ displayText: "LocalExpAndTestReq" }],
+  LocalOrGlobalAndTestReq: [{ displayText: "LocalOrGlobalAndTestReq" }],
+  PredictorReq: [{ displayText: "PredictorReq" }],
+  TestReq: [{ displayText: "TestReq" }]
+};
+
+// calculateMetrics can be used with local flask server for debugging
+/*export function calculateMetrics(postData: IMetricRequest): Promise<IMetricResponse> {
```
imatiach-msft, line 38: remove commented out code?

imatiach-msft (approved):

apps/dashboard/src/fairness/utils.ts
romanlutz: This isn't new in this PR (it just moved files) and actually commented out on purpose. It can be used if one wants to debug with flask using Jupyter. That's why it uses localhost

PR #376, opened 2021-03-10
Set up tabs on the left side for model assessment dashboard

romanlutz:
Changes in this PR include

- Tabs are now on the left using a `Nav`. Before, they were in a `Pivot` stretching horizontally across the page.
- small changes to components to make them render over the full page width for model-assessment:
  - removed a couple of `className`s in Error Analysis 
  - added `inherit` for width in `InstanceView`
  - fairness: width of 100%
- move `ErrorCohort`, `CohortStats`, `CohortInfo`, `CohortList`, `EditCohort`, `PredictionPath`, `SaveCohort`, and `ShiftCohort` to `core-ui` into the `Cohort`-specific directory
- move shared context building code from `error-analysis` and `model-assessment` to `core-ui`
- Create `CohortBasedComponent` as an abstract base class for `model-assessment`, but in the future also for `error-analysis` and `interpret`. The main reason why it's not immediately usable by them is that they don't yet use `IDataset`.
- Add `ErrorAnalysisViewTab` as a wrapper around `ErrorAnalysisView` that encapsulates the tree or heat map and the tree/heat map-specific controls like the dropdown to select the map and the feature list button.
- Add `ModelAssessmentContext` (to be used in the future) 

image

Follow-up PRs will include:
- model statistics tab
- what if tab scaffold
- actually using the new context!
- moving data explorer into a separate package
- and several others...


libs/core-ui/src/lib/Cohort/CohortBasedComponent.ts
```diff
@@ -0,0 +1,77 @@
...
+  cohorts: ErrorCohort[];
+  selectedCohort: ErrorCohort;
+  jointDataset: JointDataset;
+}
+
+export abstract class CohortBasedComponent<
+  TProps extends ICohortBasedComponentProps,
+  TState extends ICohortBasedComponentState
+> extends React.PureComponent<TProps, TState> {
+  public updateSelectedCohort(
```
imatiach-msft, line 29: any reason this was moved here?

libs/model-assessment/src/lib/ModelAssessmentDashboard/ModelAssessmentDashboard.tsx
```diff
@@ -968,4 +488,34 @@ export class ModelAssessmentDashboard extends React.PureComponent<
...
+    _: React.FormEvent<HTMLDivElement>,
+    item?: IDropdownOption
+  ): void => {
+    if (item) {
+      if (item.key === ErrorAnalysisOptions.HeatMap) {
+        // Note comparison above is actually string comparison (key is string), we have to set the enum
+        const selectedOptionHeatMap = ErrorAnalysisOptions.HeatMap;
+        this.setErrorDetector(selectedOptionHeatMap);
+      } else {
+        // Note comparison above is actually string comparison (key is string), we have to set the enum
```
imatiach-msft, line 502: maybe just have one comment above if instead of duplicates

imatiach-msft (approved):

imatiach-msft:
can you refactor the ModelAssessmentDashboard.tsx code to keep it at 300 lines max? can be in separate PR.

libs/core-ui/src/lib/Cohort/CohortBasedComponent.ts
romanlutz: Yes! I noticed that there are lots of cohort-related operations underlying our dashboards (model-assessment and error-analysis, but also interpret). There's a whole bunch of things that they could just have in common. For now, this is only from model-assessment, because the `IDataset` is only used by model-assessment (because I haven't migrated others to use it yet). My goal was to put everything that's this general into core-ui so that everyone can build on it with the same foundation. That will require a bit more work in the future, of course, but this is a first step. What do you think?

libs/model-assessment/src/lib/ModelAssessmentDashboard/ModelAssessmentDashboard.tsx
```diff
@@ -968,4 +488,34 @@ export class ModelAssessmentDashboard extends React.PureComponent<
...
+      this.setState({
+        mapShiftErrorAnalysisOption: key,
+        openMapShift: true
+      });
+    } else {
+      this.setState({
+        errorAnalysisOption: key
+      });
+    }
+  };
```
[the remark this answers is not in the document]
romanlutz: And there I was thinking 500 is great :-) It was at 1300 before I starting moving things into their own components. I'd probably prefer doing it in a separate PR since this one already got so large from my first attempts at moving things into their own components. The use of the ModelAssessmentContext should help me trim it more as well.

libs/core-ui/src/lib/Cohort/Cohort.ts
```diff
@@ -10,6 +10,12 @@ import {
 import { JointDataset } from "../util/JointDataset";
 import { ModelExplanationUtils } from "../util/ModelExplanationUtils";
 
+export enum ErrorDetectorCohortSource {
```
xuke444, line 13: since we are moving cohort to shared place, we should get rid of word "error"

libs/core-ui/src/lib/Interfaces/ExplanationInterfaces.ts
```diff
@@ -33,8 +34,8 @@ export interface ISerializedExplanationData {
 }
 
 export interface IModelInformation {
```
xuke444, line 36: it might be better to move this to the other model interface file, since this related to model

libs/core-ui/src/lib/Cohort/Cohort.ts
romanlutz: Just `CohortSource` should be sufficient.

More generally, there is `Cohort` in interpret and `ErrorCohort` which is used in error-analysis and model-assessment. That won't be changeable easily.

libs/core-ui/src/lib/Interfaces/ExplanationInterfaces.ts
romanlutz: This is one of several interfaces I want to completely get rid of. See the comment below this interface and at the top of the file. IDataset and IModelExplanationData should have everything we need. It's just a matter of migrating to that set of cleaner interfaces.

xuke444 (approved):

PR #399, opened 2021-03-15
improve the features list to use DetailsList

imatiach-msft:
- improve the features list to use DetailsList for selecting features to retrain the tree map
image
- improve the description in the features list to include information about feature importances

### DIFF
--- a/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/FeatureList/FeatureList.tsx
+++ b/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/FeatureList/FeatureList.tsx
@@ -3,8 +3,13 @@
 
 import { localization } from "@responsible-ai/localization";
 import {
+  ConstrainMode,
+  DetailsList,
+  DetailsListLayoutMode,
   getTheme,
+  MarqueeSelection,
   PrimaryButton,
+  IColumn,
   IFocusTrapZoneProps,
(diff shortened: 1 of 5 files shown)


libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/FeatureList/FeatureList.tsx
```diff
@@ -65,39 +77,76 @@ export class FeatureList extends React.Component<
   IFeatureListState
 > {
   private layerHostId: string;
+  private _selection: Selection;
+  private _theme: ITheme;
```
romanlutz, line 81: Any reason why were storing this on the object? I'm not a React expert, but would this work if we change the theme halfway through? Or would it still keep the initial theme?

Ideally we should use global context for things like theme, and in the meanwhile `getTheme()` is just fine. @xuke444 may have thoughts.

romanlutz (approved):

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/FeatureList/FeatureList.tsx
imatiach-msft: good point!  I put it inside the render function instead.

imatiach-msft:
close reopen for CI

PR #401, opened 2021-03-17
added tooltips to legend

imatiach-msft:
improve information in tree view and matrix filter legends:
- add tooltips to matrix filter
image
image
image

- add tooltips to tree view, which also involved making the legend html positioned on top of SVG to make tooltips consistent
image
image

- change header text in tree view to be expandable
image

### DIFF
--- a/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InfoCallout/InfoCallout.tsx
+++ b/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InfoCallout/InfoCallout.tsx
@@ -1,0 +1,70 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { Callout, IconButton } from "office-ui-fabric-react";
+import React from "react";
+
+import { infoCalloutStyles } from "./InfoCallout.styles";
+
+export interface IInfoCalloutProps {
+  iconId: string;
+  infoText: string;
+  title: string;
+}
+
+export interface IInfoCalloutState {
+  isCalloutVisible: boolean;
+}
+
+export class InfoCallout extends React.Component<
+  IInfoCalloutProps,
+  IInfoCalloutState
+> {
+  public constructor(props: IInfoCalloutProps) {
+    super(props);
+    this.state = {
+      isCalloutVisible: false
+    };
+  }
+  public render(): React.ReactNode {
+    const classNames = infoCalloutStyles();
+    return (
+      <span>
+        <IconButton
+          id={this.props.iconId}
+          iconProps={{ iconName: "Info" }}
+          title={this.props.title}
+          onClick={this.onIconClick}
+          styles={{
+            root: {
+              color: "rgb(0, 120, 212)",
+              height: "24px",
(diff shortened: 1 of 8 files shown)


libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InfoCallout/InfoCallout.tsx
```diff
@@ -0,0 +1,72 @@
...
+        />
+        {this.state.isCalloutVisible && (
+          <Callout
+            target={"#" + this.props.iconId}
+            setInitialFocus={true}
+            onDismiss={this.onDismiss}
+            role="alertdialog"
+          >
+            <div className={classNames.calloutInfo}>
+              <div>
```
romanlutz, line 55: why two `div`s? The inner one doesn't seem to add anything. What am I missing?

libs/localization/src/lib/en.json
```diff
@@ -680,7 +687,13 @@
       "cohortName": "Cohort name"
     },
     "TreeView": {
-      "treeDescription": "To find nodes in the tree map with the most errors, look for higher values and fuller circles."
+      "treeDescription": "The tree visualization uses the mutual information between each feature and the error to best separate error instances from success instances hierarchically in the data.",
+      "treeDescriptionExpandedPar1": "This simplifies the process of discovering and highlighting common failure patterns.",
+      "treeDescriptionExpandedPar2": "To ",
```
romanlutz, line 692: I don't think this will work. We can't expect the sentence structure to align with this in all languages. Is it a huge issue if there no bold text?

libs/core-ui/src/lib/components/ExpandableText.tsx
```diff
@@ -6,7 +6,7 @@ import { Text, Link, ITextProps, Icon } from "office-ui-fabric-react";
 import React from "react";
 
 export interface IExpandableTextProps extends ITextProps {
-  expandedText?: string;
+  expandedText?: string | React.ReactNode;
```
romanlutz, line 9: Hmm, that's an odd type... I think this is all related to bold and normal text being mixed. Perhaps something like this is a more elegant solution? https://stackoverflow.com/questions/34778181/styling-text-in-localization-files-bold-italic-etc

libs/localization/src/lib/en.json
imatiach-msft: talked to PMs, will remove it

libs/core-ui/src/lib/components/ExpandableText.tsx
imatiach-msft: yes it was for above, I just removed it

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/InfoCallout/InfoCallout.tsx
imatiach-msft: Oh, this was used when I had a button in it, but I didn't remove the extra div when I removed the button.  Removed it now, thanks!

romanlutz (approved):

PR #420, opened 2021-03-26
refactor error analysis python code into separate common package

imatiach-msft:


xuke444 (changes requested):
Do you mind hold on merging?
Waiting for pm to confirm if they want one package or separate packages or ui and model assessment sdk.

erroranalysis/erroranalysis/version.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+name = 'erroranalysis'
```
xuke444, line 4: do you mind change this to match raiwidget?
So all packages share same version for better control

imatiach-msft: discussed, this will be done for raianalysis

xuke444 (approved):

PR #427, opened 2021-03-30
initial shell implementation of the raitools package

imatiach-msft:
The Responsible AI Tools SDK enables users to analyze their machine learning models in one API. Users will be able to analyze errors, explain the most important features, validate fairness, compute counterfactuals and run causal analysis in a single API.

Highlights of the package include:

- `fairness.add()` allows users to run Fairlearn to assess model fairness
- `explainer.add()` allows users to explain their model
- `counterfactuals.add()` allows users to compute counterfactuals
- `error_analysis.add()` allows users to run error analysis
- `causal.add()` allows users to run causal analysis

### DIFF
--- a/raitools/README.md
+++ b/raitools/README.md
@@ -1,0 +1,16 @@
+# Responsible AI Tools SDK for Python
+
+### This package has been tested with Python 3.6, 3.7 and 3.8
+
+The Responsible AI Tools SDK enables users to analyze their machine learning models in one API. Users will be able to analyze errors, explain the most important features, validate fairness, compute counterfactuals and run causal analysis using a single API.
+
+Highlights of the package include:
+
+- `fairness.add()` runs Fairlearn to assess model fairness
+- `explainer.add()` explains the model
+- `counterfactuals.add()` computes counterfactuals
+- `error_analysis.add()` runs error analysis
+- `causal.add()` runs causal analysis
+
+The source code can be found here:
+https://github.com/microsoft/responsible-ai-widgets
--- a/raitools/raitools/_managers/base_manager.py
+++ b/raitools/raitools/_managers/base_manager.py
@@ -1,0 +1,64 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+"""Defines the base class for managers."""
+
+from abc import ABC, abstractmethod
+
+
+class BaseManager(ABC):
+    """The base class for managers."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialize the BaseManager."""
+        super(BaseManager, self).__init__(*args, **kwargs)
+
+    @abstractmethod
+    def add(self):
+        """Abstract method to add a computation to the manager."""
+
+    @abstractmethod
+    def compute(self):
+        """Abstract method to compute the new work in the add method."""
+
+    @abstractmethod
+    def get(self):
+        """Abstract method to get the computed results.
+
+        :return: The computed results.
+        :rtype: object
+        """
+
+    @abstractmethod
+    def list(self):
+        """Abstract method to list information about the manager.
+
+        :return: A dictionary of properties.
+        :rtype: dict
+        """
+
+    @property
+    @abstractmethod
+    def name(self):
+        """Abstract property to get the name of the manager.
+
+        :return: The name of the manager.
+        :rtype: str
+        """
+
+    @abstractmethod
+    def save(self, path):
+        """Abstract method to save the manager.
+
+        :param path: The path to save the manager to.
+        :type path: str
+        """
+
+    @staticmethod
+    @abstractmethod
+    def load(path):
+        """Static method to load the manager.
+
+        :param path: The path to load the manager from.
+        :type path: str
+        """
--- a/raitools/raitools/_managers/counterfactual_manager.py
+++ b/raitools/raitools/_managers/counterfactual_manager.py
@@ -1,0 +1,51 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+"""Defines the Counterfactual Manager class."""
+
+from raitools._managers.base_manager import BaseManager
+
+
+class CounterfactualManager(BaseManager):
+    def __init__(self):
+        pass
+
+    def add(self,
+            continuous_features=None,
+            outcome_name=None,
+            total_CFs=None,
+            desired_range=None,
+            desired_class="opposite",
+            permitted_range=None,
+            features_to_vary=None,
+            method=None):
+        raise NotImplementedError(
+            "Add not implemented for CounterfactualManager")
+
+    def compute(self):
+        pass
+
+    def get(self):
+        raise NotImplementedError(
+            "Get not implemented for CounterfactualManager")
+
+    def list(self):
+        pass
+
+    @property
+    def name(self):
+        """Get the name of the counterfactual manager.
+
+        :return: The name of the counterfactual manager.
+        :rtype: str
+        """
+        return "counterfactual"
+
+    def save(self, path):
+        raise NotImplementedError(
+            "Save not implemented for CounterfactualManager")
+
+    @staticmethod
+    def load(path):
+        raise NotImplementedError(
+            "Load not implemented for CounterfactualManager")
--- a/raitools/raitools/_managers/explainer_manager.py
+++ b/raitools/raitools/_managers/explainer_manager.py
@@ -1,0 +1,138 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+"""Defines the Explainer Manager class."""
+
+import warnings
+from scipy.sparse import issparse
+from interpret_community.mimic.mimic_explainer import MimicExplainer
+from interpret_community.mimic.models.lightgbm_model import (
+    LGBMExplainableModel)
+from interpret_community.mimic.models.linear_model import (
+    LinearExplainableModel)
+from interpret_community.common.constants import ModelTask
+from raitools._managers.base_manager import BaseManager
+
+SPARSE_NUM_FEATURES_THRESHOLD = 1000
+
+
+class ExplainerManager(BaseManager):
+
+    """Defines the ExplainerManager for explaining a model.
+
+    :param model: The model to explain.
+        A model that implements sklearn.predict or sklearn.predict_proba
+        or function that accepts a 2d ndarray.
+    :type model: object
+    :param initialization_examples: A matrix of feature vector
+        examples (# examples x # features) for initializing the explainer.
+    :type initialization_examples: pandas.DataFrame
+    :param evaluation_examples: A matrix of feature vector
+        examples (# examples x # features) on which to explain the
+        model's output.
+    :type evaluation_examples: pandas.DataFrame
+    :param target_column: The name of the label column.
+    :type target_column: str
+    :param classes: Class names as a list of strings.
+        The order of the class names should match that of the model
+        output.  Only required if explaining classifier.
+    :type classes: list
+    """
+
+    def __init__(self, model, initialization_examples, evaluation_examples,
+                 target_column, classes=None):
+        """Defines the ExplainerManager for explaining a model.
+
+        :param model: The model to explain.
+            A model that implements sklearn.predict or sklearn.predict_proba
+            or function that accepts a 2d ndarray.
+        :type model: object
+        :param initialization_examples: A matrix of feature vector
+            examples (# examples x # features) for initializing the explainer.
+        :type initialization_examples: pandas.DataFrame
+        :param evaluation_examples: A matrix of feature vector
+            examples (# examples x # features) on which to explain the
+            model's output.
+        :type evaluation_examples: pandas.DataFrame
+        :param target_column: The name of the label column.
+        :type target_column: str
+        :param classes: Class names as a list of strings.
+            The order of the class names should match that of the model
+            output.  Only required if explaining classifier.
+        :type classes: list
+        """
+        self._model = model
+        self._initialization_examples = \
+            initialization_examples.drop(columns=[target_column])
+        self._evaluation_examples = \
+            evaluation_examples.drop(columns=[target_column])
+        self._is_run = False
+        self._is_added = False
+        self._surrogate_model = None
+        self._features = list(self._initialization_examples.columns)
+        self._classes = classes
+        self._target_column = target_column
+        self._explanation = None
+
+    def add(self):
+        """Add an explainer to be computed later."""
+        if self._is_added:
+            warnings.warn(("Ignoring.  Explanation has already been added, "
+                           "currently limited to one explainer type."),
+                          UserWarning)
+            return
+        is_sparse = issparse(self._evaluation_examples)
+        num_cols = self._evaluation_examples.shape[1]
+        many_cols = num_cols > SPARSE_NUM_FEATURES_THRESHOLD
+        self._surrogate_model = LGBMExplainableModel
+        if is_sparse and many_cols:
+            self._surrogate_model = LinearExplainableModel
+        self._is_added = True
+
+    def compute(self):
+        """Creates an explanation by running the explainer on the model."""
+        if self._is_run:
+            return
+        model_task = ModelTask.Unknown
+        explainer = MimicExplainer(self._model,
+                                   self._initialization_examples,
+                                   self._surrogate_model,
+                                   features=self._features,
+                                   model_task=model_task,
+                                   classes=self._classes)
+        self._explanation = explainer.explain_global(self._evaluation_examples)
+
+    def get(self):
+        """Get the computed explanation.
+
+        Must be called after add and compute methods.
+
+        :return: The computed explanations.
+        :rtype:
+            list[interpret_community.explanation.explanation.BaseExplanation]
+        """
+        if self._explanation:
+            return [self._explanation]
+        else:
+            return []
+
+    def list(self):
+        pass
+
+    @property
+    def name(self):
+        """Get the name of the explainer manager.
+
+        :return: The name of the explainer manager.
+        :rtype: str
+        """
+        return "explain"
+
+    def save(self, path):
+        raise NotImplementedError(
+            "Save not implemented for ExplainerManager")
+
+    @staticmethod
+    def load(path):
+        raise NotImplementedError(
+            "Load not implemented for ExplainerManager")
--- a/raitools/raitools/raianalyzer/constants.py
+++ b/raitools/raitools/raianalyzer/constants.py
@@ -1,0 +1,14 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+"""Defines common constants."""
+
+from enum import Enum
+
+
+class ModelTask(str, Enum):
+    """Provide model task constants. Can be 'classification' or 'regression'.
+    """
+
+    CLASSIFICATION = 'classification'
+    REGRESSION = 'regression'
--- a/raitools/raitools/raianalyzer/raianalyzer.py
+++ b/raitools/raitools/raianalyzer/raianalyzer.py
@@ -1,0 +1,146 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+"""Defines the RAIAnalyzer class."""
+
+from raitools._managers.causal_manager import CausalManager
+from raitools._managers.counterfactual_manager import CounterfactualManager
+from raitools._managers.error_analysis_manager import ErrorAnalysisManager
+from raitools._managers.explainer_manager import ExplainerManager
+from raitools._managers.fairness_manager import FairnessManager
+
+
+class RAIAnalyzer(object):
+
+    """Defines the top-level RAI Analyzer.
+
+    Use the RAI Analyzer to analyze errors, explain the most important
+    features, validate fairness, compute counterfactuals and run causal
+    analysis in a single API.
+
+    :param model: The model to explain.
+        A model that implements sklearn.predict or sklearn.predict_proba
+        or function that accepts a 2d ndarray.
+    :type model: object
+    :param train: The training dataset.
+    :type train: pandas.DataFrame
+    :param test: The test dataset.
+    :type test: pandas.DataFrame
+    :param target_column: The name of the label column.
+    :type target_column: str
+    :param task_type: The task to run, can be `classification` or
+        `regression`.
+    :type task_type: str
+    """
+
+    def __init__(self, model, train, test, target_column,
+                 task_type):
+        """Defines the top-level RAI Analyzer.
+
+        Use the RAI Analyzer to analyze errors, explain the most important
+        features, validate fairness, compute counterfactuals and run causal
+        analysis in a single API.
+
+        :param model: The model to explain.
+            A model that implements sklearn.predict or sklearn.predict_proba
+            or function that accepts a 2d ndarray.
+        :type model: object
+        :param train: The training dataset.
+        :type train: pandas.DataFrame
+        :param test: The test dataset.
+        :type test: pandas.DataFrame
+        :param target_column: The name of the label column.
+        :type target_column: str
+        :param task_type: The task to run, can be `classification` or
+            `regression`.
+        :type task_type: str
+        """
+        self.model = model
+        self.train = train
+        self.test = test
+        self.target_column = target_column
+        self.task_type = task_type
+        self._causal_manager = CausalManager()
+        self._counterfactual_manager = CounterfactualManager()
+        self._error_analysis_manager = ErrorAnalysisManager()
+        self._classes = train[target_column].unique()
+        self._explainer_manager = ExplainerManager(model, train, test,
+                                                   target_column,
+                                                   self._classes)
+        self._fairness_manager = FairnessManager()
+        self._managers = [self._causal_manager,
+                          self._counterfactual_manager,
+                          self._error_analysis_manager,
+                          self._explainer_manager,
+                          self._fairness_manager]
+
+    @property
+    def causal(self):
+        """Get the causal manager.
+
+        :return: The causal manager.
+        :rtype: CausalManager
+        """
+        return self._causal_manager
+
+    @property
+    def counterfactual(self):
+        """Get the counterfactual manager.
+
+        :return: The counterfactual manager.
+        :rtype: CounterfactualManager
+        """
+        return self._counterfactual_manager
+
+    @property
+    def erroranalysis(self):
+        """Get the error analysis manager.
+
+        :return: The error analysis manager.
+        :rtype: ErrorAnalysisManager
+        """
+        return self._error_analysis_manager
+
+    @property
+    def explainer(self):
+        """Get the explainer manager.
+
+        :return: The explainer manager.
+        :rtype: ExplainerManager
+        """
+        return self._explainer_manager
+
+    @property
+    def fairness(self):
+        """Get the fairness manager.
+
+        :return: The fairness manager.
+        :rtype: FairnessManager
+        """
+        return self._fairness_manager
+
+    def compute(self):
+        """Calls compute on each of the managers."""
+        for manager in self._managers:
+            manager.compute()
+
+    def list(self):
+        """List information about each of the managers.
+
+        :return: Information about each of the managers.
+        :rtype: dict
+        """
+        configs = {}
+        for manager in self._managers:
+            configs[manager.name] = manager.list()
+
+    def save(self, path):
+        # TODO: implement
+        raise NotImplementedError(
+            "Save not implemented for RAIAnalyzer")
+
+    @staticmethod
+    def load(path):
+        # TODO: implement
+        raise NotImplementedError(
+            "Load not implemented for RAIAnalyzer")
--- a/raitools/requirements.txt
+++ b/raitools/requirements.txt
@@ -1,0 +1,7 @@
+numpy>=1.17.2
+pandas>=0.25.1
+scipy>=1.4.1
+scikit-learn>=0.22.1
+lightgbm>=2.2.3
+interpret-community>=0.17.0
+fairlearn>=0.4.6
--- a/raitools/setup.py
+++ b/raitools/setup.py
@@ -1,0 +1,44 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+import os
+import setuptools
+
+# Version will be read from version.py
+version = ''
+name = 'raitools'
+# Fetch Version
+with open(os.path.join(name, '__version__.py')) as f:
+    code = compile(f.read(), f.name, 'exec')
+    exec(code)
+
+# Fetch ReadMe
+with open('README.md', 'r') as fh:
+    long_description = fh.read()
+
+# Use requirements.txt to set the install_requires
+with open('requirements.txt') as f:
+    install_requires = [line.strip() for line in f]
+
+setuptools.setup(
+    name=name,  # noqa: F821
+    version=version,  # noqa: F821
+    author="Roman Lutz, Ilya Matiach, Ke Xu",
+    author_email="raiwidgets-maintain@microsoft.com",
+    description="SDK API to assess fairness, explain "
+                "models, and analyze errors in Machine Learning",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    url="https://github.com/microsoft/responsible-ai-widgets",
+    packages=setuptools.find_packages(),
+    python_requires='>=3.6',
+    install_requires=install_requires,
+    classifiers=[
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
+        "License :: OSI Approved :: MIT License",
+        "Operating System :: OS Independent",
+        "Development Status :: 3 - Alpha"
+    ]
+)
--- a/raitools/tests/common_utils.py
+++ b/raitools/tests/common_utils.py
@@ -1,0 +1,100 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+# Defines common utilities for raitools tests
+import numpy as np
+import pandas as pd
+from sklearn import svm
+from sklearn.datasets import load_iris, load_breast_cancer, make_classification
+from sklearn.ensemble import RandomForestClassifier
+from sklearn.linear_model import LogisticRegression
+from sklearn.model_selection import train_test_split
+from sklearn.pipeline import Pipeline
+from lightgbm import LGBMClassifier
+from xgboost import XGBClassifier
+
+
+def create_sklearn_random_forest_classifier(X, y):
+    rfc = RandomForestClassifier(n_estimators=10, max_depth=4,
+                                 random_state=777)
+    model = rfc.fit(X, y)
+    return model
+
+
+def create_lightgbm_classifier(X, y):
+    lgbm = LGBMClassifier(boosting_type='gbdt', learning_rate=0.1,
+                          max_depth=5, n_estimators=200, n_jobs=1,
+                          random_state=777)
+    model = lgbm.fit(X, y)
+    return model
+
+
+def create_xgboost_classifier(X, y):
+    xgb = XGBClassifier(learning_rate=0.1, max_depth=3, n_estimators=100,
+                        n_jobs=1, random_state=777)
+    model = xgb.fit(X, y)
+    return model
+
+
+def create_sklearn_svm_classifier(X, y, probability=True):
+    clf = svm.SVC(gamma=0.001, C=100., probability=probability,
+                  random_state=777)
+    model = clf.fit(X, y)
+    return model
+
+
+def create_sklearn_logistic_regressor(X, y, pipeline=False):
+    lin = LogisticRegression(solver='liblinear')
+    if pipeline:
+        lin = Pipeline([('lin', lin)])
+    model = lin.fit(X, y)
+    return model
+
+
+def create_iris_data():
+    # Import Iris dataset
+    iris = load_iris()
+    # Split data into train and test
+    x_train, x_test, y_train, y_test = train_test_split(
+        iris.data, iris.target, test_size=0.2, random_state=0)
+    feature_names = [name.replace(' (cm)', '') for name in iris.feature_names]
+    classes = iris.target_names
+    return x_train, x_test, y_train, y_test, feature_names, classes
+
+
+def create_cancer_data():
+    breast_cancer_data = load_breast_cancer()
+    classes = breast_cancer_data.target_names.tolist()
+
+    # Split data into train and test
+    x_train, x_test, y_train, y_test = train_test_split(
+        breast_cancer_data.data, breast_cancer_data.target,
+        test_size=0.2, random_state=0)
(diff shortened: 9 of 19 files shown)


raianalysis/README.md
```diff
@@ -0,0 +1,16 @@
+# Responsible AI Analysis SDK for Python
+
+### This package has been tested with Python 3.5, 3.6 and 3.7
```
romanlutz, line 3: why not 3.8? We shouldn't really bother with 3.5 at this point IMO

raianalysis/raianalysis/raianalyzer/raianalyzer.py
```diff
@@ -0,0 +1,49 @@
...
+        pass
+
+    def list(self):
+        pass
+
+    def save(self, path):
+        pass
+
+    @staticmethod
+    def load(path):
```
riedgar-ms, line 48: This is a potential problem. If this is going to load a fully reusable `RAIAnalyzer` then the saved file is going to have to embed the `model`.

raianalysis/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+The responsible AI analysis sdk enables users to analyze their machine learning models in one API. Users will be able to analyze errors, explain the most important features, analyze fairness, compute counterfactuals and run causal analysis in a single API.
+
+Highlights of the package include:
+
+- add_fairness allows users to run fairlearn to assess model fairness
```
romanlutz, line 9: nit: add backquotes around function names to make them look like code

raianalysis/raianalysis/raianalyzer/raianalyzer.py
```diff
@@ -0,0 +1,49 @@
...
+                           method=None):
+        pass
+
+    def add_erroranalysis(self):
+        pass
+
+    def compute(self):
+        pass
+
+    def list(self):
```
riedgar-ms, line 41: What exactly are we listing? The things pending to compute?

raianalysis/requirements.txt
```diff
@@ -0,0 +1,7 @@
+numpy>=1.17.2
+pandas>=0.25.1
+scipy>=1.4.1
+scikit-learn>=0.22.1
+lightgbm>=3.1.1
+interpret-community
+fairlearn
```
romanlutz, line 7: Why not pin all of them?

raianalysis/raianalysis/raianalyzer/raianalyzer.py
```diff
@@ -0,0 +1,49 @@
...
+        self.train = train
+        self.test = test
+        self.target_column = target_column
+        self.transformation = transformation
+        self.task_type = task_type
+
+    def add_fairness(self, sensitive_features):
+        pass
+
+    def add_explainer(self):
```
riedgar-ms, line 18: I am concerned that these are going to end up having monstrously long argument lists as we gradually add features.

raianalysis/README.md
romanlutz: It actually says 3.6-3.8 in the setup file

imatiach-msft: done

raianalysis/README.md
```diff
@@ -0,0 +1,16 @@
+# Responsible AI Analysis SDK for Python
+
+### This package has been tested with Python 3.5, 3.6 and 3.7
+
+The responsible AI analysis sdk enables users to analyze their machine learning models in one API. Users will be able to analyze errors, explain the most important features, analyze fairness, compute counterfactuals and run causal analysis in a single API.
```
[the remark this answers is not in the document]
imatiach-msft: done

raianalysis/README.md
imatiach-msft: oh cool idea, done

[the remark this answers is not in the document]
imatiach-msft: done

raianalysis/requirements.txt
imatiach-msft: pinning makes it more difficult for users to install a package in an existing environment or to integrate it with other packages

raianalysis/raianalysis/raianalyzer/raianalyzer.py
imatiach-msft: great question, my question too!

raianalysis/raianalysis/raianalyzer/raianalyzer.py
imatiach-msft: +1

raianalysis/requirements.txt
romanlutz: Then we're just asking for people to complain why things don't work. Perhaps I expressed myself badly. I meant "why not put a min version"? Fairlearn pre-v0.5.0 shouldn't really be used because of massive breaking changes. @riedgar-ms probably has thoughts, too.

raianalysis/raianalysis/raianalyzer/raianalyzer.py
riedgar-ms: That was the motivation behind the 'managers' - each component could present its own API as it saw fit. For explanations, this might mean having separate `add_mimic_explainer()` and `add_lime_explainer()` methods, keeping the individual argument lists to within reasonable bounds. The reason for the extra layer of managers was to group related functionality together, and keep the API surface of the `RAIAnalyzer` object reasonably small.

The alternative proposed yesterday of having 'Assets' (not AML Assets... I don't like overloading the name) would also work in a similar way. We could start with a basic `ExplainerAsset` which had few, if any, arguments, but then have a `LIMEExplainerAsset` or `MimicExplainerAsset` etc. as required.

raianalysis/requirements.txt
riedgar-ms: I would request an inclusive lower bound of 0.5.0 on Fairlearn, though :-)

imatiach-msft: oh ok I added lower bound pins

raianalysis/raianalysis/raianalyzer/raianalyzer.py
imatiach-msft: for explanations the API should be really simple for now.  In future I can see either passing in an explainer object or doing something like 
```
add_explainer(MimicExplainerThing(...))
```

imatiach-msft: I agree it would be too much to have methods like add_mimic_explainer() and add_lime_explainer() for each individual thing

riedgar-ms: I'll go with `Thing` rather than `Asset` :-) I feel that that would keep things extensible and coherent.

raianalysis/README.md
```diff
@@ -0,0 +1,16 @@
+# Responsible AI Analysis SDK for Python
```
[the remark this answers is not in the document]
imatiach-msft: we decided to call it raitools, so closing this comment

raianalysis/raianalysis/raianalyzer/raianalyzer.py
imatiach-msft: we discussed, it will be a dictionary of properties like on ExplanationClient list_model_explanations

raianalysis/raianalysis/raianalyzer/raianalyzer.py
imatiach-msft: punting this to a later time, resolving for now

raitools/raitools/raianalyzer/raianalyzer.py
```diff
@@ -0,0 +1,69 @@
...
+from raitools._managers.causal_manager import CausalManager
+from raitools._managers.counterfactual_manager import CounterfactualManager
+from raitools._managers.error_analysis_manager import ErrorAnalysisManager
+from raitools._managers.explainer_manager import ExplainerManager
+from raitools._managers.fairness_manager import FairnessManager
+
+
+class RAIAnalyzer(object):
+    def __init__(self, model, train, test, target_column,
+                 task_type, classes=None):
```
riedgar-ms, line 15: Is `classes` used by anything other than Explainer? If not, shouldn't it be on the Explainer's `add()` method?

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -0,0 +1,135 @@
...
+        self._surrogate_model = None
+        self._features = list(self._initialization_examples.columns)
+        self._classes = classes
+        self._target_column = target_column
+        self._explanation = None
+
+    def add(self):
+        """Add an explainer to be computed later."""
+        if self._is_added:
+            warnings.warn(("Ignoring.  Explanation has already been added, "
```
riedgar-ms, line 80: Shouldn't a user be able to add as many explanations as they want?

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -0,0 +1,135 @@
...
+        :type evaluation_examples: pandas.DataFrame
+        :param target_column: The name of the label column.
+        :type target_column: str
+        :param classes: Class names as a list of strings.
+            The order of the class names should match that of the model
+            output.  Only required if explaining classifier.
+        :type classes: list
+        """
+        self._model = model
+        self._initialization_examples = \
```
riedgar-ms, line 65: I have a feeling that all the managers will be doing something like this... might want to refactor

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -0,0 +1,135 @@
...
+        if self._is_run:
+            return
+        model_task = ModelTask.Unknown
+        explainer = MimicExplainer(self._model,
+                                   self._initialization_examples,
+                                   self._surrogate_model,
+                                   features=self._features,
+                                   model_task=model_task,
+                                   classes=self._classes)
+        self._explanation = explainer.explain_global(self._evaluation_examples)
```
riedgar-ms, line 102: I thought we were going to allow users to chose between local and global?

raitools/raitools/raianalyzer/constants.py
```diff
@@ -0,0 +1,12 @@
...
+
+from enum import Enum
+
+
+class ModelTask(str, Enum):
+    """Provide model task constants. Can be 'classification' or 'regression'.
+    """
+
+    Classification = 'classification'
+    Regression = 'regression'
```
riedgar-ms, line 12: Probability? For when you have a classification problem but want to work with prediction probabilities for each class?

raitools/raitools/raianalyzer/constants.py
```diff
@@ -0,0 +1,12 @@
...
+# Licensed under the MIT License.
+
+from enum import Enum
+
+
+class ModelTask(str, Enum):
+    """Provide model task constants. Can be 'classification' or 'regression'.
+    """
+
+    Classification = 'classification'
```
riedgar-ms, line 11: BinaryClassification, surely?

raitools/raitools/raianalyzer/raianalyzer.py
imatiach-msft: it's used by error analysis too, I'm not sure if it is used by anything else, but it's an optional parameter (really it's only used for the visualizations, it's not used by the explainer and is just passed onto the explanation object)

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: I don't think it makes sense to add multiple explanations, since they are all computed on the same model/dataset, but it will make sense for other things since they have parameters on the add method

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: doesn't fairness use the whole dataset?

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: no we decided it doesn't make sense, since it doesn't change the visualization at all (the dashboard just uses the local importance values and always recomputes the global importance values)

raitools/raitools/_managers/explainer_manager.py
riedgar-ms: Might a user not want to have both Mimic and Shap explanations for a Model? OK, we're not allowing that now, but we should have the machinery set up to do that.

raitools/raitools/raianalyzer/constants.py
imatiach-msft: we automatically recognize this in our explainers, I'm not sure about other libraries like fairlearn, but if they need this specified I can add it

raitools/raitools/_managers/explainer_manager.py
riedgar-ms: That's actually something which popped up in the latest update to the AzureML API proposal. The `add()` method for the `FairnessManager` has acquired a `dataset` argument which can be `initialisation` `evaluation` or `both` (which would concat the two first)

raitools/raitools/raianalyzer/constants.py
imatiach-msft: a probability task?  Sorry I'm confused.  Is there a corresponding list of tasks in fairlearn?

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: yes, we can change that machinery later, but right now (and possibly in the future) I want to keep explainers as simple as possible in this new API

raitools/raitools/raianalyzer/raianalyzer.py
riedgar-ms: Have we figured out how to make a useful 'static' version of `ErrorAnalysis`? I wasn't sure if it was included in this yet. If it's only used for visualisation, when does it actually need to be supplied? How do we get from this object to starting up a Model Assessment Dashboard in a notebook?

raitools/raitools/raianalyzer/constants.py
riedgar-ms: It's one of the options for Fairlearn, where instead of a predicted class, you have a 'probability of class 1'. Perhaps a PM question?

raitools/raitools/raianalyzer/constants.py
riedgar-ms: We only have metrics for binary classification and regression (and the probability ones). The meaning of things like 'false positive rate' start becoming a little fuzzy when you have multiclass classification. Again, perhaps a PM question.

raitools/raitools/_managers/explainer_manager.py
riedgar-ms: But aren't we doing this so that users can extract an explanation object if they want as well? In that case, isn't there a difference?

raitools/raitools/raianalyzer/raianalyzer.py
imatiach-msft: it's an optional argument that is not required.  I have a feeling it will be used by other libraries too though.  I'm surprised fairlearn doesn't use it.  Do you just say "for class 1" instead of "for class >=50k" in the adult census income dataset for fairlearn dashboard, for example?  It's much easier to understand when you use class names.

riedgar-ms: The class names are just the unique values of the given sensitive feature column.

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -0,0 +1,135 @@
...
+        if is_sparse and many_cols:
+            self._surrogate_model = LinearExplainableModel
+        self._is_added = True
+
+    def compute(self):
+        """Creates an explanation by running the explainer on the model."""
+        if self._is_run:
+            return
+        model_task = ModelTask.Unknown
+        explainer = MimicExplainer(self._model,
```
riedgar-ms, line 96: Also, do the explainers also embed copies of the data? I've never dug into the details of the structure of the returned objects. If so, then we ought to delete it from this copy, and refer to the centrally held data.

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: The global explanation includes the local explanation for mimic explainer, so really there's not much difference, adding one more param here just increases the complexity unnecessarily for now.  We can always add it later.

raitools/raitools/_managers/explainer_manager.py
riedgar-ms: Also, are these going to be copies or references? We don't want to be keeping multiple copies of the data around.

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: The explanation has a copy of the dataset, and the explainer does too.  But, I'm not sure why we would want to delete the dataset from the explanation?  Is that just to reduce memory usage?

imatiach-msft: RAIAnalyzer won't be modifying the input dataset internally, will it?

raitools/raitools/raianalyzer/raianalyzer.py
imatiach-msft: Oh I see, I can remove this parameter then and assume that the classes are the unique values in the target_column of the training dataset, we can just really limit the scope for now

raitools/raitools/_managers/explainer_manager.py
riedgar-ms: Well, yes - to reduce memory usage. On the fairness side, I was expecting that we would only store the metrics, and not the copies of the `y_true` `y_pred` and `sensitive_feature` arrays. Those will all add up, for no good reason. Then the dashboard wanted the particular arrays, it could look them up in the central store in the Analyser object.

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: references, but removing datasets on the explanations means we need to make changes in interpret-community, I think that's out of scope for this PR but it might be something we can do long-term

riedgar-ms: You don't need to make changes to `interpret-community`. All you need to do is have the `get()` method reinsert them before returning the explanation object to the user.

imatiach-msft: Sorry I'm confused.  Why would we need to reinsert them if we want to get rid of them from the explanation?  And if we are ok with keeping them on the explanation, and they are references already, why would we want to reinsert them?

riedgar-ms: Will they still be references after we've saved to disk and reloaded?

If the user wants an explanation, then we would need to reinsert the datasets. However, if they're then firing up the ModelAssessmentDashboard, won't that know how to get the common datasets whenever required, rather than needing them embedded in the explanations themselves? Again, the the transition to the TS layer, will the data remain referenced, or would we end up with multiple copies?

raitools/raitools/_managers/explainer_manager.py
romanlutz: I think @minthigpen had this in her plans already. Can you confirm whether this is a short-term feature to add or not @minthigpen ?

raitools/raitools/raianalyzer/raianalyzer.py
imatiach-msft: I removed the parameter from RAIAnalyzer, and I just compute the classes based on the unique values in the target column on the training dataset now as you suggested.  In some cases models can't handle string classes though and user might want to keep them separate from training dataset, so I feel like long-term we will probably have to add it back as an optional parameter.

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: I think we agreed in several meetings that we would only support a single explainer for now, I definitely don't want to support all of them for the initial release, or even after that, it requires a lot of maintenance.  At that point, it would be much better to allow the user to pass in the computed explanation object instead of the explainer.

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: I think you've highlighted multiple issues here, but that code hasn't been implemented -- when it will be implemented, we should make sure to use references and not make copies.

raitools/raitools/_managers/explainer_manager.py
riedgar-ms: Certainly only support one particular explainer in what's shown to the user, but why not build the machinery for more now? We know that the requirement for having more than one explanation is coming, so why not build the infrastructure now?

raitools/raitools/_managers/explainer_manager.py
riedgar-ms: But why not build it in from the beginning?

We certainly could make it so that each component is a literal copy of what we have in the individual dashboards now. However, that then means we have technical debt going forward, in removing the duplication. And if we've written files like that, then things are particularly painful since we'll have to keep supporting the older files.

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: I guess I don't agree with that requirement when it comes to explainers, I'm not sure we will want to support more than one explainer.  Or, if we will, then we should just take in explanations, we shouldn't have even more wrappers around our explainers, we already have too many wrappers to maintain.

riedgar-ms: Users will never want to try explaining their model with more than one technique?

imatiach-msft: If they will, RAIAnalyzer should take in the explanations, not explainers.  Or if they want to customize their explanations this much they should use the ExplanationDashboard instead.  Trying to maintain all of the wrappers for everything will create a lot of bugs and will be a maintenance headache.

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: I wrote a new comment on teams.  Maybe we can allow users to return explanations without datasets on them.  There was already a user request for this:
https://github.com/interpretml/interpret-community/issues/368
I can add this as a new feature in interpret-community.

raitools/raitools/raianalyzer/constants.py
imatiach-msft: Asked PM to clarify, waiting for PM reply.

raitools/raitools/raianalyzer/constants.py
imatiach-msft: Asked PM to clarify, waiting for PM reply.

raianalysis/raianalysis/raianalyzer/raianalyzer.py
imatiach-msft: maybe we can take in a model serializer object that defines how to serialize the model?  Otherwise we can allow the user to set the model on the RAIAnalyzer.

raitools/raitools/_managers/explainer_manager.py
riedgar-ms: I may have been mixing up explanations and explainers and creating confusion that way.

Taking a step back, I'm thinking of a case where (once we've allowed the user to select whether they want to use LIME/shap/Mimic etc.) the user looks at the 'Explanation' tab of the dashboard and says "This isn't showing me what I need." Or perhaps "This shows me what I need to know in region X of parameter space, but not in region Y." They will then want to generate another interpretation of their model to rectify this, but without regenerating the entire `RAIAnalyzer` object - after all, they may well have been using other components and not want to be forced to recompute all of those.

In this case, they're going to end up with two interpretations hanging off the `ExplainerManager`.

Does that make more sense?

raitools/raitools/raianalyzer/constants.py
imatiach-msft: actually I'm a bit surprised about this issue because in the mock notebook from PM it just says:

```
task_type = 'classification or regression'
```

definitely would be great to get more clarity on this functionality

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: Yes, we discussed offline, agree that this is something we might want to plan/discuss more, but currently it's not needed and out of scope for the initial implementation.  Actually my hope is this API can somehow take an explanation instead or we just keep the simple default explainer, since adding every explainer/parameter will be nontrivial and could also create a lot of issues/bugs/confusion since we would then have to maintain those wrappers.  But for now based on our offline discussion we seemed to agree that it's fine to punt this until later, as changing this implementation shouldn't be too hard and shouldn't break backwards compatibility since we will always have a `default` case.

raianalysis/raianalysis/raianalyzer/raianalyzer.py
imatiach-msft: will leave this question to a future PR, as I won't implement save/load in this PR since it's already quite large for the functionality it already has

raitools/README.md
```diff
@@ -0,0 +1,16 @@
+# Responsible AI Tools SDK for Python
+
+### This package has been tested with Python 3.6, 3.7 and 3.8
+
+The Responsible AI Tools SDK enables users to analyze their machine learning models in one API. Users will be able to analyze errors, explain the most important features, validate fairness, compute counterfactuals and run causal analysis in a single API.
+
```
gaugup, line 6: Maybe change to "using a single API" towards the end

raitools/README.md
```diff
@@ -0,0 +1,16 @@
...
+Highlights of the package include:
+
+- `fairness.add()` allows users to run Fairlearn to assess model fairness
+- `explainer.add()` allows users to explain their model
+- `counterfactuals.add()` allows users to compute counterfactuals
+- `error_analysis.add()` allows users to run error analysis
+- `causal.add()` allows users to run causal analysis
```
gaugup, line 13: Weren't these suppose to to fairness_manager, explanation_manager and so on.

raitools/requirements.txt
```diff
@@ -0,0 +1,7 @@
+numpy>=1.17.2
+pandas>=0.25.1
+scipy>=1.4.1
+scikit-learn>=0.22.1
+lightgbm>=3.1.1
+interpret-community>=0.17.0
+fairlearn>=0.5.0
```
gaugup, line 7: the fairlearn dependency should be compatible with azuremlcli.

gaugup (changes requested):

raitools/README.md
```diff
@@ -0,0 +1,16 @@
+# Responsible AI Tools SDK for Python
+
+### This package has been tested with Python 3.6, 3.7 and 3.8
+
+The Responsible AI Tools SDK enables users to analyze their machine learning models in one API. Users will be able to analyze errors, explain the most important features, validate fairness, compute counterfactuals and run causal analysis in a single API.
+
+Highlights of the package include:
+
+- `fairness.add()` allows users to run Fairlearn to assess model fairness
```
gregorybchris, line 9: Nit: I'd prefer "runs Fairlearn to assess model fairness" since it's uncommon in my experience to talk about "users" in open source READMEs, but up to you

raitools/raitools/_managers/base_manager.py
```diff
@@ -0,0 +1,71 @@
...
+    """The base class for managers."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialize the BaseManager."""
+        super(BaseManager, self).__init__(*args, **kwargs)
+
+    @abstractmethod
+    def add(self):
+        """Abstract method to add a computation to the manager."""
+        pass
```
gregorybchris, line 19: Do you want to `raise UnimplementedError' for these abstract methods? I think sometimes abstract classes do that to make it obvious to callers when interfaces were not properly implemented by the child class

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -0,0 +1,135 @@
...
+    LGBMExplainableModel)
+from interpret_community.mimic.models.linear_model import (
+    LinearExplainableModel)
+from interpret_community.common.constants import ModelTask
+from raitools._managers.base_manager import BaseManager
+
+SparseNumFeaturesThreshold = 1000
+
+
+class ExplainerManager(BaseManager):
```
gregorybchris, line 19: If managers are going to be public we should deliberate a bit on the name of explainability/interpret that we want to use for all of these things. One reason to prefer interpret is that counterfactuals are also considered explanations. The distinction might start to get confusing if we don't have good names for these two components.

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -0,0 +1,135 @@
...
+from scipy.sparse import issparse
+from interpret_community.mimic.mimic_explainer import MimicExplainer
+from interpret_community.mimic.models.lightgbm_model import (
+    LGBMExplainableModel)
+from interpret_community.mimic.models.linear_model import (
+    LinearExplainableModel)
+from interpret_community.common.constants import ModelTask
+from raitools._managers.base_manager import BaseManager
+
+SparseNumFeaturesThreshold = 1000
```
gregorybchris, line 16: Can all constants use upper snake case? (`SPARSE_NUM_FEATURES_THRESHOLD`) Pascal case is usually reserved for classes in Python.

raitools/raitools/raianalyzer/constants.py
```diff
@@ -0,0 +1,14 @@
...
+"""Defines common constants."""
+
+from enum import Enum
+
+
+class ModelTask(str, Enum):
+    """Provide model task constants. Can be 'classification' or 'regression'.
+    """
+
+    Classification = 'classification'
```
gregorybchris, line 13: Same comment about upper snake case here

raitools/raitools/raianalyzer/raianalyzer.py
```diff
@@ -0,0 +1,144 @@
...
+
+        :return: Information about each of the managers.
+        :rtype: dict
+        """
+        configs = {}
+        for manager in self._managers:
+            configs[manager.name] = manager.list()
+
+    def save(self, path):
+        # TODO: implement
```
gregorybchris, line 138: Same comment about `UnimplementedError` here

raitools/setup.py
```diff
@@ -0,0 +1,44 @@
...
+    version=version,  # noqa: F821
+    author="Roman Lutz, Ilya Matiach, Ke Xu",
+    author_email="raiwidgets-maintain@microsoft.com",
+    description="SDK API to assess fairness, explain "
+                "models, and analyze errors in Machine Learning",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    url="https://github.com/microsoft/responsible-ai-widgets",
+    packages=setuptools.find_packages(),
+    python_requires='>=3.6',
```
gregorybchris, line 34: Are we leaving this as >=3.6 for now? Or should we stick on >=3.5 for a bit?

raitools/tests/common_utils.py
```diff
@@ -0,0 +1,100 @@
...
+                                                        test_size=0.2,
+                                                        random_state=0)
+    classes = np.unique(y_train).tolist()
+    feature_names = ["col" + str(i) for i in list(range(x_train.shape[1]))]
+    x_train = pd.DataFrame(x_train, columns=feature_names)
+    x_test = pd.DataFrame(x_test, columns=feature_names)
+    return x_train, y_train, x_test, y_test, classes
+
+
+def create_models(x_train, y_train):
```
gregorybchris, line 93: This would be a great time to use pytest fixtures, then you don't need to loop over models in the test function. Up to you though

raitools/README.md
imatiach-msft: done!

raitools/README.md
imatiach-msft: see this internal thread https://teams.microsoft.com/l/message/19:4df5f33d000a4ec98cfe6e17122ce022@thread.tacv2/1617285188141?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=fb4c3c0e-a518-43fa-9b92-96b7bdc852f7&parentMessageId=1617283381204&teamName=AzureML%20Responsible%20AI&channelName=RAI%20Engineering&createdTime=1617285188141
pasting image below if it helps:
image
note we use fairness directly, not fairness_manager now

raitools/requirements.txt
imatiach-msft: the hope is that azuremlcli will upgrade by the time this is made a dependency

raitools/README.md
imatiach-msft: done!

raitools/raitools/_managers/base_manager.py
imatiach-msft: seems it is discouraged:
https://stackoverflow.com/questions/45826692/body-of-abstract-method-in-python-3-5
https://stackoverflow.com/questions/60726156/best-way-to-implement-abstract-classes-in-python
actually I will just remove the pass as recommended in the first link, it seems it is redundant if you have doc strings

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: agree, but FeatureExplainerManager seems too long, and interpret is never used in code by industry, it's only used in product names.  Counterfactual Explanations are different from Feature Importance Explanations, but since the latter is more prevalent I would argue to keep it as just Explainer?

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: done!

raitools/raitools/raianalyzer/constants.py
imatiach-msft: done!

raitools/raitools/raianalyzer/raianalyzer.py
imatiach-msft: done, I also raised it in the managers, except for list and compute for now, since top level calls those on all managers

raitools/setup.py
imatiach-msft: leaving it for >=3.6 since raiwidgets is also >=3.6

raitools/tests/common_utils.py
imatiach-msft: won't the models be different depending on the datasets used?  eg in test_raianalyzer_iris I use a different dataset from test_raianalyzer_cancer when I call:
models = create_models(x_train, y_train)
I'm not sure I want to create a fixture with a list of models for every type of dataset, or did I misunderstand?  Maybe this might make more sense in the future if I start using the same dataset in multiple tests.

raitools/raitools/_managers/base_manager.py
gregorybchris: Oh! Good to know, thanks for sending the links

raitools/requirements.txt
gaugup: Can we still keep this at 0.4.6 as in azuremlcli and upgrade once azuremlcli upgrade?

raitools/raitools/_managers/counterfactual_manager.py
```diff
@@ -0,0 +1,47 @@
...
+            continuous_features=None,
+            outcome_name=None,
+            total_CFs=None,
+            desired_range=None,
+            desired_class="opposite",
+            permitted_range=None,
+            features_to_vary=None,
+            method=None):
+        raise NotImplementedError("Add is not implemented")
+
```
gaugup, line 23: Can we special case the message per manager? Like "Add for counterfactual manager is not implemented"? It would be difficult to figure out from the exception message as to which manager has not implemented.

gaugup (changes requested):

raitools/requirements.txt
imatiach-msft: ok done!

raitools/raitools/_managers/counterfactual_manager.py
imatiach-msft: done!

gregorybchris (approved):

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        if self._is_run:
+            return
+        model_task = ModelTask.Unknown
+        explainer = MimicExplainer(self._model,
+                                   self._initialization_examples,
+                                   self._surrogate_model,
+                                   features=self._features,
+                                   model_task=model_task,
+                                   classes=self._classes)
+        self._explanation = explainer.explain_global(self._evaluation_examples)
```
gaugup, line 102: Looks like we track only one one explanation object. Doesn't the get return multiple explanation objects?

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        Must be called after add and compute methods.
+
+        :return: The computed explanations.
+        :rtype:
+            list[interpret_community.explanation.explanation.BaseExplanation]
+        """
+        if self._explanation:
+            return [self._explanation]
+        else:
+            return []
```
gaugup, line 116: Even though we return a list this list only has one explanation. Should this be all the explanations computed by explanation manager.

raitools/requirements.txt
```diff
@@ -0,0 +1,7 @@
+numpy>=1.17.2
+pandas>=0.25.1
+scipy>=1.4.1
+scikit-learn>=0.22.1
+lightgbm>=3.1.1
```
gaugup, line 5: Is the minimum lightgbm version needed at 3.1.1? This might be an issue when itegrating with azuremlcli.

raitools/tests/test_raianalyzer.py
```diff
@@ -0,0 +1,63 @@
...
+
+def run_raianalyzer(model, x_train, x_test, target_column, classes):
+    error_analyzer = RAIAnalyzer(model, x_train, x_test, target_column,
+                                 task_type=ModelTask.CLASSIFICATION)
+    error_analyzer.explainer.add()
+    error_analyzer.explainer.compute()
+    explanations = error_analyzer.explainer.get()
+    assert isinstance(explanations, list)
+    assert len(explanations) == 1
+    explanation = explanations[0]
```
gaugup, line 59: Can we add a test case to compute multiple explanations and checking for all the computed explanations?

gaugup (changes requested):

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: there are no parameters to explanation manager currently and since we have same model/dataset it doesn't make sense to compute multiple explanations right now

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: exactly, this is by design, see comment above:
"there are no parameters to explanation manager currently and since we have same model/dataset it doesn't make sense to compute multiple explanations right now"

raitools/requirements.txt
imatiach-msft: changed it to lightgbm>=2.2.3

raitools/tests/test_raianalyzer.py
imatiach-msft: I added validation to ensure warning is printed in this case

raitools/raitools/_managers/explainer_manager.py
gaugup: Agreed that currently it can only do one type of explanation.

gaugup (approved):

Issue #406, opened 2021-03-24
MAD: Use context to replace all the inputs to various components


romanlutz:
This particular item is for the model-assessment dashboard, but we should use this everywhere in the future.

PR #431, opened 2021-04-01
Expand global context usage to jointDataset, cohorts, metadata, and theme

romanlutz:
This is another PR that's related to #406 .

This PR replaces several occurrences of `jointDataset`, `cohorts`, `metadata`, and `theme` being passed as props when they could just be used from the global context. Notably, sometimes they were used in the constructor, but context isn't available during the constructor yet, so for those cases I added `componentDidMount` methods to update the state afterwards.
 
This also fixes a typo (`stringifedValues` --> `stringifiedValues`).

In the two implementations of What-If (`WhatIf` and `WhatIfTab`) `featuresOption` is now part of the state rather than a private member. This is following @xuke444 's recommendation.

Finally, `ModelPerformanceTab` does not require any props anymore now! Since we need to pass something the former props interface is still there, but since empty interfaces cause eslint issues I've converted it to an empty class (which doesn't cause issues). I've also made the state non-public since there's really no reason for it to be public.


xuke444 (approved):

xuke444 (approved):

libs/core-ui/src/lib/Cohort/ShiftCohort/ShiftCohort.tsx
```diff
@@ -14,19 +14,23 @@ import {
...
   isOpen: boolean;
-  cohorts: ErrorCohort[];
   onDismiss: () => void;
   onApply: (selectedCohort: ErrorCohort) => void;
 }
 
 export interface IShiftCohortState {
   options: IDropdownOption[];
   selectedCohort: number;
+  savedCohorts: ErrorCohort[];
```
imatiach-msft, line 33: is this fixing some bug with the cohorts in shift cohort?  not sure why this change is needed

libs/dataset-explorer/src/lib/DatasetExplorerTab.tsx
```diff
@@ -87,18 +89,20 @@ export class DatasetExplorerTab extends React.PureComponent<
       return <div />;
     }
     const plotlyProps = generatePlotlyProps(
-      this.props.jointDataset,
+      this.context.jointDataset,
       this.state.chartProps,
-      this.props.cohorts[this.state.selectedCohortIndex]
+      this.context.cohorts.map((cohort) => cohort.cohort)[
```
imatiach-msft, line 94: are these error cohorts instead of cohorts?  maybe change name to errorCohorts, then you won't have cohort.cohort.name like below which is a bit confusing but instead:
```
errorCohort.cohort.name
```

libs/dataset-explorer/src/lib/DatasetExplorerTab.tsx
```diff
@@ -240,8 +244,8 @@ export class DatasetExplorerTab extends React.PureComponent<
           </div>
           <SidePanel
             chartProps={this.state.chartProps}
-            cohorts={this.props.cohorts}
-            jointDataset={this.props.jointDataset}
+            cohorts={this.context.cohorts.map((cohort) => cohort.cohort)}
```
imatiach-msft, line 247: same suggestion here, cohort -> errorCohorts

imatiach-msft (approved):

libs/dataset-explorer/src/lib/DatasetExplorerTab.tsx
romanlutz: I agree. This does look confusing. Will change!

libs/core-ui/src/lib/Cohort/ShiftCohort/ShiftCohort.tsx
romanlutz: Previously we passed in a subset of all cohorts into this component (specifically the saved ones only). That adds logic in the layer above. Now that we have the context available we can just as well have that filtering in here. That's why this removes the cohorts from input and adds them to the state which is set internally. I did just notice that I have to move the context usage to its own method, though. Doing that now

PR #432, opened 2021-04-02
move feature importances calculation code into error analysis package

imatiach-msft:
- move feature importances calculation code into error analysis package
- fix bug with categorical features when calculating feature importances related to new package
- remove spurious console.log in SVGTooltip that I see pop up a lot in console
- add tests for computing importances
- add all erroranalysis package tests to CI pipeline, so the tests are run in builds now
- release new error analysis package with changes

### DIFF
--- a/.github/workflows/CI-python.yml
+++ b/.github/workflows/CI-python.yml
@@ -10,7 +10,7 @@ jobs:
   ci-python:
     strategy:
       matrix:
-        packageDirectory: ["raiwidgets", "wrapped-flask"]
+        packageDirectory: ["raiwidgets", "wrapped-flask", "erroranalysis"]
         operatingSystem: [ubuntu-latest, macos-latest]
         pythonVersion: [3.6, 3.7, 3.8]
         exclude:
@@ -23,9 +23,6 @@ jobs:
           - packageDirectory: "wrapped-flask"
             operatingSystem: macos-latest
             pythonVersion: 3.8
-          # lightgbm doesn't install properly on MacOS
-          - packageDirectory: "raiwidgets"
-            operatingSystem: macos-latest
 
     runs-on: ${{ matrix.operatingSystem }}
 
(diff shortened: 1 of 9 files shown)


imatiach-msft:
close reopen for CI

.github/workflows/CI-python.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   ci-python:
     strategy:
       matrix:
-        packageDirectory: ["raiwidgets", "wrapped-flask"]
+        packageDirectory: ["raiwidgets", "wrapped-flask", "erroranalysis"]
```
romanlutz, line 13: Do you also want to add this to the release workflow?

.github/workflows/CI-python.yml
```diff
@@ -37,6 +34,16 @@ jobs:
         with:
           python-version: ${{ matrix.pythonVersion }}
 
+      - bash: brew install libomp
```
romanlutz, line 37: This isn't an issue for this PR at all, but I'm just curious why lightgbm doesn't do this. Any idea? I think you've contributed to lightgbm in the past (do I remember that correctly?) so maybe you know...
Might be worth fixing this at the root (=lightgbm) to avoid such manual workarounds, although I'm happy we're covering Mac builds now. Thanks for this! (and feel free to resolve, it's just a thought)

erroranalysis/erroranalysis/_internal/error_analyzer.py
```diff
@@ -82,6 +83,20 @@ def compute_error_tree(self, features, filters, composite_filters):
         return compute_json_error_tree(self, features, filters,
                                        composite_filters)
 
+    def compute_importances(self):
```
romanlutz, line 86: Always good to have docstrings for public methods

romanlutz (approved):

.github/workflows/CI-python.yml
imatiach-msft: it's in the install docs for lightgbm in macos:
https://github.com/microsoft/LightGBM/blob/master/docs/Installation-Guide.rst#macos

erroranalysis/erroranalysis/_internal/error_analyzer.py
imatiach-msft: yes will add doc strings later, right now this is all in _internal so it's all private anyway

.github/workflows/CI-python.yml
imatiach-msft: yes

PR #434, opened 2021-04-07
fix github pipeline

imatiach-msft:
use run instead of bash for macos specific code in the pipeline


romanlutz (approved):
Thanks for fixing!

PR #435, opened 2021-04-07
fix indexing issue on cohort shift

imatiach-msft:
I noticed sometimes the indexes are not sorted in order on cohort shift in the local importances view, I realized this is actually similar to a previous bug but somehow the sort method I added is not working - apparently, it was not doing anything because the index column was still the same in this specific case (in a different workflow, the sort column changed in the aggregate importances/dataset views).  This PR forces the sort to happen.

### DIFF
--- a/libs/core-ui/src/lib/Cohort/Cohort.ts
+++ b/libs/core-ui/src/lib/Cohort/Cohort.ts
@@ -72,8 +72,12 @@ export class Cohort {
       this.currentSortKey = columnName;
       this.currentSortReversed = false;
     }
+    if (reverse === undefined) {
+      reverse = false;
+    }
     if (this.currentSortReversed !== reverse) {
       this.filteredData.reverse();
+      this.currentSortReversed = true;
     }
   }
 


libs/core-ui/src/lib/Cohort/Cohort.ts
```diff
@@ -63,9 +63,10 @@ export class Cohort {
 
   public sort(
     columnName: string = JointDataset.IndexLabel,
-    reverse?: boolean
+    reverse?: boolean,
+    force?: boolean
   ): void {
-    if (this.currentSortKey !== columnName) {
+    if (this.currentSortKey !== columnName || force) {
```
romanlutz, line 69: So just to clarify: In the `force` (only) case the column name would be the same as the current sort key, but it's not sorted yet? Is there ever a reason not to sort? If not, we should probably just sort whenever the current sort key changes (?)

romanlutz (approved):

libs/core-ui/src/lib/Cohort/Cohort.ts
imatiach-msft: "In the force (only) case the column name would be the same as the current sort key, but it's not sorted yet"
exactly, I think it was in reverse order
"we should probably just sort whenever the current sort key changes"
yes, in this case sort key remained the same, but I think it was in reverse order

PR #444, opened 2021-04-13
Add empty causal and counterfactual projects

### LINKED ISSUE #412: MAD: create a stub for what-if based on current local explanation tab
This is a prereq for the future what-if tab

romanlutz:
closes #412


imatiach-msft (approved):

imatiach-msft:
this looks like the wrong issue?

> closes #412

xuke444 (approved):

romanlutz:
> this looks like the wrong issue?
> 
> > closes #412

It's actually the correct one. By "stub" I meant "empty project". I asked @xuke444 yesterday whether I should make it be exactly like the current local explanation tab from error analysis since that contains one version of what-if, so it would provide a starting point. He said empty is better and @zhb000 / @zhb789 (I think both of these are Bo) will take it from there.

imatiach-msft:
that issue is for what if tab though, not causal and counterfactual?  maybe the issue is just worded in a confusing way

romanlutz:
> that issue is for what if tab though, not causal and counterfactual? maybe the issue is just worded in a confusing way

Sorry, somehow this didn't show when I completed it. What-if is the same as counterfactuals. I do hope that we can refactor out the corresponding pieces from `interpret` and `error-analysis`, but I don't know if that's in scope for @zhb000 / @zhb789

PR #445, opened 2021-04-13
reduce the lower-bound on the lightgbm dependency

imatiach-msft:
reduce lightgbm lower bound dependency to >=2.0.11, since raiwidgets, raitools and erroranalysis are all compatible with it

### DIFF
--- a/erroranalysis/requirements.txt
+++ b/erroranalysis/requirements.txt
@@ -2,4 +2,4 @@ numpy>=1.17.2
 pandas>=0.25.1
 scipy>=1.4.1
 scikit-learn>=0.22.1
-lightgbm>=3.1.1
+lightgbm>=2.0.11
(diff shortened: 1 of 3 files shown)


erroranalysis/requirements.txt
```diff
@@ -2,4 +2,4 @@ numpy>=1.17.2
 pandas>=0.25.1
 scipy>=1.4.1
 scikit-learn>=0.22.1
-lightgbm>=3.1.1
\ No newline at end of file
+lightgbm>=2.0.11
```
gaugup, line 5: Why the magic version 2.0.11? Seems like there are even versions here: https://pypi.org/project/lightgbm/#history?

romanlutz: That's the one required by automl

romanlutz: or rather: the minimum required by automl

romanlutz (approved):

erroranalysis/requirements.txt
gaugup: oh. I thought it was 2.3.0. That is the upper bound.

gaugup (approved):

PR #446, opened 2021-04-13
implement save, load and list methods on RAIAnalyzer and ExplainerManager

imatiach-msft:
implement save, load and list methods on RAIAnalyzer and ExplainerManager


raitools/raitools/raianalyzer/raianalyzer.py
```diff
@@ -135,12 +156,85 @@ def list(self):
...
+        """Save the RAIAnalyzer to the given path.
+
+        :param path: The path to save the RAIAnalyzer to.
+        :type path: str
+        """
+        # save each of the individual managers
+        for manager in self._managers:
+            manager.save(Path(path) / manager.name)
+        # save current state
+        self.train.to_pickle(Path(path) / TRAIN)
```
riedgar-ms, line 168: We pickle data?

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -117,7 +133,24 @@ def get(self):
...
     def list(self):
-        pass
+        """List information about the ExplainerManager.
+
+        :return: A dictionary of properties.
+        :rtype: dict
+        """
+        props = {ListProperties.MANAGER_TYPE: self.name}
+        if self._explanation:
+            props['id'] = self._explanation.id
```
riedgar-ms, line 143: Should these keys be constants? And related, do we have a JSON schema for this?

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -126,13 +159,55 @@ def name(self):
...
         """
-        return "explain"
+        return ManagerNames.EXPLAINER
 
     def save(self, path):
-        raise NotImplementedError(
-            "Save not implemented for ExplainerManager")
+        """Save the ExplainerManager to the given path.
+
+        :param path: The path to save the ExplainerManager to.
```
riedgar-ms, line 167: Is the 'path' supposed to be a directory or a file?

raitools/raitools/raianalyzer/raianalyzer.py
```diff
@@ -135,12 +156,85 @@ def list(self):
...
+                serializer = pickle.load(file)
+            inst.__dict__['_' + SERIALIZER] = serializer
+            with open(top_dir / MODEL_PKL, 'rb') as file:
+                inst.__dict__[MODEL] = serializer.load(file)
+        else:
+            inst.__dict__['_' + SERIALIZER] = None
+            with open(top_dir / MODEL_PKL, 'rb') as file:
+                inst.__dict__[MODEL] = pickle.load(file)
+        # load each of the individual managers
+        manager_names = ManagerNames.get_managers()
```
riedgar-ms, line 226: Something feels strange about this. We have the list of names being extracted here, but then the `kvp_manager` has to list the managers explicitly.

raitools/raitools/raianalyzer/raianalyzer.py
```diff
@@ -135,12 +156,85 @@ def list(self):
...
+        managers = []
+        kvp_manager = {ManagerNames.CAUSAL: CausalManager,
+                       ManagerNames.COUNTERFACTUAL: CounterfactualManager,
+                       ManagerNames.ERROR_ANALYSIS: ErrorAnalysisManager,
+                       ManagerNames.EXPLAINER: ExplainerManager,
+                       ManagerNames.FAIRNESS: FairnessManager}
+        for manager_name in manager_names:
+            manager_cls = kvp_manager[manager_name]
+            manager = manager_cls.load(top_dir / manager_name, inst)
+            _manager_name = manager_name.replace(' ', '_')
```
riedgar-ms, line 236: This feels like its adding an assumption which could trip us up later

raitools/raitools/raianalyzer/raianalyzer.py
```diff
@@ -135,12 +156,85 @@ def list(self):
...
+            inst.__dict__['_' + _manager_name + '_manager'] = manager
```
riedgar-ms, line 237: I'm guessing that this is inserting the underscore-prefixed name into the object?

Given that everything is being listed in `kvp_manager` anyway, I'm not convinced that this complexity is getting us anything.

raitools/tests/test_raianalyzer.py
```diff
@@ -50,18 +54,38 @@ def test_raianalyzer_binary(self):
...
-    explanations = error_analyzer.explainer.get()
+        raianalyzer.explainer.add()
+    raianalyzer.explainer.compute()
+    validate_explainer(raianalyzer, x_train, x_test, classes)
+    with TemporaryDirectory() as tempdir:
+        path = Path(tempdir) / 'explanation'
+        # save the raianalyzer
+        raianalyzer.save(path)
+        # load the raianalyzer
+        raianalyzer = RAIAnalyzer.load(path)
```
riedgar-ms, line 70: Would be nice to check that you got back what you saved.

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: each manager defines its own dictionary of properties that it returns

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: it's a path to a directory

raitools/raitools/raianalyzer/raianalyzer.py
imatiach-msft: I need both manager names and the uninitialized manager classes.  I don't want to add a second name static property to managers, since I already have a name property on the managers (but I can't call it until they are initialized).  I thought getting the list from the enums might be a cleaner way to do it.

raitools/raitools/raianalyzer/raianalyzer.py
imatiach-msft: It's just for "error analysis" manager name.  If you prefer, I can change it to "error_analysis" instead and then this can be removed.

raitools/raitools/raianalyzer/raianalyzer.py
imatiach-msft: this is setting the managers on the loaded raianalyzer object:
```
self._causal_manager
self._counterfactual_manager
self._error_analysis_manager
self._explainer_manager
self._fairness_manager
```
how would you prefer to do it?

raitools/tests/test_raianalyzer.py
imatiach-msft: that's what validate_explainer() is doing in the test, each of the managers can be validated on raianalyzer after it's loaded back

imatiach-msft: added more validation for raianalyzer train/test dataset, task type and target column before and after serialization

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: I updated the docs to clarify it's a directory

raitools/raitools/raianalyzer/raianalyzer.py
riedgar-ms: I suppose that I was more expecting to see JSON. I don't have hugely strong opinions on the matter, though.

raitools/raitools/_managers/explainer_manager.py
riedgar-ms: Doesn't mean that the keys shouldn't be constants. Or that each manager can't provide its own JSON Schema

raitools/raitools/_managers/explainer_manager.py
riedgar-ms: The docstring should make that clear.

raitools/raitools/raianalyzer/raianalyzer.py
riedgar-ms: Wouldn't something like:
```python
self._explainer_manager = ExplainerManager.load(...)
```
be easier to understand? I don't quite see what the highly dynamic approach is getting us, particularly when the `kvp_manager` dict is created just before (and I don't see a check that the keys in that are matching those in `manager_names`).

raitools/tests/test_raianalyzer.py
riedgar-ms: But what about the rest of the `RAIAnalyzer` - the properties on the root object itself, for example. Plus, does the validation mentioned check that it got back exactly what it saved?

raitools/raitools/raianalyzer/raianalyzer.py
imatiach-msft: hmm sure, I could change it to json.  I just saw here that pickle is faster, especially the newer pickle version (pickle-p2 in the graph), and it seemed to be the more recommended way just based on the number of answers recommending it:
https://stackoverflow.com/questions/17098654/how-to-store-a-dataframe-using-pandas
But it's probably better to save as json to be platform invariant.  I don't think this save/load method will be used for uploading the various artifacts anyway (at least for explanations it currently can't, since current UI code won't be able to handle it?).

raitools/tests/test_raianalyzer.py
imatiach-msft: yes, exactly, I just added that validation in latest update

raitools/raitools/raianalyzer/raianalyzer.py
imatiach-msft: done, changed to json!

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: done, changed keys to constants

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -126,13 +158,55 @@ def name(self):
...
+        meta = {IS_RUN: self._is_run,
+                IS_ADDED: self._is_added}
+        with open(Path(path) / META_JSON, 'w') as file:
+            json.dump(meta, file)
 
     @staticmethod
-    def load(path):
-        raise NotImplementedError(
-            "Load not implemented for ExplainerManager")
+    def load(path, raianalyzer):
```
romanlutz, line 179: Any particular reason why it's `raianalyzer` and not `rai_analyzer` ?

romanlutz (approved):

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: done, renamed raianalyzer to rai_analyzer everywhere

imatiach-msft: wait, why don't we call raiwidgets rai_widgets by the same logic then?

raitools/raitools/raianalyzer/raianalyzer.py
imatiach-msft: I cleaned it up a bit more

raitools/raitools/raianalyzer/raianalyzer.py
imatiach-msft: I changed it to error_analysis

raitools/raitools/raianalyzer/raianalyzer.py
imatiach-msft: I removed get_managers

raitools/raitools/rai_analyzer/constants.py
```diff
@@ -0,0 +1,48 @@
...
+    MODEL = 'model'
+
+
+class ListProperties(str):
+    """Provide constants for listing manager properties."""
+
+    MANAGER_TYPE = 'manager_type'
+
+
+class ExplainerManagerKeys(str):
```
gaugup, line 40: Could ExplainerManagerKeys be private? If this is only meant for internal consumption and doesn't serve any purpose to the end user it should be private.

raitools/raitools/rai_analyzer/constants.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class Metadata(str):
+    """Provide constants for metadata and saved files."""
+
+    META_JSON = 'meta.json'
+    MODEL = 'model'
+
+
+class ListProperties(str):
```
gaugup, line 34: Could ListProperties be private? If this is only meant for internal consumption and doesn't serve any purpose to the end user it should be private.

raitools/raitools/rai_analyzer/constants.py
```diff
@@ -0,0 +1,48 @@
...
+    """Provide the manager names."""
+
+    CAUSAL = 'causal'
+    COUNTERFACTUAL = 'counterfactual'
+    ERROR_ANALYSIS = 'error_analysis'
+    EXPLAINER = 'explainer'
+    FAIRNESS = 'fairness'
+
+
+class Metadata(str):
```
gaugup, line 27: Could Metadata be private? If this is only meant for internal consumption and doesn't serve any purpose to the end user it should be private.

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -117,7 +132,24 @@ def get(self):
...
+        :rtype: dict
+        """
+        props = {ListProperties.MANAGER_TYPE: self.name}
+        if self._explanation:
+            props[Keys.ID] = self._explanation.id
+            props[Keys.METHOD] = self._explanation.method
+            props[Keys.MODEL_TASK] = self._explanation.model_task
+            props[Keys.MODEL_TYPE] = self._explanation.model_type
+            if FeatureImportanceExplanation._does_quack(self._explanation):
+                props[Keys.IS_RAW] = self._explanation.is_raw
```
gaugup, line 147: Is one of raw or engineered enough? Why add two?

raitools/raitools/rai_analyzer/rai_analyzer.py
```diff
@@ -0,0 +1,262 @@
...
+TEST = 'test'
+TARGET_COLUMN = 'target_column'
+TASK_TYPE = 'task_type'
+MODEL = Metadata.MODEL
+MODEL_PKL = MODEL + '.pkl'
+SERIALIZER = 'serializer'
+CLASSES = 'classes'
+MANAGERS = 'managers'
+META_JSON = Metadata.META_JSON
+
```
gaugup, line 28: The above constants should be private if these aren't user facing.

raitools/raitools/rai_analyzer/rai_analyzer.py
```diff
@@ -0,0 +1,262 @@
...
+class RAIAnalyzer(object):
+
+    """Defines the top-level RAI Analyzer.
+
+    Use the RAI Analyzer to analyze errors, explain the most important
+    features, validate fairness, compute counterfactuals and run causal
+    analysis in a single API.
+
+    :param model: The model to explain.
+        A model that implements sklearn.predict or sklearn.predict_proba
```
gaugup, line 39: Should we say  "The model to compute RAI insights for" instead of " The model to explain"

raitools/raitools/rai_analyzer/rai_analyzer.py
```diff
@@ -0,0 +1,262 @@
...
+        or function that accepts a 2d ndarray.
+    :type model: object
+    :param train: The training dataset.
+    :type train: pandas.DataFrame
+    :param test: The test dataset.
+    :type test: pandas.DataFrame
+    :param target_column: The name of the label column.
```
gaugup, line 46: We should say for test/train "The training dataset including the label column"

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -126,13 +158,55 @@ def name(self):
...
+        """Save the ExplainerManager to the given path.
+
+        :param path: The directory path to save the ExplainerManager to.
+        :type path: str
+        """
+        top_dir = Path(path)
+        # save the explanation
+        if self._explanation:
+            save_explanation(self._explanation, top_dir / EXPLANATION)
+        meta = {IS_RUN: self._is_run,
```
gaugup, line 173: It might be easier if we choose the same sub-directory as the manager name. Here manager name is "explainer" but the directory where the explanation will be stored is explanation. It will be more intuitive for us,

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -126,13 +158,55 @@ def name(self):
...
+                IS_ADDED: self._is_added}
+        with open(Path(path) / META_JSON, 'w') as file:
+            json.dump(meta, file)
```
gaugup, line 177: We should also store the config of the explanation manager. Or since explanation manager is one static onfig it doesn't matter. I guess other components will have to do that as well. So maybe add a comment here to recover the configs in the respective managers.

raitools/raitools/rai_analyzer/rai_analyzer.py
```diff
@@ -0,0 +1,262 @@
...
+            inst.__dict__['_' + SERIALIZER] = serializer
+            with open(top_dir / MODEL_PKL, 'rb') as file:
+                inst.__dict__[MODEL] = serializer.load(file)
+        else:
+            inst.__dict__['_' + SERIALIZER] = None
+            with open(top_dir / MODEL_PKL, 'rb') as file:
+                inst.__dict__[MODEL] = pickle.load(file)
+        # load each of the individual managers
+        managers = []
+        kvp_manager = {ManagerNames.CAUSAL: CausalManager,
```
riedgar-ms, line 252: I'm still not clear why this dictionary and subsequent loop is preferable to something like:
```python
inst._explanation_manager = ExplanationManager.load(top_dir)
```
I think that would be a lot clearer

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -126,13 +158,55 @@ def name(self):
...
+        train = rai_analyzer.train.drop(columns=[target_column])
+        test = rai_analyzer.test.drop(columns=[target_column])
+        inst.__dict__[U_INITIALIZATION_EXAMPLES] = train
+        inst.__dict__[U_EVALUATION_EXAMPLES] = test
+        inst.__dict__['_' + FEATURES] = list(train.columns)
+        inst.__dict__['_' + IS_ADDED] = False
+        # reset self._surrogate_model
+        if meta[IS_ADDED]:
+            inst.add()
+        return inst
```
gaugup, line 212: So if save/load catch an exception the entire rai_analyzer process will crash. Do we need this to be more graceful so that individual component don't crash the entire rai_analyzer. I guess it applies to other functions as well. But this is something to keep in mind while designing these managers..

raitools/raitools/rai_analyzer/rai_analyzer.py
imatiach-msft: I feel like calling load method on each manager:
inst._explanation_manager = ExplanationManager.load(top_dir, rai_analyzer)
creates a lot of duplicate load method calls, this way is cleaner and reduces redundant code?

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: explanations can be raw, engineered, or neither?

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: great question, I was thinking about this as well, I'm not sure at this point yet, but my intuition tells me we should fail if any one things fails to save/load

imatiach-msft: hmm maybe I can at least improve the exception handling logic here

raitools/raitools/rai_analyzer/constants.py
imatiach-msft: done!

raitools/raitools/rai_analyzer/constants.py
imatiach-msft: done!

raitools/raitools/rai_analyzer/constants.py
imatiach-msft: done!

raitools/raitools/rai_analyzer/rai_analyzer.py
imatiach-msft: done!

raitools/raitools/rai_analyzer/rai_analyzer.py
imatiach-msft: done!

raitools/raitools/rai_analyzer/rai_analyzer.py
imatiach-msft: done!

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: done!

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: discussed, not needed in explainer_manager, but seems like a great idea for other explainers

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: discussed, the module name is raiwidgets, file and variables rai_widgets, to follow python conventions

raitools/raitools/rai_analyzer/rai_analyzer.py
riedgar-ms: Given that you explicitly list out all the managers in a dictionary immediately beforehand, the current way is guaranteed to take more lines of code and implicitly apply a naming convention on the managers. And you still have to read it closely to figure out what's going on.

riedgar-ms: If you want to do this with reflection, why not iterate through the properties of the class, and call `load()` on those derived from `BaseManager` ?

imatiach-msft: "
If you want to do this with reflection, why not iterate through the properties of the class, and call load() on those derived from BaseManager ?
"
I don't think I can do that in a static method on a class that does not yet exist?  The managers haven't been created yet?

riedgar-ms: Can you declare the return types with the `->` syntax?

imatiach-msft: done, used reflection on type hints

raitools/raitools/raianalyzer/rai_analyzer.py
```diff
@@ -0,0 +1,266 @@
...
+            inst.__dict__['_' + _SERIALIZER] = serializer
+            with open(top_dir / _MODEL_PKL, 'rb') as file:
+                inst.__dict__[_MODEL] = serializer.load(file)
+        else:
+            inst.__dict__['_' + _SERIALIZER] = None
+            with open(top_dir / _MODEL_PKL, 'rb') as file:
+                inst.__dict__[_MODEL] = pickle.load(file)
+        # load each of the individual managers
+        managers = []
+        for prop_name, prop_type in dict(vars(RAIAnalyzer)).items():
```
riedgar-ms, line 255: So this is at least consistently using reflection, but.... is it really simpler than just having
```python
self._explanation_manager = ExplanationManager.load(....)
```
?

I really feel that this obfuscates what's going on, when listing out each manager load explicitly would be clearer.

imatiach-msft: you mean inst.__dict__['_explanation_manager'] = ExplanationManager.load(top_dir/manager_name, inst)
I feel like it's just as many lines of code (10 here, that would probably also be 10 lines), but this is more maintainable and reduces code duplication.  Also means if we add or remove more managers we don't have to modify this code.

riedgar-ms: You're right, it would be
```python
inst._explanation_manager = ExplanationManager.load(...)
```
There would indeed be one of those for each manager.... five right now, I believe? Personally, I'd prefer the explicit listing, rather than doing reflection on the properties, which also ends up implicitly imposing a naming convention on the managers and properties.

imatiach-msft: done, changed code to call managers explicitly instead of using the code with reflection

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -126,13 +159,56 @@ def name(self):
...
+        :type rai_analyzer: RAIAnalyzer
+        """
+        # create the ExplainerManager without any properties using the __new__
+        # function, similar to pickle
+        inst = ExplainerManager.__new__(ExplainerManager)
+        top_dir = Path(path)
+        explanation_path = top_dir / ManagerNames.EXPLAINER
+        if explanation_path.exists():
+            explanation = load_explanation(explanation_path)
+            inst.__dict__['_' + ManagerNames.EXPLAINER] = explanation
```
riedgar-ms, line 196: Could even just do `inst._explainer_manager=explanation`

What we really want here is something like C#'s `public explanation_manager { get; private set; }` but this is probably the best we can do.

riedgar-ms (approved):

riedgar-ms:
Could you make the PR title a bit more explicit too?

raitools/raitools/_managers/explainer_manager.py
gaugup: I thought that explanations could be either raw/engineered. Is that not so?

gaugup (changes requested):

raitools/raitools/_managers/explainer_manager.py
```diff
@@ -126,13 +159,56 @@ def name(self):
...
+        with open(top_dir / META_JSON, 'r') as meta_file:
+            meta = meta_file.read()
+        meta = json.loads(meta)
+        inst.__dict__['_' + IS_RUN] = meta[IS_RUN]
+        inst.__dict__['_' + CLASSES] = rai_analyzer._classes
+        target_column = rai_analyzer.target_column
+        train = rai_analyzer.train.drop(columns=[target_column])
+        test = rai_analyzer.test.drop(columns=[target_column])
+        inst.__dict__[U_INITIALIZATION_EXAMPLES] = train
+        inst.__dict__[U_EVALUATION_EXAMPLES] = test
```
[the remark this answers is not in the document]
riedgar-ms: I hadn't noticed that - and I agree 100%. A major benefit of grouping all these components together is that things like the datasets only need to be stored once. If users want to request a particular component object (e.g. an `Explanation` object from `interpret-community`) then the appropriate manage should take responsibility for assembling it from the component specific data and the common data (held by the `RAIAnalyzer`).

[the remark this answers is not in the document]
imatiach-msft: "initialization samples and evaluation samples are both saved by rai-analyzer and the the individual managers"
no, the managers just load the datasets from the raianalyzer, they don't save them?

raitools/raitools/_managers/explainer_manager.py
imatiach-msft: it can be neither too, the flag will not be set if transformations/feature maps was not passed in.  We have a bunch of tests that validate this too, for example see:

```
    def test_get_global_raw_explanations_classification(self, iris, tabular_explainer):
        model = create_sklearn_svm_classifier(iris[DatasetConstants.X_TRAIN], iris[DatasetConstants.Y_TRAIN])

        exp = tabular_explainer(model, iris[DatasetConstants.X_TRAIN], features=iris[DatasetConstants.FEATURES],
                                classes=iris[DatasetConstants.CLASSES])

        global_explanation = exp.explain_global(iris[DatasetConstants.X_TEST])
        assert not global_explanation.is_raw
        assert not global_explanation.is_engineered
```

in interpret-community\test\raw_explain\test_raw_explanation.py:

https://github.com/interpretml/interpret-community/blob/master/test/raw_explain/test_raw_explanation.py#L28

in fact there are even more weird cases where it can be neither, for example if it's not a FeatureImportanceExplanation:

https://github.com/interpretml/interpret-community/blob/master/python/interpret_community/explanation/explanation.py#L1805

although technically all of them are currently

gaugup: Thanks. Didn't know that

raitools/raitools/_managers/explainer_manager.py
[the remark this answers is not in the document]
gaugup: Thanks for confirming.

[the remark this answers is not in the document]
gaugup: Thanks for clarification.

gaugup (approved):

PR #449, opened 2021-04-15
Use current cohort in dataset explorer

### LINKED ISSUE #413: MAD: make data explorer use the currently selected cohort
Includes removing custom cohort selection dropdown (see below) in favor of using global cohort settings

image

romanlutz:
closes #413 

This PR removes the cohort selection dropdown in `error-analysis` and `model-assessment`, but not in `interpret` since there's no global cohort switching mechanism in `interpret`.
Also, I moved `selectedCohort` into the global context since that seemed very useful.

Interestingly, the `initialCohortIndex` wasn't really used anyway, so removing it as an argument made no difference in behavior. I wanted to remove the corresponding props altogether, but our linter doesn't like empty interfaces or props in any shape (`{}`, `null`, etc.) so I just converted it into an empty class.

A more comprehensive fix would have been to add global cohort control to `interpret` to completely get rid of the dropdown, but that's quite a bit more work than we've budgeted on this.

Finally, I also commented out the causality and counterfactuals components in `model-assessment` since they're causing the entire dashboard to crash.


libs/model-assessment/src/lib/ModelAssessmentDashboard/ModelAssessmentDashboard.tsx
```diff
@@ -15,7 +15,7 @@ import {
   IGenericChartProps,
   ModelAssessmentContext
 } from "@responsible-ai/core-ui";
-import { CounterfactualsTab } from "@responsible-ai/counterfactuals";
+// import { CounterfactualsTab } from "@responsible-ai/counterfactuals";
```
imatiach-msft, line 18: remove?

libs/model-assessment/src/lib/ModelAssessmentDashboard/ModelAssessmentDashboard.tsx
```diff
@@ -384,8 +387,8 @@ export class ModelAssessmentDashboard extends CohortBasedComponent<
                         theme={this.props.theme}
                       />
                     )}
-                    <CausalInsightsTab />
-                    <CounterfactualsTab />
+                    {/* <CausalInsightsTab />
```
imatiach-msft, line 390: remove?

imatiach-msft (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/ModelAssessmentDashboard.tsx
romanlutz: We do want this there, but it's currently crashing the app because it's not return anything. I'd like to leave it here as a placeholder for @zhb000 / @zhb789 if possible

romanlutz: I added a comment so that it doesn't just look like we forgot about it.

PR #461, opened 2021-04-20
make save and load methods private on each of the managers

imatiach-msft:
Add "_" to the save and load methods on each of the individual managers.  Only the top-level RAIAnalyzer save/load APIs will be public.

### DIFF
--- a/raitools/raitools/_managers/base_manager.py
+++ b/raitools/raitools/_managers/base_manager.py
@@ -47,7 +47,7 @@ class BaseManager(ABC):
         """
 
     @abstractmethod
-    def save(self, path):
+    def _save(self, path):
         """Abstract method to save the manager.
 
         :param path: The directory path to save the manager to.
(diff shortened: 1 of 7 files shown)


raitools/raitools/_managers/base_manager.py
```diff
@@ -56,7 +56,7 @@ def save(self, path):
 
     @staticmethod
     @abstractmethod
-    def load(path, rai_analyzer):
+    def _load(path, rai_analyzer):
```
gaugup, line 59: These methods are in private directory. do we need to make these private as well? Seems like unnecessary to me.

imatiach-msft: yes, we are being extra careful, since other methods appear on managers and the methods on managers are public since they are exposed through the main RAIAnalyzer class, we've discussed this before

riedgar-ms (approved):

gregorybchris (approved):

PR #469, opened 2021-04-24
disable zoom in error analysis tree

xuke444:


imatiach-msft:
built this PR locally, the labels seems to be broken:

image

also the tree seems really small, the text is hard to see.  Maybe in error analysis app we should keep zoom, and only disable in model assessment dashboard?

romanlutz (approved):

PR #471, opened 2021-04-26
Fix dependencies for raitools

riedgar-ms:
The `raitools` package has a dependency on `erroranalysis`

Signed-off-by: Richard Edgar <riedgar@microsoft.com>


imatiach-msft (approved):

imatiach-msft:
I need to do another release of error analysis as well with the recent PR merge, raitools tests will fail otherwise if you don't have the latest error analysis package installed locally

imatiach-msft:
no this is fine to merge I think

PR #480, opened 2021-05-04
Casual aggregate chart

zhb000:


imatiach-msft:
the info buttons don't seem to work when I click on them:

image

zhb000:
> the info buttons don't seem to work when I click on them:
> 
> image



> the info buttons don't seem to work when I click on them:
> 
> image

This is the draft version, I will add more function in following PR

xuke444 (approved):

PR #510, opened 2021-05-07
Add manual cohort creation to the model assessment dashboard

### LINKED ISSUE #408: MAD: Manual cohort creation in model-assessment
Add the capability as it exists in the explanation dashboard to the model assessment dashboard. Design TBD

romanlutz:
closes #408 

Manual cohort creation means that we can create cohorts by selecting certain filters, rather than just through heat or tree maps. This feature already existed in `interpret`, of course. To make the functionality available for `model-assessment` this PR moves the code to `core-ui` and then uses a subset thereof (namely `CohortEditor`) to achieve the desired functionality. This is a bit simplified since we don't need to worry about cohort editing and deletion (unlike in `interpret`) because there's a custom experience for that inherited from `error-analysis`' way of handling cohorts.

The same `ShiftCohort`, `CohortEditor`, and `SaveCohort` components from `interpret` are reused and added to the `ModelAssessmentDashboard`. However, we now have buttons to change the cohort or create a new cohort in both the cohort settings panel (see `CohortSettingsPanel`) and the `CohortInfoSection` which is a new section at the very top above the error analysis section. Next to the tree or heat map view we also have a new cohort summary section similar to the previous `CohortInfo` that was reachable via dropdown.

Screenshots:

image

image

image


xuke444 (approved):

libs/core-ui/src/lib/Cohort/CohortInfo/CohortInfoPanel.tsx
```diff
@@ -0,0 +1,59 @@
...
+> {
+  public render(): React.ReactNode {
+    return (
+      <Panel
+        headerText={localization.ErrorAnalysis.CohortInfo.cohortInformation}
+        isOpen={this.props.isOpen}
+        focusTrapZoneProps={focusTrapZoneProps}
+        // You MUST provide this prop! Otherwise screen readers will just say "button" with no label.
+        closeButtonAriaLabel="Close"
+        // layerProps={{ hostId: this.props.hostId }}
```
imatiach-msft, line 46: maybe remove layerProps comment?

imatiach-msft (approved):

PR #543, opened 2021-05-16
add model analysis widget 

xuke444:
support multi instance for each component, currently default to [0] will add selection ui later.
added class names to model analysis
added categorical feature to model analysis
fix error analysis with display numerical categorical feature
Todo: 
error analysis filter in model analysis is not used.
static tree view when load initially
share error_analysis and model analysis dashboard input in python
context immutable 
pip install package and version control of raiwidget and raianalysis


.github/workflows/CI-python.yml
```diff
@@ -38,11 +42,19 @@ jobs:
         name: Use Homebrew to install libomp on MacOS for lightgbm
         run: brew install https://publictestdatasets.blob.core.windows.net/data/libomp-11.1.0.catalina.bottle.tar.gz
 
-      - name: Install dependencies
+      - name: Upgrade pi
```
romanlutz, line 45: pi?

apps/widget/src/app/ModelAssessment.tsx
```diff
@@ -42,49 +40,20 @@ export class ModelAssessment extends React.Component {
...
-        theme={modelData.theme}
-        supportedBinaryClassificationPerformanceKeys={
-          modelData.classification_methods
-        }
-        supportedRegressionPerformanceKeys={modelData.regression_methods}
-        supportedProbabilityPerformanceKeys={modelData.probability_methods}
+        localUrl={config.baseUrl}
+        locale={config.locale}
+        theme={undefined}
+        supportedBinaryClassificationPerformanceKeys={[]}
```
romanlutz, line 54: this only exists for fairness, so perhaps you want to remove it...

responsibleai/requirements-dev.txt
```diff
@@ -0,0 +1,30 @@
+# Requirements for raiwidgets development
```
romanlutz, line 1: raiwidgets --> should it be responsibleai ?

responsibleai/requirements-dev.txt
```diff
@@ -0,0 +1,30 @@
...
+# Required for notebook tests
+nbformat
+papermill
+scrapbook
+jupyter
+nbval
+
+sphinx==3.1.1
+sphinx-gallery==0.8.1
+pydata-sphinx-theme==0.3.0
```
romanlutz, line 29: Are we deploying this to a website? If not why are we building docs?

raiwidgets/tests/model_analysis.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+from raiwidgets import ModelAnalysisDashboard
+from responsibleai import ModelAnalysis
+import sklearn
+import shap
```
romanlutz, line 7: Is shap a dependency?

raiwidgets/requirements-dev.txt
```diff
@@ -28,4 +28,5 @@ nbval
 
 sphinx==3.1.1
 sphinx-gallery==0.8.1
-pydata-sphinx-theme==0.3.0
\ No newline at end of file
+pydata-sphinx-theme==0.3.0
+../responsibleai/.
```
romanlutz, line 32: Does this actually work? Perhaps just saying "responsibleai" would be fine?

raiwidgets/raiwidgets/model_analysis_dashboard_input.py
```diff
@@ -0,0 +1,357 @@
...
+        except Exception as e:
+            print(e)
+            traceback.print_exc()
+            return {
+                WidgetRequestResponseConstants.error: "Model threw exception"
+                " while predicting...",
+                WidgetRequestResponseConstants.data: []
+            }
+
+    def debug_ml(self, data):
```
romanlutz, line 75: I'm assuming most of this file is duplicated from error analysis, right? If so, can we get that functionality in a common super class instead of duplicating it?

raiwidgets/raiwidgets/model_analysis_dashboard.py
```diff
@@ -0,0 +1,58 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+"""Defines the Explanation dashboard class."""
```
romanlutz, line 4: please adjust doc

raiwidgets/raiwidgets/interfaces.py
```diff
@@ -0,0 +1,79 @@
...
+    EBM = "EBM"
+    blackbox = "blackbox"
+
+
+class ModelMethod(str, Enum):
+    classifier = "classifier"
+    regressor = "regressor"
+
+
+class ModelExplanationData(object):
```
romanlutz, line 54: do we need to duplicate all of these compared to error analysis?

raiwidgets/raiwidgets/fairness_dashboard.py
```diff
@@ -82,14 +82,12 @@ def __init__(
                 metrics_module.probability_methods,
         }
 
-        if locale is not None:
```
romanlutz, line 85: What if locale is None?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Context/buildModelAssessmentContext.ts
```diff
@@ -39,25 +39,23 @@ export function buildInitialModelAssessmentContext(
...
     metadata: modelMetadata,
-    predictedProbabilities: props.modelExplanationData.probabilityY,
-    predictedY: props.modelExplanationData.predictedY,
+    predictedProbabilities: props.dataset.probabilityY,
+    predictedY: props.dataset.predictedY,
     trueY: props.dataset.trueY
   });
   const globalProps = buildGlobalProperties(
-    props.modelExplanationData.precomputedExplanations
+    props.modelExplanationData?.[0]?.precomputedExplanations
```
romanlutz, line 58: This `?.[0]?` pops up all over the place. Should it really be a list if we always take the first element? Not just referring to this instance but all such in several files.

responsibleai/requirements-dev.txt
xuke444: I copied this file from raiwidget
removed.

raiwidgets/raiwidgets/model_analysis_dashboard_input.py
xuke444: yes,
I didn't do that because we will remove error analysis eventually, didn't want to handle difference through extra parameter to shared super class, or accidentally break existing error anslysis.

raiwidgets/raiwidgets/fairness_dashboard.py
xuke444: locale is passed into super (dashboard) class and get transmitted to ui through "config" object, this part is not used in ui, that's why remove.
If locale is none, we use en-us by default

libs/model-assessment/src/lib/ModelAssessmentDashboard/Context/buildModelAssessmentContext.ts
xuke444: no it should not, I listed few things in this pr comment.
This one we need to build a picker for user choose from.
Don't want to make this pr too huge that's why I always pick first one here.

raiwidgets/raiwidgets/interfaces.py
xuke444: same as previous, I don't want to bother to share things with assumption error-analysis and interpret dashboard will be gone eventually, and the code even right now are different between error-analysis and mad.

And this one is done through contract(class) vs error-analysis and interpret was defined by constant of property name.

So this way has extra benefit on define nested structure.
Eventually I will find someway either convert py to ts or the other way, so we can make sure contract synced between backend and front end. To do that I will have to create this python class contract first and use it in py.

raiwidgets/raiwidgets/model_analysis_dashboard.py
xuke444: done

raiwidgets/tests/model_analysis.py
xuke444: through interpret-community
I don't want to introduce extra dependency because of one dataset.
If this is concern we can check if other dataset is good to use.
For now this file is just for manual testing, we don't have automated on it.

responsibleai/requirements-dev.txt
xuke444: done

apps/widget/src/app/ModelAssessment.tsx
xuke444: removed

.github/workflows/CI-python.yml
xuke444: changed

raiwidgets/requirements-dev.txt
xuke444: I have to change pipeline to run inside this folder.
if we use responsibleai it will install from pypi, which will not include local changes of responsibleai

romanlutz: Isn't that how it should be? Because if you run a release of raiwidgets then it will depend on the responsibleai version from pypi, not the repo version. Also, changes to packages should be to one package only, and otherwise split over multiple PRs. Wdyt?

romanlutz (commented):
Can you add a note to CHANGES.md?

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/TreeViewRenderer/TreeViewRenderer.tsx
```diff
@@ -91,6 +93,10 @@ export class TreeViewRenderer extends React.PureComponent<
   ITreeViewRendererProps,
   ITreeViewRendererState
 > {
+  public static contextType = ModelAssessmentContext;
```
imatiach-msft, line 96: why do we have model assessment context stuff in TreeViewRenderer?  Just seems like a really bad composition of things.  This should be passed in through the props.

raiwidgets/raiwidgets/_input_processing.py
```diff
@@ -103,6 +103,36 @@ def _convert_to_string_list_dict(
     return result
 
 
+def _safe_serializer(o):
+    """
```
imatiach-msft, line 107: why can't the method below be used?  It looks like there is duplicate code with _serialize_json_safe

imatiach-msft (approved):

raiwidgets/raiwidgets/_input_processing.py
xuke444: line 133 is the key, we need to return __dict__ in order for json.dump to serialize and class object.

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/TreeViewRenderer/TreeViewRenderer.tsx
xuke444: The purpose of context is to avoid multi level prop pass to make code simpler.
Our design is to keep top level props from sdk to context so that whenever any component needs to reference we don't have to pass props though multi layer.
ModelAssesmentContext is just a naming, it is used in interpret and error analysis dashboard already.

imatiach-msft: "The purpose of context is to avoid multi level prop pass to make code simpler."
sure, it makes it simpler, but it also makes it a lot more magical, harder to follow the code, and much less reusable.  To me it just seems extremely unexpected to see anything related to model assessment here.  If someone wanted to, they should be able to just take this component and reuse it in their code with the right properties passed in.  By having this context we are really making everything a lot more difficult from reusability point of view and encapsulation.

raiwidgets/raiwidgets/_input_processing.py
xuke444: refactored.

PR #567, opened 2021-05-29
change to use sub sample for model analysis

xuke444:

### DIFF
--- a/responsibleai/responsibleai/modelanalysis/model_analysis.py
+++ b/responsibleai/responsibleai/modelanalysis/model_analysis.py
@@ -16,8 +16,8 @@ from responsibleai._managers.explainer_manager import ExplainerManager
 
 
 _DTYPES = 'dtypes'
-_TRAIN = 'train'
-_TEST = 'test'
+_DATASET = 'dataset'
+_SUBSAMPLE = 'subsample'
 _TARGET_COLUMN = 'target_column'
 _TASK_TYPE = 'task_type'
 _MODEL = Metadata.MODEL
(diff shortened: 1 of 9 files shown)


responsibleai/responsibleai/modelanalysis/model_analysis.py
```diff
@@ -64,11 +41,12 @@ def __init__(self, model, train, test, target_column,
...
             or function that accepts a 2d ndarray.
         :type model: object
-        :param train: The training dataset including the label column.
-        :type train: pandas.DataFrame
-        :param test: The test dataset including the label column.
-        :type test: pandas.DataFrame
-        :param target_column: The name of the label column.
+        :param dataset: The dataset including the label column.
+        :type: pandas.DataFrame
+        :param subsample_indexes:
```
gregorybchris, line 46: I think we want to have two different subsamples of the full dataset. 

One subsample will be used as the evaluation set. This is the dataset one which we compute insights for the SDK. It can have greater than `MAX_VISUALIZATION_SAMPLES` samples (where this constant is 5000 right now). It is what SDK users will use to specify which points they are about explaining (also for causal local, counterfactual samples, etc)

On the other hand we need a separate subsampled dataset that is used for UI visualization. This subsample will have to be less than `MAX_VISUALIATION_SAMPLES` both for UI performance reasons and for usability so the UI is not cluttered with too many points.

These two datasets can both be subsampled from the full dataset and the full dataset can be used to train explainer surrogate models, DiCE models, and EconML DML models, but these two subsamples have different requirements and probably need to be represented separately.

My proposal would be to call one the `evaluation_index`->`evaluation_data` and one `visualization_index`->`visualization_data` (or `eval` and `viz` if you prefer less verbose names)

gregorybchris: If we want to start with one subsample dataset that can have >5k samples and raise a warning for UI performance as a first step, that's fine with me. But we should pay close attention to usage to check user experience when the number of samples is high.

gregorybchris (approved):

PR #580, opened 2021-06-07
Add policy tree to causal manager

gregorybchris:
EconML has added policy tree support to the CausalAnalysis class and we will include this in the test set output for RAI ModelAnalysis


responsibleai/tests/causal_manager_validator.py
```diff
@@ -57,34 +72,85 @@ def validate_causal(model_analysis, data, target_column):
         model_analysis.causal.compute()
 
 
+def _check_causal_results(causal_results):
+    assert isinstance(causal_results, dict)
+    print(causal_results.keys())
```
imatiach-msft, line 77: remove print?

gregorybchris: Nice catch, thanks

PR #587, opened 2021-06-09
downgrade flask version and expose stop function

xuke444:


romanlutz (approved):

imatiach-msft:
hmm I recall there was some issue with older version of flask, I would be careful and make sure to test this on different environments

xuke444:
> hmm I recall there was some issue with older version of flask, I would be careful and make sure to test this on different environments

tested in vm and aznb, works fine for 1.0.0

PR #600, opened 2021-06-11
Update model analysis notebook to use Boston housing dataset

gregorybchris:


xuke444 (approved):

xuke444:
counterfactual breaks

xuke444 (commented):
counterfactual breaks

imatiach-msft:
@gregorybchris I'm just wondering why are we doing this as opposed to adding another notebook for regression scenario?

gregorybchris:
> @gregorybchris I'm just wondering why are we doing this as opposed to adding another notebook for regression scenario?

That's a good point. I'll do that; can't hurt to have two notebooks

PR #601, opened 2021-06-11
remove error handling from counter factual

xuke444:

### DIFF
--- a/responsibleai/responsibleai/_managers/counterfactual_manager.py
+++ b/responsibleai/responsibleai/_managers/counterfactual_manager.py
@@ -185,48 +185,31 @@ class CounterfactualManager(BaseManager):
         for cf_config in self._counterfactual_config_list:
             if not cf_config.is_computed:
                 cf_config.is_computed = True
-                try:
-                    dice_explainer = self._create_diceml_explainer(
-                        method=cf_config.method,
-                        continuous_features=cf_config.continuous_features)
+                dice_explainer = self._create_diceml_explainer(
+                    method=cf_config.method,
+                    continuous_features=cf_config.continuous_features)
 
-                    X_test = self._test.drop([self._target_column], axis=1)
+                X_test = self._test.drop([self._target_column], axis=1)
 
-                    counterfactual_obj = \
-                        dice_explainer.generate_counterfactuals(
-                            X_test, total_CFs=cf_config.total_CFs,
-                            desired_class=cf_config.desired_class,
-                            desired_range=cf_config.desired_range)
+                counterfactual_obj = \
+                    dice_explainer.generate_counterfactuals(
+                        X_test, total_CFs=cf_config.total_CFs,
+                        desired_class=cf_config.desired_class,
+                        desired_range=cf_config.desired_range)
 
-                    cf_config.counterfactual_obj = \
-                        counterfactual_obj
-                except Exception as e:
-                    cf_config.has_computation_failed = True
-                    cf_config.failure_reason = str(e)
+                cf_config.counterfactual_obj = \
+                    counterfactual_obj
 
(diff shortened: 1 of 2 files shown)


responsibleai/responsibleai/_managers/counterfactual_manager.py
```diff
@@ -185,48 +185,31 @@ def compute(self):
...
-                except Exception as e:
-                    cf_config.has_computation_failed = True
-                    cf_config.failure_reason = str(e)
-
-    def get(self, failed_to_compute=False):
-        """Return the computed counterfactual examples objects or failure reason.
-
-        :param failed_to_compute: Get the failure reasons if counterfactual
-                                  examples failed to compute.
-        :type failed_to_compute: bool
```
gaugup, line 212: Why is the error handling removed? This handles the error scenario from dice-ml library and records the reason within the counterfactual manager.

responsibleai/responsibleai/_managers/counterfactual_manager.py
```diff
@@ -185,48 +185,31 @@ def compute(self):
...
-                        counterfactual_config.counterfactual_obj)
-            return counterfactual_obj_list
-        else:
-            failure_reason_list = []
-            for counterfactual_config in self._counterfactual_config_list:
-                if counterfactual_config.is_computed and \
-                        counterfactual_config.has_computation_failed:
-                    failure_reason_list.append(
-                        counterfactual_config.failure_reason)
-            return failure_reason_list
```
gaugup, line 229: The get() method returns the counterfactual object or the reson why the computation failed. Why this removed?

gaugup (changes requested):

responsibleai/responsibleai/_managers/counterfactual_manager.py
xuke444: none of other manger is doing this.
This way is pretty confused to end user
We should line up with other manger throw error during compute.

responsibleai/responsibleai/_managers/counterfactual_manager.py
xuke444: This is the issue all the error get hidden with this get, user will not notice when the counter factual fail

responsibleai/responsibleai/_managers/counterfactual_manager.py
gaugup: Sorry, what is confusing to the other users? We can add a warning to signal to the user that the computation failed. 

If the other managers don't handle the error handling , they should address that issue.

responsibleai/responsibleai/_managers/counterfactual_manager.py
gaugup: As I mentioned in the above comment, we could throw a warning that the computation failed as some reason. The error is not hidden. The error message is recorded in the class and user can access this error message. I am not able to understand in what capacity in the user not able to know that the counterfactual computation failed.

gaugup (changes requested):
In my opinion, this PR removes the meaningful error handling during counterfactual examples and shouldn't me merged.

gregorybchris:
I'm personally in favor of removing error handling like this. It's not very obvious to users how to retrieve error messages from failed analyses. In Python it is more common to raise an exception and let the caller handle the error than have a two step process of checking for a null result and then fetching the error for the failed computation.

gaugup:
> I'm personally in favor of removing error handling like this. It's not very obvious to users how to retrieve error messages from failed analyses. In Python it is more common to raise an exception and let the caller handle the error than have a two step process of checking for a null result and then fetching the error for the failed computation.

Why wasn't this a concern in the PR where the error handling was added https://github.com/microsoft/responsible-ai-widgets/pull/460?

PR #605, opened 2021-06-13
various minor fixes for regression error analysis scenario

imatiach-msft:
- fix root all data cohort to display correct metric
- fix cohort list/cohort info to show metric instead of error rate, also change it to use proper localization
- add new superconductor example notebook for regression scenario
- allow user to specify the max depth and num leaves for the error surrogate model in the widget constructor


xuke444 (approved):

xuke444:
approve for now.
As we discussed, once we load pre-calculate treeview for non-static view, we can remove rootStats and read from root node.

PR #620, opened 2021-06-21
Add feature importance to counterfactual manager

gaugup:
- Adds feature_importance flag to add() function of counterfactual manager.
- Adds tests for testing feature importance and counterfactual generation.

### DIFF
--- a/responsibleai/responsibleai/_managers/counterfactual_manager.py
+++ b/responsibleai/responsibleai/_managers/counterfactual_manager.py
@@ -144,7 +144,7 @@ class CounterfactualManager(BaseManager):
             desired_range=None,
             permitted_range=None,
             features_to_vary=None,
-            feature_importance=False):
+            feature_importance=True):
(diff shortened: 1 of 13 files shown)


responsibleai/responsibleai/_managers/counterfactual_manager.py
```diff
@@ -29,7 +29,7 @@ class CounterfactualConfig(BaseConfig):
     def __init__(self, method, continuous_features, total_CFs,
                  desired_class=CounterfactualConstants.OPPOSITE,
                  desired_range=None, permitted_range=None,
-                 features_to_vary=None):
+                 features_to_vary=None, feature_importance=False):
```
xuke444, line 32: Should we default to true?
I think PM want this.

gaugup: good point. Setting default to True.

xuke444 (approved):

PR #622, opened 2021-06-22
Add serialization for causal analysis

gregorybchris:
Changes made:
- Make serialize_json_safe public so AzureML can use it
- Created new DTOs for causal policy for the dashboard
- Updated causal manager params to accept latest from EconML
- Local policies returns a pandas dataframe like local and global effects
- Added tests for both getand get_data

### DIFF
--- a/raiwidgets/raiwidgets/dashboard.py
+++ b/raiwidgets/raiwidgets/dashboard.py
@@ -3,12 +3,14 @@
 
 """Defines the dashboard class."""
 
-from rai_core_flask import FlaskHelper  # , environment_detector
 import json
 import os
-from html.parser import HTMLParser
 import uuid
-from responsibleai._input_processing import _safe_object_serializer
+
+from html.parser import HTMLParser
+from rai_core_flask import FlaskHelper  # , environment_detector
+
+from responsibleai.serialization_utilities import serialize_json_safe
 
 
 class InLineScript(HTMLParser):
@@ -106,7 +108,7 @@ class Dashboard(object):
             content = content.replace(
                 "__rai_model_data__",
                 json.dumps(self.model_data,
-                           default=_safe_object_serializer))
+                           default=serialize_json_safe))
             return content
 
     def add_url_rule(self, func, route, methods):
--- a/responsibleai/responsibleai/_interfaces.py
+++ b/responsibleai/responsibleai/_interfaces.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 
 from enum import Enum
-from typing import List, Tuple
+from typing import Dict, List, Tuple, Union
 
 
 class Dataset:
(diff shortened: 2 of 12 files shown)


responsibleai/responsibleai/_interfaces.py
```diff
@@ -75,9 +75,36 @@ class CausalMetric:
...
+    leaf: bool
+    n_samples: int
+    treatment: str
+
+
+class CausalPolicyTreeInternal:
+    leaf: bool
+    feature: str
+    threshold: Union[float, str]  # TODO: Categorical features
+    left: Union['CausalPolicyTreeInternal', CausalPolicyTreeLeaf]
```
imatiach-msft, line 93: nit: maybe make this string a constant

raiwidgets/raiwidgets/dashboard.py
```diff
@@ -106,7 +108,7 @@ def load_widget_file(self, path):
             content = content.replace(
                 "__rai_model_data__",
                 json.dumps(self.model_data,
-                           default=_safe_object_serializer))
+                           default=serialize_json_safe))
```
imatiach-msft, line 111: this doesn't look right, it doesn't set forObject=True

responsibleai/responsibleai/serialization_utilities.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+import datetime
+import numpy as np
+
+from typing import Any
+
+
+def serialize_json_safe(o: Any):
```
imatiach-msft, line 10: this looks like:
C:\responsible-ai-widgets\responsibleai\responsibleai\_input_processing.py
except it is missing forObject parameter which Ke used.  Why did we decide to move it to a separate file and modify it like this?

imatiach-msft (approved):
approved with some comments

responsibleai/responsibleai/serialization_utilities.py
gaugup: There is a copy of this function in interpret-community as well. https://github.com/interpretml/interpret-community/blob/c6fe5c2436f2316bcf976acd45d827aeb66d99f4/python/interpret_community/common/serialization_utils.py#L11. We can import the function from interpret-community instead. Why are we having so copies of the same fuinction..

imatiach-msft: I think the reason originally was that no one depended on anyone else -- raiwidgets didn't depend on interpret-community and interpret-community didn't depend on raiwidgets

responsibleai/responsibleai/_interfaces.py
gregorybchris: We won't be using this anywhere else as it is a type hint for a recursive class definition. Normally I'd agree with you about constants because you never know when you'll want to reuse them! However, this is a rare case where reuse is highly unlikely.

responsibleai/responsibleai/serialization_utilities.py
gregorybchris: It needs to be made public for use in other packages. As discussed offline I'll update to handle the `__dict__` differently.

xuke444 (approved):

responsibleai/responsibleai/_managers/causal_manager.py
```diff
@@ -185,21 +214,24 @@ def compute(self):
...
                 alpha=config.alpha)
-            X_test = self._test.drop([self._target_column], axis=1)
             config.local_effects = analysis.local_causal_effect(
                 X_test, alpha=config.alpha)
 
             config.policies = []
             if config.treatment_features is not None:
                 for treatment_feature in config.treatment_features:
-                    local_policies = analysis._individualized_policy_dict(
+                    # TODO: Update to dataframe
```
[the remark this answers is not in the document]
gregorybchris: Yes! Fixed, thanks

responsibleai/responsibleai/serialization_utilities.py
gregorybchris: Fixed

raiwidgets/raiwidgets/dashboard.py
gregorybchris: Discussed offline new design for serialization

PR #624, opened 2021-06-22
[WIP] Propagate categoricals to MimicExplainer in ExplainManager

gaugup:
LightGBM surrogate model seems to handle string categoricals. We can leverage that to expand Explain Manager to handle string categoricals.


gaugup:
The revised version of this PR is at https://github.com/microsoft/responsible-ai-widgets/pull/629

PR #629, opened 2021-06-23
Propagate string categoricals to MimicExplainer in explainer manager

gaugup:
- Propagates categorical_features into LightGBM surrogate model
- Adds tests with complex model pipeline containing a transformer and estimator
- Skipping data check in validate_model_analysis() because of bug in dice-ml where dice-ml seems to be changing the metdata of the pandas dataframe. This is being fixed under https://github.com/interpretml/DiCE/pull/167.


responsibleai/tests/common_utils.py
```diff
@@ -108,6 +108,51 @@ def create_boston_data():
...
+    data_train, data_test, y_train, y_test = train_test_split(
+        dataset, target,
+        test_size=0.2, random_state=7, stratify=target)
+    return data_train, data_test, y_train, y_test, categorical_features, \
+        continuous_features, target_name, classes
+
+
+def create_complex_classification_pipeline(
+        X_train, y_train, continuous_features, categorical_features):
+    from sklearn.compose import ColumnTransformer
```
gregorybchris, line 131: Can these imports be at the module level?

responsibleai/tests/counterfactual_manager_validator.py
```diff
@@ -22,9 +22,16 @@ def verify_counterfactual_object(counterfactual_obj, feature_importance=False):
 
 
 def validate_counterfactual(cf_analyzer, X_train, target_column,
+                            categorical_features=None,
                             desired_class=None, desired_range=None,
                             feature_importance=False):
-    continuous_features = list(set(X_train.columns) - set([target_column]))
+
+    if categorical_features is None:
```
gregorybchris, line 29: This is fine, but another way to do this would be:
```
    categorical_features = set()

continuous features = ...
```
You still need to test both branches, but you duplicate less logic

gregorybchris (approved):

responsibleai/tests/common_utils.py
gaugup: Done.

responsibleai/tests/counterfactual_manager_validator.py
gaugup: The tests are for both cases. Don't want to create an empty set if not needed.

gregorybchris: Sure, you're trading off the small memory overhead of creating an empty set vs a small amount of code duplication. Probably just a matter of preference

imatiach-msft (approved):

PR #631, opened 2021-06-24
Add dependency test for graphiz

gaugup:


gregorybchris (approved):
I would propose this way of separating data from logic: https://github.com/microsoft/responsible-ai-widgets/pull/632/files

gaugup:
> I would propose this way of separating data from logic: https://github.com/microsoft/responsible-ai-widgets/pull/632/files
> 
> But please go ahead and check in if you like your way better

We should skip the test rather than just returning. At least someone would know clearly why this test was skipped.

gaugup:
> > I would propose this way of separating data from logic: https://github.com/microsoft/responsible-ai-widgets/pull/632/files
> > But please go ahead and check in if you like your way better
> 
> We should skip the test rather than just returning. At least someone would know clearly why this test was skipped.

Merging this for now.

PR #641, opened 2021-06-29
Remove deferred compute architecture for causal and counterfactual

gregorybchris:
Changes `add()/compute()` design of managers to be `compute()` only. The computed insights will still be saved to the managers for `get()` and `list()` to work as previously designed.

### DIFF
--- a/notebooks/model-analysis.ipynb
+++ b/notebooks/model-analysis.ipynb
@@ -81,11 +81,10 @@
     "\n",
     "# Queue Responsible AI insights\n",
     "model_analysis.explainer.add()\n",
-    "model_analysis.counterfactual.add(10, desired_class='opposite')\n",
+    "model_analysis.counterfactual.compute(10, desired_class='opposite')\n",
     "model_analysis.error_analysis.add()\n",
-    "model_analysis.causal.add(treatment_features=['Hours per week'])\n",
+    "model_analysis.causal.compute(treatment_features=['Hours per week'])\n",
     "\n",
-    "# Compute insights\n",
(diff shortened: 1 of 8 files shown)


imatiach-msft:
@gregorybchris I thought we weren't going to do this for the initial release?  Didn't we want to discuss this more with PM, or do we have agreement on this revised API now?  Even if we have agreement, shouldn't we wait until we finish the end to end notebook, and then we can plan/discuss the next steps?

gregorybchris:
> @gregorybchris I thought we weren't going to do this for the initial release? Didn't we want to discuss this more with PM, or do we have agreement on this revised API now? Even if we have agreement, shouldn't we wait until we finish the end to end notebook, and then we can plan/discuss the next steps?

I got PM approval on the design doc (both of them) and it seemed from the chat like your concerns had been addressed. I'm open to holding off, but not sure I totally understand your concern.

imatiach-msft:
@gregorybchris I guess my concern is we are making this API change too quickly, and there are a lot of moving parts already.  We want to hit the tight deadline and this is adding extra work that might cause it to slip.

imatiach-msft:
@gregorybchris we discussed and it seems like this is an important change to get in sooner rather than later

gaugup:
> @gregorybchris I guess my concern is we are making this API change too quickly, and there are a lot of moving parts already. We want to hit the tight deadline and this is adding extra work that might cause it to slip.

This is a major change. I fear that other managers may not align by June 30th. This change even though approved by PMs should have been made earlier.

PR #650, opened 2021-06-30
Add validations for input parameters for ModelAnalysis class

gaugup:
- Add basic validations for various input parameters for ModelAnalysis class
- Add tests for the above validations.


responsibleai/responsibleai/modelanalysis/model_analysis.py
```diff
@@ -117,6 +133,84 @@ def __init__(self, model, train, test, target_column,
...
+                small_train_data = train.iloc[0:2].drop(
+                    [target_column], axis=1)
+                small_test_data = test.iloc[0:2].drop([target_column], axis=1)
+                model.predict(small_train_data)
+                model.predict(small_test_data)
+            except Exception:
+                raise UserConfigValidationException(
+                    'The model passed cannot be used for getting predictions'
+                )
+
```
xuke444, line 213: this validation probably better in model-analysis.py, since that is where user input values.
And can benifit all components.

gaugup: The validations are the first thing that happen in the ModelAnalysis class __init__() function. All managers would benefit from these validations.

xuke444 (approved):

responsibleai/responsibleai/modelanalysis/model_analysis.py
```diff
@@ -117,6 +133,120 @@ def __init__(self, model, train, test, target_column,
                           self._error_analysis_manager,
                           self._explainer_manager]
 
+    def _validate_model_analysis_input_parameters(
```
gregorybchris, line 136: What do you think about starting a new module for validation? It might be easier to test if we can break validation up into smaller functions and that's hard to do all within one class.

responsibleai/responsibleai/modelanalysis/model_analysis.py
```diff
@@ -117,6 +133,120 @@ def __init__(self, model, train, test, target_column,
...
+                # Run predict of the model
+                try:
+                    small_train_data = train.iloc[0:1].drop(
+                        [target_column], axis=1)
+                    small_test_data = test.iloc[0:1].drop(
+                        [target_column], axis=1)
+                    model.predict(small_train_data)
+                    model.predict(small_test_data)
+                except Exception:
+                    raise UserConfigValidationException(
```
gregorybchris, line 245: When is this expected?

gregorybchris (approved):

responsibleai/responsibleai/modelanalysis/model_analysis.py
gregorybchris: Oh, just realized you had that originally.. why did you move it in the class?

responsibleai/responsibleai/modelanalysis/model_analysis.py
gaugup: Since almost RAI components rely on predcit()/predict_proba() so adding a check that the predict() works on the train and test data. I should get predeict_proba() as well.

responsibleai/responsibleai/modelanalysis/model_analysis.py
gaugup: Yes, initially I had it that way. But all the validations were around the ModelAnalysis class so moved it in this class. 

 We should think about more generic validations routines that can be used in multiple places. tIf you have other similar validations that need to be added, we can discuss about moving them into a new module.

responsibleai/responsibleai/modelanalysis/model_analysis.py
gaugup: Added validation for predcit_proba()

PR #660, opened 2021-07-08
Pin networkx to prevent installation of metplotlib

gaugup:
It seems new release of networkx brings in matplotlib which we don't want in responsibleai RAI insights computation engine. Hence pinning networkx.


responsibleai/requirements.txt
```diff
@@ -4,6 +4,7 @@ erroranalysis>=0.1.8
 interpret-community>=0.18.1
 ipykernel<6.0
 lightgbm>=2.0.11
+networkx==2.5
```
gregorybchris, line 7: Will you update this to a range `<=2.5` we want to continue to support all versions of `networkx` that do not take the `matplotlib` dependency.

gaugup: Done

gregorybchris (approved):

PR #669, opened 2021-07-12
Add unit test multi-class classification for causal

gaugup:
It seems the test was missing. Adding the test for completeness.

### DIFF
--- a/responsibleai/tests/causal_manager_validator.py
+++ b/responsibleai/tests/causal_manager_validator.py
@@ -4,6 +4,7 @@
 import pytest
 
 import pandas as pd
+import numpy as np
 
 from responsibleai.exceptions import (
     DuplicateManagerConfigException,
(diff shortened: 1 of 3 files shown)


gregorybchris (approved):

responsibleai/tests/test_model_analysis.py
```diff
@@ -51,6 +52,7 @@ def test_model_analysis_iris(self, manager_type):
         X_train[LABELS] = y_train
         X_test[LABELS] = y_test
         manager_args = {
+            ManagerParams.TREATMENT_FEATURES: [feature_names[0]],
```
gregorybchris, line 55: Personally I might avoid `feature_names[0]` in case the order of feature names changes, but it should be fine

gregorybchris (approved):

responsibleai/tests/causal_manager_validator.py
```diff
@@ -36,6 +37,16 @@
 
 def validate_causal(model_analysis, data, target_column,
                     treatment_features, max_cat_expansion):
+    if len(np.unique(data[target_column])) > 2:
+        with pytest.raises(AssertionError) as ae:
```
gregorybchris, line 41: You can use `pytest.raises(AssertionError, match="Multiclass classification isn't supported"):`

gaugup: Done.

responsibleai/tests/test_model_analysis.py
gaugup: It should be fine as the causal test is only for the assert condition.

PR #671, opened 2021-07-13
fix error analysis regression case for 1d heatmap matrix metrics being calculated incorrectly

imatiach-msft:
- fix error analysis regression case for 1d heatmap matrix metrics being calculated incorrectly
- add test for regression 1d heatmap case


imatiach-msft:
closing in favor of https://github.com/microsoft/responsible-ai-widgets/pull/672

PR #672, opened 2021-07-13
fix error analysis regression case for 1d heatmap matrix metrics being calculated incorrectly

imatiach-msft:
- fix error analysis regression case for 1d heatmap matrix metrics being calculated incorrectly
- add test for regression 1d heatmap case

### DIFF
--- a/erroranalysis/erroranalysis/_internal/matrix_filter.py
+++ b/erroranalysis/erroranalysis/_internal/matrix_filter.py
@@ -3,6 +3,7 @@
 
 import numpy as np
 import pandas as pd
+import math
 from erroranalysis._internal.cohort_filter import filter_from_cohort
 from erroranalysis._internal.constants import (PRED_Y,
                                                TRUE_Y,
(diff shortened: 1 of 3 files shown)


erroranalysis/erroranalysis/_internal/matrix_filter.py
```diff
@@ -178,21 +179,29 @@ def compute_json_matrix(analyzer, features, filters, composite_filters):
...
+            cut_err = df_err
+        # Compute the given metric for each group, if not using error rate
+        if metric != Metrics.ERROR_RATE:
+            aggfunc = _AggFunc(metric_to_func[metric])
+            cutdf_err = pd.DataFrame(cut_err)
+            cutdf_err['metric_values'] = list(zip(df_err[TRUE_Y],
+                                                  df_err[PRED_Y]))
+            grouped = cutdf_err.groupby([feat1])
+            counts_err = grouped.agg(aggfunc._agg_func_triplet)
+            counts_err = counts_err.values.flatten()
```
gregorybchris, line 202: I'm not super familiar with this code, but what do you think about using `ravel` to use less memory? https://stackoverflow.com/questions/28930465/what-is-the-difference-between-flatten-and-ravel-functions-in-numpy

gregorybchris (approved):

erroranalysis/erroranalysis/_internal/matrix_filter.py
imatiach-msft: done!

PR #682, opened 2021-07-14
remove ipykernel dependency from erroranalysis

romanlutz:


imatiach-msft (approved):

romanlutz:
Force-pushing since the broken gates test dependencies of a published pypi package. We need to release to fix that.

PR #686, opened 2021-07-14
Error Analysis: Raise warning which older version pandas with feature-names containing '-'

gaugup:
Signed-off-by: gaugup <gaugup@microsoft.com>

### DIFF
--- a/erroranalysis/erroranalysis/_internal/version_checker.py
+++ b/erroranalysis/erroranalysis/_internal/version_checker.py
@@ -1,0 +1,15 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+import pandas as pd
+from warnings import warn
+
+
+def check_pandas_version(feature_names):
+    if pd.__version__[0] == '0':
+        for feature_name in feature_names:
+            if '-' in feature_name:
+                message = 'The feature name ' + feature_name + \
+                    ' contains `-` which has issues with pandas version ' + \
+                    pd.__version__ + '. Please upgrade your pandas to 1.x'
+                warn(message)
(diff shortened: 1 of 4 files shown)


erroranalysis/erroranalysis/_internal/version_checker.py
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+import pandas as pd
+from warnings import warn
+
+
+def check_pandas_version(feature_names):
```
imatiach-msft, line 8: maybe add some method doc, even though it's under internal?

erroranalysis/tests/test_error_report.py
```diff
@@ -53,11 +56,19 @@ def is_valid_uuid(id):
 
 
 def run_error_analyzer(model, X_test, y_test, feature_names,
-                       categorical_features):
-    model_analyzer = ModelAnalyzer(model, X_test, y_test,
-                                   feature_names,
-                                   categorical_features)
-    error_report1 = model_analyzer.create_error_report()
+                       categorical_features, expect_user_warnings=False):
```
imatiach-msft, line 59: I guess this expect_user_warnings parameter is kind of a defunct parameter because we can't really test with an older version of pandas?

imatiach-msft (approved):

erroranalysis/erroranalysis/_internal/version_checker.py
gaugup: Done

erroranalysis/tests/test_error_report.py
gaugup: It is not a defunct parameter per say. This is trying to test for warning only if feature names have - and pandas version is less than 1.0.

gaugup: using this param. Thanks for pointing this out.

PR #689, opened 2021-07-15
fix search functionality in features list

imatiach-msft:
fixes 1.) search functionality not updating list of features (not working) 2.) random synchronization issues by calling updateSelection after setState completes 3.) when changing selected features that are searched for, previously selected features should not get wiped out

### DIFF
--- a/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/FeatureList/FeatureList.tsx
+++ b/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/FeatureList/FeatureList.tsx
@@ -107,7 +107,14 @@ export class FeatureList extends React.Component<
     this.layerHostId = getId("featuresListHost");
     this._selection = new Selection({
       onSelectionChanged: (): void => {
-        const newSelectedFeatures = this.getSelectionDetails();
+        let newSelectedFeatures = this.getSelectionDetails();
+        const oldSelectedFeaturesNotSearched = this.state.selectedFeatures.filter(
+          (oldSelectedFeature) =>
+            !this.state.searchedFeatures.includes(oldSelectedFeature)
+        );
+        newSelectedFeatures = newSelectedFeatures.concat(
+          oldSelectedFeaturesNotSearched
+        );
         const enableApplyButton =
           this.state.lastAppliedFeatures.size !== newSelectedFeatures.length ||
           newSelectedFeatures.some(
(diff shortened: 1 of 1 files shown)


libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/FeatureList/FeatureList.tsx
```diff
@@ -123,8 +130,14 @@ export class FeatureList extends React.Component<
     this.updateSelection();
   }
 
-  public componentDidUpdate(prevProps: IFeatureListProps): void {
-    if (this.props.importances !== prevProps.importances) {
+  public componentDidUpdate(
+    prevProps: IFeatureListProps,
+    prevState: IFeatureListState
```
xuke444, line 135: we should not monitor state in component did update, this is circular refreence.
Can we do action directly when we set state?

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/FeatureList/FeatureList.tsx
```diff
@@ -107,7 +107,14 @@ export class FeatureList extends React.Component<
...
     this._selection = new Selection({
       onSelectionChanged: (): void => {
-        const newSelectedFeatures = this.getSelectionDetails();
+        let newSelectedFeatures = this.getSelectionDetails();
+        const oldSelectedFeaturesNotSearched = this.state.selectedFeatures.filter(
+          (oldSelectedFeature) =>
+            !this.state.searchedFeatures.includes(oldSelectedFeature)
+        );
+        newSelectedFeatures = newSelectedFeatures.concat(
+          oldSelectedFeaturesNotSearched
```
xuke444, line 116: why do we need to concat non-searched feature?

xuke444 (approved):

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/FeatureList/FeatureList.tsx
imatiach-msft: so the selection is only for the active/searched items in the features list.  If there were items already selected before search, and this logic is removed, suddenly they get unselected when a user changes selection for a searched feature.

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/FeatureList/FeatureList.tsx
imatiach-msft: done!

xuke444 (approved):

PR #697, opened 2021-07-19
Add user feedback to compute() function

gaugup:
-Also add the time taken in each compute() function.

The output looks like below:-

image


riedgar-ms:
Weren't we planning to get rid of the `compute()` functions?

responsibleai/responsibleai/_managers/error_analysis_manager.py
```diff
@@ -202,9 +202,12 @@ def add(self, max_depth=3, num_leaves=31, filter_features=None):
         else:
             self._ea_config_list.append(ea_config)
 
+    @measure_time
     def compute(self):
         """Creates an ErrorReport by running the error analyzer on the model.
         """
+        print("Error Analysis")
```
imatiach-msft, line 209: I don't think we should be adding prints?  maybe log.info or something, but not prints?

gaugup: The intension was to give users feedback as what computation is happening. Hence, added the prints.

The notebook output looks like this now:-

image

gregorybchris: Looks really good. But I also wonder if maybe that can be off by default and have some way to turn it on from a notebook?

gaugup:
> Weren't we planning to get rid of the `compute()` functions?

With or without comoute() this prints some useful information time taken. We can add more things to the decorator.

responsibleai/responsibleai/_managers/error_analysis_manager.py
gaugup: I have it on by default. If we want to give user feedback we should leave it on by default.

gregorybchris: If a user calls this code in a script and they don't want to see the output can they turn it off? (Python SDKs usually avoid printing to stdout)

gregorybchris: That would be awesome. Maybe consider putting a `verbose`/`show_progress` flag in the constructor since we plan to remove the shared `compute()`?

Personally the way I'd do it is to implement a new `self._log()` method on `ModelAnalysis` that does the check for `self._verbose` to avoid repeating the `if self._verbose` check. But that's just an idea.

imatiach-msft (approved):

responsibleai/responsibleai/_managers/error_analysis_manager.py
gaugup: closing this thread as deciding to just print the status.

PR #699, opened 2021-07-19
Explicitly pass model task into MimicExplainer

gaugup:


responsibleai/responsibleai/_managers/explainer_manager.py
```diff
@@ -119,7 +119,10 @@ def compute(self):
             return
         if self._is_run:
             return
-        model_task = ModelTask.Unknown
+        if self._classes is not None:
```
imatiach-msft, line 122: this isn't quite right I think, the classes are optional and aren't needed, if not specified I think we use the numeric values but the task can still be classification.  I think we can just pass the model task from model analysis to the explainer manager as a parameter?

imatiach-msft: oh, it seems you are setting the classes explicitly always in model analysis, then this is fine

imatiach-msft (approved):

responsibleai/responsibleai/_managers/explainer_manager.py
gaugup: Doesn't the code change in model_analysis.py makes sure that classes is always set in case of classification scenario?

imatiach-msft: Yes, it seems fine.  I was thinking from explanations API perspective where this can be set optionally by the user.  But here it seems user isn't passing classes but rather model analysis is setting it now.  Actually I think this is a bit inconvenient, since user has to modify their labels column if they want to see it appear as strings in the dashboard, but it's fine.

PR #702, opened 2021-07-19
Save and load for causal manager

gregorybchris:
This PR updated the causal manager in model analysis to enable save and load of full causal results:
- Causal analysis now returns a `CausalResult` object from a call to `get()` rather than a dictionary. `CausalResult` has the ability to serialize and deserialize itself from the file system.
- Created shared utilities for serializing attributes on an object. While JSON serialization is useful for objects that can be fully JSON serialized, these new utilities offer support for pickle file serialization which can be used for ML models.
- Update UI expected format of causal results so `treatment_features` are part of the config object.
- The computation of the causal insights happens in the call to `add()` rather than the call to `compute()` and the `compute()` method is a no-op (will be maintained until all components deprecate `compute()`)

Additional (small) bug fixes:
- Passing heterogeneity model to EconML now
- Updated default max policy tree depth from 3 to 2 for UI compatibility


responsibleai/responsibleai/_tools/causal/causal_manager.py
```diff
@@ -0,0 +1,255 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
```
imatiach-msft, line 2: why was it moved to _tools?

gregorybchris: @imatiach-msft I was trying out this new file structure. I personally like the idea of keeping abstract/base classes in one place and component-specific code in component-specific folders.

The benefit is that it's really clear we're breaking abstraction boundaries when component specific code goes into base classes. The most composable design will be one where the base classes reference each component only once in order to "register" them as valid managers.

I'm happy to change it back if you want to keep the current file structure.

gregorybchris: For the sake of getting the PR in faster I've moved it back.

responsibleai/responsibleai/_managers/causal_manager.py
```diff
@@ -192,171 +93,121 @@ def add(
...
+            ResultAttributes.LOCAL_POLICIES: local_policies,
+            ResultAttributes.POLICY_GAINS: {
+                ResultAttributes.RECOMMENDED_POLICY_GAINS:
+                    tree.policy_value,
+                ResultAttributes.TREATMENT_GAINS: tree.always_treat,
+            },
+            ResultAttributes.POLICY_TREE: tree.tree_dictionary
+        }
 
     def compute(self):
```
xuke444, line 194: have we decided to remove compute?

responsibleai/responsibleai/_managers/causal_manager.py
```diff
@@ -367,8 +218,44 @@ def name(self):
...
+        :param model_analysis: The loaded parent ModelAnalysis.
+        :type model_analysis: ModelAnalysis
+        """
+        this = cls.__new__(cls)
+        causal_dir = Path(path)
+
+        # Rehydrate results
+        results_path = causal_dir / SerializationAttributes.RESULTS
+        paths = results_path.resolve().glob('*')
+        this.__dict__['_results'] = [CausalResult.load(p) for p in paths]
```
xuke444, line 251: should we have load function in causal result to return dict?
So both save and load stay in same file/place for better maintanace in future?

responsibleai/responsibleai/_managers/causal_manager.py
gregorybchris: Yes, but I will not do it in this PR

xuke444 (approved):

PR #705, opened 2021-07-20
Update default max_tree_depth to 2

gregorybchris:


imatiach-msft (approved):

responsibleai/responsibleai/_managers/causal_manager.py
```diff
@@ -28,7 +28,7 @@ class CausalConstants:
     DEFAULT_MAX_CAT_EXPANSION = 50
     DEFAULT_TREATMENT_COST = 0
     DEFAULT_MIN_TREE_LEAF_SAMPLES = 2
-    DEFAULT_MAX_TREE_DEPTH = 3
+    DEFAULT_MAX_TREE_DEPTH = 2
```
gaugup, line 31: Is there a test checking this change?

gregorybchris: How would you want this to be tested?

gaugup: Maybe check the ploicy tree dimensions, somehow. I am not an expert on the causal. But it is strange that there is no test covering this scenario,

gregorybchris: Good point. I think I'll just add these tests to my other causal PR to speed things up. Waiting for gates to pass is taking too long

PR #712, opened 2021-07-21
fix calculating feature importances after error report due to dataframe on analyzer being modified

imatiach-msft:
While debugging another issue in model analysis workflow, I discovered a bug where feature importances failed to calculate.  This was because in model analysis we calculate tree view/matrix filter on error report first, and that modified the original dataset.  Making a non-deep copy (since we just add columns but not change the original data) resolved the issue.

### DIFF
--- a/erroranalysis/tests/test_error_report.py
+++ b/erroranalysis/tests/test_error_report.py
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation
 # Licensed under the MIT License.
 
+import numpy as np
 import pandas as pd
 import pytest
 import uuid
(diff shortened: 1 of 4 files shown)


erroranalysis/tests/test_error_report.py
```diff
@@ -49,6 +50,18 @@ def test_error_report_boston(self):
             run_error_analyzer(model, X_test, y_test, feature_names,
                                categorical_features)
 
+    def test_error_report_boston_pandas(self):
+        X_train, X_test, y_train, y_test, feature_names = \
+            create_boston_data()
+        X_train = pd.DataFrame(X_train, columns=feature_names)
+        X_test = pd.DataFrame(X_test, columns=feature_names)
```
gaugup, line 57: Can you move lines 65 and 57 into file common_utils.py?

gregorybchris (approved):

gaugup (approved):

erroranalysis/tests/test_error_report.py
imatiach-msft: done!

romanlutz (approved):

romanlutz:
I'm merging this since notebooks in main are not working properly without it. If there was anything else missing please send a follow-up and I'll review it asap.

PR #715, opened 2021-07-22
add sample_dataset parameter to error analysis to allow it to run on much larger data when no explanation exists

imatiach-msft:
Add sample_dataset parameter.  This is in scenario where no explanation is specified (otherwise, the sampled dataset on the explanation is used).  The sampled dataset is passed to the UX and shown in the datasets tab, but the full dataset (specified with "dataset" parameter) is used for the tree view and heatmap.

### DIFF
--- a/raiwidgets/raiwidgets/error_analysis_dashboard.py
+++ b/raiwidgets/raiwidgets/error_analysis_dashboard.py
@@ -29,7 +29,7 @@ class ErrorAnalysisDashboard(Dashboard):
         (# examples x # features), the same samples used to build the
         explanation. Overwrites any existing dataset on the
         explanation object.
-    :type dataset: numpy.ndarray or list[][]
+    :type dataset: pd.DataFrame or numpy.ndarray or list[][]
     :param true_y: The true labels for the provided explanation. Overwrites
         any existing dataset on the explanation object.
         Note if explanation is sample of dataset, you will need to specify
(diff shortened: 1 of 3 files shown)


raiwidgets/raiwidgets/error_analysis_dashboard_input.py
```diff
@@ -37,7 +37,8 @@ def __init__(
             model_task,
             metric,
             max_depth,
-            num_leaves):
+            num_leaves,
+            sample_dataset):
```
romanlutz, line 41: Should this be None by default? Just wondering because of the logic below

imatiach-msft: this is the dashboard input, none of the parameters are optional, the user-facing API in error_analysis_dashboard.py has the optional parameters

raiwidgets/raiwidgets/error_analysis_dashboard.py
```diff
@@ -73,6 +73,10 @@ class ErrorAnalysisDashboard(Dashboard):
     :param num_leaves: The number of leaves of the surrogate tree
         trained on errors.
     :type num_leaves: int
+    :param sample_dataset: Used only if explanation is not specified for
```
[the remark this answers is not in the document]
imatiach-msft: you can specify an explanation which has a dataset tied to it.  We by default use that downsampled explanation in the 'explanation' tabs when hitting the explanation button in the Error Analysis widget.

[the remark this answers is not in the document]
imatiach-msft: I added several examples to the docs as discussed

raiwidgets/raiwidgets/error_analysis_dashboard.py
```diff
@@ -73,15 +73,50 @@ class ErrorAnalysisDashboard(Dashboard):
...
+        the dataset explorer.  Specify less than 10k points for optimal
+        performance.
+    :type sample_dataset: pd.DataFrame or numpy.ndarray or list[][]
+
+    :Example:
```
romanlutz, line 81: Why is this both here and below? Shouldn't once be enough?

raiwidgets/raiwidgets/error_analysis_dashboard.py
```diff
@@ -121,6 +156,11 @@ def __init__(self, explanation=None, model=None, *, dataset=None,
             alternative to the model and explanation for a more limited
             view.
         :type pred_y: numpy.ndarray or list[]
+        :param pred_y_dataset: The predicted labels for the provided dataset.
```
gregorybchris, line 159: If we have `dataset` and `sample_dataset` should we also then have `y_pred` and `sample_y_pred` for consistency?

raiwidgets/raiwidgets/error_analysis_dashboard.py
```diff
@@ -73,15 +73,50 @@ class ErrorAnalysisDashboard(Dashboard):
     :param num_leaves: The number of leaves of the surrogate tree
         trained on errors.
     :type num_leaves: int
+    :param sample_dataset: Used only if explanation is not specified for
```
gregorybchris, line 76: Maybe this docstring can start with a description of the parameter? Then go on to say when it's used?

Something like: "Dataset with fewer samples than the main dataset. Used to improve performance only when an Explanation object is not provided."?

raiwidgets/raiwidgets/error_analysis_dashboard.py
imatiach-msft: We have true_y and true_y_dataset already, and this is following this with pred_y and pred_y_dataset.
This naming makes sense because in the common case a downsampled explanation is passed, we need the pred_y/true_y to follow the explanation but the pred_y_dataset/true_y_dataset to follow the dataset parameter.  However in the case there is no explanation passed, and user only passes in a dataset, we give the user the option to still pass in a sample_dataset for the dataset explorer only (similar to the dataset that is on the downsampled explanation).  So I think the current naming is best considering all of the possible different scenarios.

raiwidgets/raiwidgets/error_analysis_dashboard.py
imatiach-msft: done, improved the description!

raiwidgets/raiwidgets/error_analysis_dashboard.py
imatiach-msft: I thought one was needed for sphinx and the other for help():
https://erroranalysis.ai/raiwidgets.html
but it looks like the sphinx is fine and help is fine.  I'm not sure how you made the sphinx use the top-level class doc instead of the init method doc.  Maybe it's only a problem in the azureml-sdk documentation.  Let me remove this then.

imatiach-msft: Removed the doc below.

raiwidgets/raiwidgets/error_analysis_dashboard_input.py
```diff
@@ -93,13 +100,21 @@ class and for the regression case a method of predict()
         :param num_leaves: The number of leaves of the surrogate tree
             trained on errors.
         :type num_leaves: int
+        :param sample_dataset: Used only if explanation is not specified for
```
gregorybchris, line 103: Might want to update this docstring like you did the other one

gregorybchris (approved):

raiwidgets/raiwidgets/error_analysis_dashboard_input.py
imatiach-msft: done!

PR #721, opened 2021-07-23
Add missing documentation for counterfactual manager

gaugup:


gregorybchris (approved):

imatiach-msft (approved):

responsibleai/responsibleai/_managers/counterfactual_manager.py
```diff
@@ -156,6 +156,9 @@ def add(self,
             feature_importance=True):
         """Add a counterfactual generation configuration to be computed later.
 
+        :param method: Type of dice-ml explainer. Either of "random", "genetic"
```
romanlutz, line 159: Is "dice-ml explainer" a class or well-defined object? We could link to it to avoid making people track this down on their own.
```
:class:`dice.module.DiceMLExplainer`  # if it's a class in a module dice.module
`dice-ml explainer <https://a.link.to/thedescription.txt>`_
```
For the first option we'd need to add DICE docs to the intersphinx mapping (which we probably should do anyway...): https://github.com/microsoft/responsible-ai-widgets/blob/f93bf8f61d75bab1280efe249b59935b86a91075/docs/source/conf.py#L57

gaugup: I need to find the sphinx link for dice-ml. Not sure they have one. Will find this from the dice-ml maintainers.

gaugup: Inside the Cf manager, the dice-ml explainer is exposed under a private function. The add() function here is just a wrapper. 

Also, dice-ml has validations to guide users that what values need to supplied.

PR #722, opened 2021-07-24
add static tree to model analysis

xuke444:


zhb000 (approved):

romanlutz (approved):

libs/core-ui/src/lib/Interfaces/IErrorAnalysisData.ts
```diff
@@ -0,0 +1,79 @@
...
+  parentNodeName: string;
+  pathFromRoot: string;
+  size: number;
+  sourceRowKeyHash: string;
+  success: number;
+  badFeaturesRowCount: number;
+}
+
+export type IErrorAnalysisMatrixCategory =
+  | IErrorAnalysisMatrixClassificationCategory
```
imatiach-msft, line 34: this shouldn't be classification vs regression, but error rate vs mean squared error, as more metrics will be added

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/MatrixArea/MatrixAreaUtils.ts
```diff
@@ -152,14 +138,14 @@ export function createCohortStatsFromSelectedCells(
...
-  selectedFeature2: string,
+  jsonMatrix: IErrorAnalysisMatrix,
+  selectedFeature1: string | undefined,
+  selectedFeature2: string | undefined,
   baseCohort: ErrorCohort,
   features: string[]
 ): ICompositeFilter[] {
   const feature2IsSelected =
-    selectedFeature2 !== noFeature && selectedFeature2 !== selectedFeature1;
+    selectedFeature2 && selectedFeature2 !== selectedFeature1;
```
imatiach-msft, line 148: the user can actually select noFeature and user should only see one, I don't see where this check was moved to, but this would be a bug if the check was removed altogether

libs/model-assessment/src/lib/ModelAssessmentDashboard/AvailableTabs.ts
```diff
@@ -10,42 +9,26 @@ import { GlobalTabKeys } from "./ModelAssessmentEnums";
 
 export function getAvailableTabs(
   props: IModelAssessmentDashboardProps,
-  jointDataset: JointDataset,
   excludeErrorAnalysis: boolean
 ): IDropdownOption[] {
   const availableTabs: IDropdownOption[] = [];
-  if (
-    !excludeErrorAnalysis &&
-    props.requestDebugML &&
```
romanlutz, line 19: shouldn't we have at least one of:
-  these methods
- preprocessed data?
Now it looks like we're not checking at all.

libs/model-assessment/src/lib/ModelAssessmentDashboard/AvailableTabs.ts
```diff
@@ -10,42 +9,26 @@ import { GlobalTabKeys } from "./ModelAssessmentEnums";
...
-    });
-  }
-
-  if (jointDataset.hasDataset) {
-    availableTabs.push({
-      key: GlobalTabKeys.DataExplorerTab,
-      text: localization.ModelAssessment.ComponentNames.DataExplorer
-    });
-  }
+  availableTabs.push({
```
romanlutz, line 22: What if we don't have that data? Why are we removing this?

apps/dashboard/src/app/applications.ts
```diff
@@ -153,21 +153,21 @@ export const applications: IApplications = <const>{
...
       } as IModelAssessmentDataSet,
       adultCensusIncomeNoCounterfactualData: {
         causalAnalysisData: [adultCensusCausalAnalysisData],
         classDimension: 2,
         dataset: adultCensusWithFairnessDataset,
-        errorAnalysisConfig: [adultCensusCausalErrorAnalysisConfig],
+        errorAnalysisData: [adultCensusCausalErrorAnalysisData],
         modelExplanationData: [adultCensusWithFairnessModelExplanationData]
       } as IModelAssessmentDataSet,
       adultCensusIncomeNoModelData: {
```
romanlutz, line 173: This renders WITH model data when I click through the UI in the blob. Something is off here.

libs/model-assessment/src/lib/ModelAssessmentDashboard/AvailableTabs.ts
xuke444: Technically yes.
But that would not happen, errorAnalysisData only added with either requestDebugML or preprocesseddata
I added extra check for safty.

apps/dashboard/src/app/applications.ts
xuke444: I am confused

PR #728, opened 2021-07-27
Fix refresh on global cohort change

romanlutz:
Making EA and Dataset Explorer components non-pure to allow them to refresh on global cohort changes. Additionally, the dataset explorer code can have the selected cohort index get out of sync (this is a remnant of the code used in the EA app) so this PR resets it before rendering.


libs/dataset-explorer/src/lib/DatasetExplorerTab.tsx
```diff
@@ -94,6 +94,16 @@ export class DatasetExplorerTab extends React.PureComponent<
         </MissingParametersPlaceholder>
       );
     }
+
+    const selectedCohortIndex = this.context.errorCohorts.findIndex(
+      (errorCohort) =>
+        errorCohort.cohort.name === this.context.selectedErrorCohort.cohort.name
+    );
+    if (selectedCohortIndex !== this.state.selectedCohortIndex) {
+      this.setState({ selectedCohortIndex });
```
xuke444, line 103: usually we should not set state in render.
move this part to component did update should be better.

xuke444 (approved):

PR #729, opened 2021-07-27
add columns to dashboard dataframe settings

xuke444:


gaugup (approved):

raiwidgets/raiwidgets/model_analysis_dashboard_input.py
```diff
@@ -25,8 +25,8 @@ def __init__(
         self._is_classifier = model is not None\
             and hasattr(model, SKLearn.PREDICT_PROBA) and \
             model.predict_proba is not None
-        self._dataframeColumns = None
         self.dashboard_input = analysis.get_data()
+        self._dataframeColumns = self.dashboard_input.dataset.feature_names
```
gregorybchris, line 29: `_dataframeColumns` should be use snake case and we should avoid naming variables with the type "dataframe" like this. Maybe `_dataset_features`, `_dataset_feature_names`, or `_dataset_columns`?

raiwidgets/tests/test_model_analysis_dashboard_input.py
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+from sklearn.model_selection import train_test_split
+import shap
+import sklearn
+import mock
+from responsibleai import ModelAnalysis
+from raiwidgets.model_analysis_dashboard_input import \
```
gregorybchris, line 9: Will customers ever import `ModelAnalysisDashboardInput`? If not let's make the module private

xuke444: probably in separate pr, all other dashboards are like this as well.

PR #730, opened 2021-07-27
fix serialize_json_safe not escaping special characters which caused dashboard to fail to render on pandas dataframe that had double quotes in string values

imatiach-msft:
- fix for customer that had issue running error analysis dashboard on a pandas dataframe that had double quotes in it
- fixes serialize_json_safe not escaping special characters which caused dashboard to fail to render on pandas dataframe that had double quotes in string values
- also call serialize_json_safe on categorical map parameter passed in error analysis dashboard

### DIFF
--- a/responsibleai/tests/test_serialization_utilities.py
+++ b/responsibleai/tests/test_serialization_utilities.py
@@ -4,6 +4,7 @@
 import numpy as np
 
 from responsibleai.serialization_utilities import serialize_json_safe
+import json
 
 
 class TestSerializationUtilities:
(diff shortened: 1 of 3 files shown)


responsibleai/tests/test_serialization_utilities.py
```diff
@@ -28,3 +29,13 @@ def test_unknown(self):
         c = complex(1, 2)
         result = serialize_json_safe([c, 42])
         assert result == [c, 42]
+
+    def test_strings_with_special_chars(self):
```
gregorybchris, line 33: Maybe you could add a test that roundtrips this dictionary?
I think your test covers the scenario, but if someone had to update the JSON string below they could do it in a way that roundtripping technically doesn't work.
Like this:
dict -> str (or file)
str (or file) -> dict
assert dict1 == dict2

imatiach-msft: done!

gregorybchris (approved):

PR #733, opened 2021-07-28
add more filters to error analysis backend

imatiach-msft:
Added more filters to error analysis to support model analysis dashboard:
- less
- greater and equal

Added tests for all of the filtering methods (less, less and equal, equal, greater, greater and equal, in the range of, includes)

### DIFF
--- a/erroranalysis/erroranalysis/_internal/cohort_filter.py
+++ b/erroranalysis/erroranalysis/_internal/cohort_filter.py
@@ -13,7 +13,9 @@ from erroranalysis._internal.constants import (PRED_Y,
 
 METHOD_EQUAL = 'equal'
 METHOD_GREATER = 'greater'
+METHOD_LESS = 'less'
 METHOD_LESS_AND_EQUAL = 'less and equal'
+METHOD_GREATER_AND_EQUAL = 'greater and equal'
 METHOD_RANGE = 'in the range of'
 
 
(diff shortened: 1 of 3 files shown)


erroranalysis/erroranalysis/_internal/cohort_filter.py
```diff
@@ -53,8 +55,12 @@ def build_query(filters, categorical_features, categories):
             colname = filter['column']
             if method == METHOD_GREATER:
                 queries.append("`" + colname + "` > " + arg0)
+            elif method == METHOD_LESS:
+                queries.append("`" + colname + "` < " + arg0)
```
riedgar-ms, line 59: Should we be concerned with injection attacks here?

imatiach-msft: what's a scenario where that would be a problem here?  I can't think of one.

riedgar-ms: I'm not sure, but you're building a string which I'm expecting will be 'evaluated' for some definition of the term. This does look similar to the sort of thing which gets you SQL injection attacks.

Not saying it's a huge issue for us, but something to be aware. If we're sure that `arg0` (and for that matter `colname`) are always under the control of whoever runs this code, it's not going to be an issue.

imatiach-msft: Sure it's similar, but this is just a pandas query.  The pandas query can't actually run arbitrary code.  Also, this is only run on a user's machine.  I just can't think of a scenario where this would be a problem.
How would you fix this issue then?  And what scenario would this fix?

riedgar-ms: OK. It was just something I saw and immediately thought "injection attack" but I don't know any specific exploits.

riedgar-ms (approved):

PR #734, opened 2021-07-28
Fix aggregate feature importance chart for multiclass classification

romanlutz:
`activeSeries` is only referenced once, so we don't need to keep it in the state and instead calculate it on the fly when needed.

The aggregate in the test case have been calculated by me to ensure they're correct. Still, these need to be expanded to cover other scenarios in the future.


gaugup:
Is there a test we can add to this change?

imatiach-msft:
I wonder if we can add a test to validate the calculated global importance values in the UI against some golden values that we have calculated by hand and know are correct?

xuke444 (approved):

PR #742, opened 2021-07-30
Fix bug bash issues

zhb000:
causal table show number as Scientific format 

Individual causal treatment policy should have scroll when table gets tall enough

change in title of field in individualized policy


libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/TreatmentView/TreatmentStyles.ts
```diff
@@ -61,6 +62,10 @@ export const TreatmentStyles: () => IProcessedStyleSet<ITreatmentStyles> =
         padding: "10px",
         width: "70%"
       },
+      listContainer: {
+        maxHeight: "65vh",
```
xuke444, line 66: this may have issue in studio, can we use percent?

xuke444 (approved):

PR #744, opened 2021-07-31
Causal treatment costs per feature/sample/category

gregorybchris:
EconML provides the ability to specify a cost of treatment. The recommended policy will be based not just on an expected increase in the target but also the expected decrease in the target given the number of samples being treated (and which kind of treatment/how much treatment depending on discrete/continuous features).

Before this PR only a constant cost of treatment was allowed. This PR adds the ability for users of the responsibleai package to pass treatment costs:
- per treatment feature - it might cost more to provide personalized tech support than to offer a discounted upgrade
- per sample - it might cost more to provide tech support to an advanced user compared to a standard user
- per treatment category - it might cost more to provide tech support with an in-store employee compared to a call center employee 

Additional minor updates:
- improved type hints for the CausalManager
- improved docstrings for the CausalManager

### DIFF
--- a/responsibleai/responsibleai/_managers/causal_manager.py
+++ b/responsibleai/responsibleai/_managers/causal_manager.py
@@ -2,7 +2,10 @@
 # Licensed under the MIT License.
 
 """Manager for causal analysis."""
+import numpy as np
 import pandas as pd
+
+from typing import Any, List, Optional, Union
 
 from econml.solutions.causal_analysis import CausalAnalysis
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


gaugup:
Can you please provide some description on the scope of changes in this PR?

gregorybchris:
> Can you please provide some description on the scope of changes in this PR?

Yes, done. Hopefully that's clearer.

responsibleai/responsibleai/_managers/causal_manager.py
```diff
@@ -77,19 +89,25 @@ def add(
         :param upper_bound_on_cat_expansion: Maximum expansion for
                                              categorical features.
         :type upper_bound_on_cat_expansion: int
-        :param treatment_cost: Cost to treat one individual or
-                               per-individual costs as an array.
-        :type treatment_cost: float or array
+        :param treatment_cost: Cost of treatment. If 0 all treatments will
```
gaugup, line 92: Missing , after 0

responsibleai/responsibleai/_managers/causal_manager.py
```diff
@@ -182,13 +202,38 @@ def add(
...
+            message = ("treatment_cost must be a list with "
+                       "the same number of elements as "
+                       "treatment_features where each element "
+                       "is either a constant cost of treatment "
+                       "or an array specifying the cost of "
+                       "treatment per sample. "
+                       "Found treatment_cost of type "
+                       f"{type(treatment_cost)}, expected list.")
+            raise ValueError(message)
+        elif len(treatment_cost) != len(treatment_features):
```
gaugup, line 220: I there a minimum value for treatment cost? Like zero or something is minimum. Or there is no min max. If there is min and max, then we should check for those and raise UserConfigValidationException.

responsibleai/responsibleai/_managers/causal_manager.py
```diff
@@ -182,13 +202,38 @@ def add(
...
+            message = ("treatment_cost must be a list with "
+                       "the same number of elements as "
+                       "treatment_features. "
+                       "Length of treatment_cost was "
+                       f"{len(treatment_cost)}, expected "
+                       f"{len(treatment_features)}.")
+            raise ValueError(message)
```
gaugup, line 227: Could you maybe raise UserConfigValidationException instead of ValueError here?

gregorybchris: Yep, done!

responsibleai/responsibleai/_managers/causal_manager.py
gregorybchris: Nope, you can have a negative cost which would mean the default treatment is actually more expensive than the new treatment.

responsibleai/responsibleai/_managers/causal_manager.py
gregorybchris: Done, thanks!

responsibleai/responsibleai/_managers/causal_manager.py
gaugup: Ok. We should ask the maintainers of CausalAnalysis in econml if there are bounds on treatment costs. Sure, you could have any value but there should be some limits.

gaugup (approved):

PR #745, opened 2021-08-02
Handle scenario when model is passed as None

gaugup:
- Logging a warning saying explanation, error analysis and coutnerfactuals may not work as expected if no valid model is passed
- Handle model as None in each of the error analysis, coutnerfactual and explainer manager toraise excecption if user does a add() operation on either of them.
- Added relevant unit tests

### DIFF
--- a/responsibleai/responsibleai/_managers/counterfactual_manager.py
+++ b/responsibleai/responsibleai/_managers/counterfactual_manager.py
@@ -108,7 +108,7 @@ class CounterfactualManager(BaseManager):
         if self._model is None:
             raise UserConfigValidationException(
                 'Model is required for counterfactual example generation and '
-                'explanations')
+                'feature importances')
 
         if new_counterfactual_config.features_to_vary != 'all' and \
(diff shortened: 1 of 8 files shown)


xuke444 (approved):

responsibleai/responsibleai/_managers/counterfactual_manager.py
```diff
@@ -105,6 +105,11 @@ def _create_diceml_explainer(self, method, continuous_features):
         return dice_explainer
 
     def _add_counterfactual_config(self, new_counterfactual_config):
+        if self._model is None:
+            raise UserConfigValidationException(
+                'Model is required for counterfactual example generation and '
+                'explanations')
```
gregorybchris, line 111: Is this message confusing since we sometimes use "explanations" to mean interpret explanations? Was this meant to cover counterfactual importances?

gaugup: Done. Changed to "feature importances"

PR #750, opened 2021-08-03
Incorporate Uncertainty Quantification in Fairness widget v2

alexquach:
Copy of #654 but PR created within /microsoft/responsible-ai-widgets instead of a forked version


libs/mlchartlib/src/lib/components/ChartBuilder.ts
```diff
@@ -30,13 +30,21 @@ export class ChartBuilder {
       y: any;
       group: any;
       size: any;
+      xLowerBound: any;
```
xuke444, line 33: where do we use these values, can we avoid "any"?

libs/fairness/src/lib/Controls/DropdownBar.styles.ts
```diff
@@ -0,0 +1,88 @@
...
+    },
+    frame: {
+      display: "flex",
+      flex: 1,
+      flexDirection: "column"
+    },
+    infoButton: {
+      border: "1px solid",
+      borderRadius: "50%",
+      color: theme.semanticColors.bodyText,
```
riedgar-ms, line 56: From PR meeting:
Don't override `color`

libs/fairness/src/lib/Controls/DropdownBar.styles.ts
```diff
@@ -0,0 +1,88 @@
...
+      float: "left",
+      fontSize: "12px",
```
riedgar-ms, line 58: From PR meeting:
Use things like `<Text block variant="xLarge"` in the actual code

libs/fairness/src/lib/Controls/DropdownBar.styles.ts
```diff
@@ -0,0 +1,88 @@
...
+    },
+    mainRight: {
+      padding: "30px 0 0 35px",
+      width: "300px"
+    },
+    modalContentHelp: {},
+    modalContentHelpText: {
+      padding: "0px 20px",
+      textAlign: "top",
+      wordWrap: "break-word"
```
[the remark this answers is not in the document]
riedgar-ms: From PR meeting:
Try removing things like this, to see if they aren't needed.

libs/fairness/src/lib/Controls/DropdownBar.styles.ts
```diff
@@ -0,0 +1,88 @@
...
+    },
+    spinner: {
+      margin: "auto",
+      padding: "40px"
+    },
+    title: {
+      marginBottom: 12
+    },
+    toolTipWrapper: {
```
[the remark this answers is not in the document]
riedgar-ms: From PR review:
Can you clarify what you mean @xuke444 ?

libs/fairness/src/lib/Controls/ModalHelp.tsx
```diff
@@ -1,96 +1,171 @@
...
-      },
-      rootHovered: {
-        color: this.props.theme.semanticColors.bodyBackgroundHovered
-      }
-    };
+    const graphInfoButtonId = "graphInfoButtonId";
+    const errorBarInfoButtonId = "errorBarInfoButtonId";
 
-    const cancelIcon: IIconProps = { iconName: "Cancel" };
+    const graphCallout = (
```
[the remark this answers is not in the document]
riedgar-ms: From PR review:
Discussed how to split into a separate `.tsx` file

libs/fairness/src/lib/Controls/ModalHelp.tsx
```diff
@@ -1,96 +1,171 @@
...
+                  </>
+                </Text>
+              ))}
+            </Callout>
+          )}
+        </Stack>
+      </Stack>
+    );
+
+    const errorBarCallout = (
```
riedgar-ms, line 93: PR Discussion:
Probably should be in a separate `.tsx` file as well

libs/fairness/src/lib/WizardReport.tsx
```diff
@@ -342,51 +359,81 @@ export class WizardReport extends React.PureComponent<IReportProps, IState> {
...
+          overpredictions = await this.getMetric(
+            "overprediction",
+            this.props.errorPickerProps.selectedErrorKey
+          );
+          underpredictions = await this.getMetric(
+            "underprediction",
+            this.props.errorPickerProps.selectedErrorKey
+          );
+          outcomes = await this.getMetric(
+            // call will always return error bounds
```
riedgar-ms, line 412: Not sure about this - a previously saved set of metrics may not have errors included.

raiwidgets/raiwidgets/fairness_dashboard.py
```diff
@@ -39,13 +39,20 @@ def __init__(
             public_ip=None,
             port=None,
             fairness_metric_module=None,
-            fairness_metric_mapping=None):
+            fairness_metric_mapping=None,
+            custom_metric_functions=None):
```
riedgar-ms, line 43: What's this for?

raiwidgets/raiwidgets/fairness_dashboard.py
```diff
@@ -39,13 +39,20 @@ def __init__(
...
         """Initialize the fairness dashboard."""
 
         metrics_module = FairnessMetricModule(
             module_name=fairness_metric_module,
             mapping=fairness_metric_mapping)
 
+        if custom_metric_functions is not None:
+            metrics_module._metric_methods.update(custom_metric_functions)
+        # metrics_module.error_methods = [
```
riedgar-ms, line 52: Please removed commented out code like these lines

raiwidgets/raiwidgets/fairness_metric_calculation.py
```diff
@@ -13,6 +19,98 @@
...
+    adjusted_standard_deviation = np.sqrt(
+        (p * (1 - p) + z * z / (4 * n))) / np.sqrt(n)
+    lower_bound = (centre_adjusted_probability - z *
+                   adjusted_standard_deviation) / denominator
+    upper_bound = (centre_adjusted_probability + z *
+                   adjusted_standard_deviation) / denominator
+    return (round(lower_bound, digits), round(upper_bound, digits))
+
+
+# Unused
```
riedgar-ms, line 41: If this is still the case, you should remove this function.

riedgar-ms (commented):
Went over in meeting with @romanlutz  and @alexquach

apps/dashboard-e2e/src/describer/fairness/fairnessDatasets.ts
```diff
@@ -6,6 +6,7 @@ import { IFairnessMetadata, PredictionTypes } from "./IFairnessMetadata";
 const fairnessDatasets = {
   binaryClassification: {
     charts: ["Selection rate", "False positive and false negative rates"],
+    defaultErrorMetric: "Enabled",
```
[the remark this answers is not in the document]
alexquach: Yes, but the intent was to leave it extensible for more options i.e. enabled w/ multiplicity correction

libs/fairness/src/lib/Shared.styles.ts
```diff
@@ -66,6 +91,17 @@ export const SharedStyles: () => IProcessedStyleSet<ISharedStyles> = () => {
...
+    errorCalloutHeader: {
+      paddingRight: "5px"
+    },
+    errorCalloutWrapper: {
+      alignItems: "center",
+      display: "flex",
+      flexFlow: "row nowrap"
+    },
+    graphCalloutWrapper: {
+      marginLeft: "80px"
```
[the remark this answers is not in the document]
alexquach: image
To align the Callout to the left axis of the graph, anything further left looks weird

raiwidgets/raiwidgets/fairness_metric_calculation.py
```diff
@@ -13,6 +19,98 @@
     "needs to provide a MetricFrame class to calculate metrics. For an " \
     "refer to fairlearn.metrics.MetricFrame"
 
+z_score = 1.959964  # This z-score corresponds to the 95% confidence interval
+alpha = 0.95        # Conventional level of power for statistical tests
+digits_of_precision = 4
+
+
+def compute_wilson_bounds(p, n, digits=digits_of_precision, z=z_score):
```
riedgar-ms, line 27: As discussed, please put in a few tests of these new functions you're adding, even if you just have to evaluate the functions with particular values, and then have a test that verifies (using `pytest.approx`) that the value is stable.

raiwidgets/raiwidgets/fairness_metric_calculation.py
```diff
@@ -13,6 +19,98 @@
...
+
+def recall_wilson(y_true, y_pred):
+    assert len(y_true) == len(y_pred)
+    tn, fp, fn, tp = confusion_matrix(y_true, y_pred).ravel()
+    bounds = compute_wilson_bounds(
+        tp / (tp + fn), tp + fn, digits_of_precision, z_score)
+    return bounds
+
+
+def precision_wilson(y_true, y_pred):
```
riedgar-ms, line 90: For testing things like this, at a minimum please create a couple of short (say 10 element) input arrays, and have a test that the output value is unchanged. It's not a comprehensive test, but it's better than nothing.

raiwidgets/raiwidgets/fairness_metric_calculation.py
```diff
@@ -13,6 +19,98 @@
...
+    assert len(y_true) == len(y_pred)
+    tn, fp, fn, tp = confusion_matrix(y_true, y_pred).ravel()
+    bounds = compute_wilson_bounds(
+        tp / (tp + fp), tp + fp, digits_of_precision, z_score)
+    return bounds
+
+
+# def standard_normal_wrapper(y_true, y_pred, func):
```
riedgar-ms, line 98: Again, please remove commented out code.

apps/dashboard-e2e/src/describer/fairness/singleModelView/describeSingleModelViewWithError.ts
```diff
@@ -0,0 +1,96 @@
...
+        .then(() => {
+          // click on model point in model comparison view
+          props.chart = new ScatterChart("#FairnessPerformanceTradeoffChart");
+          props.chart.clickNthPoint(0);
+        });
+    });
+    it("should find single model view", () => {
+      getSpan("Assessment results").should("exist");
+
+      cy.get("#sensitiveFeatureDropdown").should(
```
romanlutz, line 29: all of these are common with the "normal" check (without "WithError" in the file name). Can't we just add
```
if (checkErrorBars) {
    <check>
}
```
and keep it in a single test?

romanlutz: same applies in the model comparison file

apps/dashboard/src/app/applications.ts
```diff
@@ -110,12 +112,18 @@ export const applications: IApplications = <const>{
...
+        data: binaryClassificationWithError
+      },
       precomputedBinary: { data: precomputedBinary },
       precomputedBinaryMissingMetrics: {
         data: precomputedBinaryMissingMetrics
       },
       precomputedBinaryTwo: { data: precomputedBinaryTwo },
+      precomputedBinaryWithError: { data: precomputedBinaryWithError },
       probability: { data: probability },
       regression: { data: regression }
```
romanlutz, line 128: are we not doing regression?

apps/dashboard/src/fairness/__mock_data__/precomputedBinaryWithError.ts
```diff
@@ -0,0 +1,162 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { IFairnessData } from "@responsible-ai/core-ui";
+
+import { precomputedBinary } from "./precomputedBinary";
+
+Object.assign(precomputedBinary, {
```
romanlutz, line 8: Am I understanding this correctly? We're assigning something new to `precomputedBinary` and then at the end of the file assign it to another variable that's exported. Why not use the exported one right away?

romanlutz: Does this change precomputedBinary?

libs/fairness/src/lib/Controls/CalloutHelpBar.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import { IErrorPickerPropsV2 } from "../FairnessWizard";
+import { SharedStyles } from "../Shared.styles";
+
+import { CalloutErrorBars } from "./CalloutErrorBars";
+import { CalloutGraph } from "./CalloutGraph";
+
+interface ICalloutHelpBarProps {
+  graphCalloutStrings: string[];
+  fairnessBounds?: Array<IBounds | undefined>;
```
romanlutz, line 17: The ? already makes it possible to be undefined. Doing it twice can lead to weird situations and should be avoided.

libs/fairness/src/lib/Controls/CalloutHelpBar.tsx
```diff
@@ -0,0 +1,67 @@
...
+        <Stack className={sharedStyles.errorBarsCalloutWrapper}>
+          <CalloutErrorBars errorBarsInfoButtonId={errorBarsInfoButtonId} />
+          <Toggle
+            className={sharedStyles.toggle}
+            id="errorMetricDropdown"
+            defaultChecked={
+              this.props.errorPickerProps.selectedErrorKey === "enabled"
+            }
+            disabled={
+              (typeof this.props.fairnessBounds === "undefined" ||
```
romanlutz, line 50: `this.props.fairnessBounds` is sufficient, because it'll be undefined and therefore false, or not undefined and therefore true.

libs/fairness/src/lib/Controls/CalloutHelpBar.tsx
```diff
@@ -0,0 +1,67 @@
...
+                _.isEmpty(this.props.fairnessBounds.filter(Boolean))) &&
```
romanlutz, line 51: What does this actually do? I've never seen `filter.(Boolean)` before.

libs/fairness/src/lib/Controls/ModelComparisonChart.tsx
```diff
@@ -162,15 +171,44 @@ export class ModelComparisonChart extends React.Component<
...
+          FairnessUpperBound: 0,
           index,
-          Performance: performance
+          Performance: performance,
+          PerformanceLowerBound: 0,
+          PerformanceUpperBound: 0
         };
       });
 
+      if (errorKey !== "disabled") {
```
romanlutz, line 188: If there's only disabled and enabled, should it be a bool instead?

libs/fairness/src/lib/Controls/ModelComparisonChart.tsx
```diff
@@ -341,15 +421,41 @@ export class ModelComparisonChart extends React.Component<
...
         });
 
       const performanceArray = (await Promise.all(performancePromises)).map(
-        (metric) => metric.global
+        (metric) => {
+          return metric.global;
+        }
+      );
+
+      const performanceBounds = (await Promise.all(performancePromises)).map(
```
romanlutz, line 435: I think I wrote this in the first iteration of the PR: why not assign the result of `await Promise.all(performancePromises)` to a variable? At first glance, this looks like you're waiting on several different things even though it's the same one.

libs/fairness/src/lib/Controls/OutcomePlot.tsx
```diff
@@ -61,10 +75,33 @@ export class OutcomePlot extends React.PureComponent<IOutcomePlotProps> {
...
+        barPlotlyProps.data[0].error_x = {
+          array: this.props.metrics.outcomes.binBounds?.map(
+            (binBound, index) => {
+              if (this.props.metrics.outcomes?.bins[index] !== undefined) {
+                return binBound.upper - this.props.metrics.outcomes.bins[index]; // convert from bounds to relative error
+              }
+              return 0;
+            }
+          ) || [0],
+          arrayminus: this.props.metrics.outcomes.binBounds?.map(
```
romanlutz, line 88: The naming here doesn't make it obvious to me what this is. We have array above and arrayminus here. What do these mean? Can you add a comment above to explain? Thanks!

libs/fairness/src/lib/Controls/OutcomePlot.tsx
```diff
@@ -128,22 +165,77 @@ export class OutcomePlot extends React.PureComponent<IOutcomePlotProps> {
...
         localization.Fairness.Report.regressionOutcomesHowToRead
       ];
     }
 
+    barPlotlyProps.data[0].customdata = [];
+    const digitsOfPrecision = 1;
+
+    for (let i = 0; i < this.props.dashboardContext.groupNames.length; i++) {
+      const tempY = barPlotlyProps.data[0].y
+        ? barPlotlyProps.data[0].y[i]
```
romanlutz, line 178: This could just be `barPlotlyProps.data[0].y?.[i] to do this in 1 line.

libs/fairness/src/lib/Controls/OutcomePlot.tsx
```diff
@@ -128,22 +165,77 @@ export class OutcomePlot extends React.PureComponent<IOutcomePlotProps> {
...
+        : undefined;
+      const tempX = barPlotlyProps.data[0].x
+        ? barPlotlyProps.data[0].x[i]
+        : undefined;
+      const x =
+        tempX !== undefined && typeof tempX == "number" ? tempX : undefined;
```
romanlutz, line 184: What's going on here? It should not be a number? Can you add a comment to explain the reasoning?

libs/fairness/src/lib/Controls/OutcomePlot.tsx
```diff
@@ -128,22 +165,77 @@ export class OutcomePlot extends React.PureComponent<IOutcomePlotProps> {
...
+      const y =
+        tempY !== undefined && typeof tempY == "string"
+          ? tempY.trim()
+          : undefined;
+      const lowerErrorX =
+        barPlotlyProps.data[0]?.error_x?.type === "data" &&
+        barPlotlyProps.data[0].error_x.arrayminus
+          ? barPlotlyProps.data[0].error_x.arrayminus[i]
```
romanlutz, line 192: please simplify as
```
barPlotlyProps.data[0].error_x.arrayminus?.[i]
```

romanlutz (commented):
I'm halfway through and will continue in an hour.

libs/fairness/src/lib/Controls/PerformancePlot.tsx
```diff
@@ -216,25 +304,98 @@ export class PerformancePlot extends React.PureComponent<IPerformancePlotProps>
...
+          const tempData = barPlotlyProps.data
+            ? barPlotlyProps.data[j]
+            : undefined;
+          const tempY = tempData && tempData.y ? tempData.y[i] : undefined;
+          const tempX = tempData && tempData.x ? tempData.x[i] : undefined;
+
+          const x =
+            tempX !== undefined && typeof tempX == "number" ? tempX : undefined;
+          const y =
+            tempY !== undefined && typeof tempY == "string"
```
romanlutz, line 335: This is kind of unintuitive. What types could it have other than string? Perhaps we want to rule out such cases ahead of time.

libs/fairness/src/lib/Controls/PerformancePlot.tsx
```diff
@@ -216,25 +304,98 @@ export class PerformancePlot extends React.PureComponent<IPerformancePlotProps>
...
+            typeof upperErrorX == "number" &&
+            x !== undefined
+              ? `[${(100 * (x - lowerErrorX)).toFixed(digitsOfPrecision)}%, ${(
+                  100 *
+                  (x + upperErrorX)
+                ).toFixed(digitsOfPrecision)}%]`
+              : "";
+
+          if (
+            barPlotlyProps !== undefined &&
```
romanlutz, line 360: `barPlotlyProps.data?.[j]?.customdata && _isArray(...)`

libs/fairness/src/lib/FairnessWizard.tsx
```diff
@@ -54,6 +55,12 @@ export interface IFairnessPickerPropsV2 {
   onFairnessChange: (newKey: string) => void;
 }
 
+export interface IErrorPickerPropsV2 {
```
romanlutz, line 58: Please don't use V2 since this is not relevant for the older V1 that's already been deleted.

libs/fairness/src/lib/FairnessWizard.tsx
```diff
@@ -175,6 +191,12 @@ export class FairnessWizardV2 extends React.PureComponent<
       defaultFairnessMetricPrioritization
     );
 
+    if (this.props.errorKey === "disabled") {
+      selectedErrorKey = "disabled";
+    } else {
+      selectedErrorKey = "enabled";
```
romanlutz, line 197: So if we run this with precomputed data and don't explicitly mark "disabled", will it assume that it's enabled?

libs/fairness/src/lib/Shared.styles.ts
```diff
@@ -25,11 +27,34 @@ export interface ISharedStyles {
...
+  title: IStyle;
+  toggle: IStyle;
+  errorCalloutHeader: IStyle;
 }
 
 export const SharedStyles: () => IProcessedStyleSet<ISharedStyles> = () => {
   const theme = getTheme();
   return mergeStyleSets<ISharedStyles>({
+    actionButton: {
+      height: "auto"
```
romanlutz, line 43: Is this needed?

libs/fairness/src/lib/Shared.styles.ts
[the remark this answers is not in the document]
romanlutz: I guess we should have put everything in a container that has default padding of 80px. Not your concern in this PR, though.

libs/fairness/src/lib/util/ErrorMetrics.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const errorOptions: { [key: string]: IErrorOption } = {
+  disabled: {
+    description: localization.Fairness.Metrics.errorMetricDisabledDescription,
+    isMinimization: false,
+    key: "disabled",
+    title: localization.Fairness.Metrics.errorMetricDisabled
+  },
+  enabled: {
+    description: localization.Fairness.Metrics.errorMetricEnabledDescription,
```
romanlutz, line 21: Are all these properties getting used somewhere? I can't seem to find them in the code. If not, we should probably not have this and just use boolean values for enabled/disabled.

libs/fairness/src/lib/util/MetricsCache.ts
```diff
@@ -48,105 +57,199 @@ export class MetricsCache {
...
       truePositiveRateMetric === Number.NaN
     ) {
-      return Number.NaN;
+      return { overall: Number.NaN };
     }
 
     if (fairnessMethod === FairnessModes.Difference) {
       const maxMetric = max([falsePositiveRateMetric, truePositiveRateMetric]);
-      return maxMetric === undefined ? Number.NaN : maxMetric;
+      return maxMetric === undefined
```
romanlutz, line 97: nit: 
```
return { overall: maxMetric ?? Number.NaN }
```

notebooks/fairness-dashboard-loan-allocation.ipynb
```diff
@@ -22,12 +22,94 @@
   },
   {
    "cell_type": "code",
```
romanlutz, line 24: please clear the output from the notebook using "Restart and clear output"

raiwidgets/raiwidgets/fairness_metric_calculation.py
```diff
@@ -52,71 +150,108 @@ def __init__(self, module_name=None, mapping=None):
...
-                    "function": selection_rate
+                    "function": selection_rate,
+                    "error_function": lambda y_true, y_pred:\
+                        wilson_wrapper(y_true, y_pred, selection_rate)
                 },
                 "auc": {
                     "model_type": ["probability"],
-                    "function": skm.roc_auc_score
+                    "function": skm.roc_auc_score,
+                    # "error_function": lambda y_true, y_pred:
```
romanlutz, line 208: What happens if you use a metric whose `error_function` is now commented out?

apps/dashboard-e2e/src/describer/fairness/IFairnessMetadata.ts
```diff
@@ -9,17 +9,20 @@ export interface IFairnessMetadata {
...
   fairnessMetrics: string[];
   defaultFairnessMetric: string;
+  defaultErrorMetric: string;
   numberOfModels: number;
   charts: string[];
 }
 
 export enum PredictionTypes {
   BinaryClassification = "binaryClassification",
+  BinaryClassificationWithError = "binaryClassificationWithError",
```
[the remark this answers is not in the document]
alexquach: It's necessary for the code later on to retrieve the right URL that has the error bars default on. While `checkErrorBars` is for the testing platform to know to test for the error bars

apps/dashboard/src/fairness/__mock_data__/precomputedBinaryWithError.ts
alexquach: I was trying to extend precomputedBinary, since what I'm exporting here is precomputedBinary with additional key-values

libs/fairness/src/lib/Controls/OutcomePlot.tsx
alexquach: array and arrayminus are from the Plotly API

libs/fairness/src/lib/Controls/PerformancePlot.tsx
alexquach: Datum allows for "string" | "date" | "number" | undefined

raiwidgets/raiwidgets/fairness_metric_calculation.py
romanlutz: OK. Why are we keeping the comment?

libs/fairness/src/lib/Controls/OutcomePlot.tsx
romanlutz: Sure, but it's still not obvious. 🙂 Believe me, anyone reading this in 3 months will be happy for any hint.

apps/dashboard/src/fairness/__mock_data__/precomputedBinaryWithError.ts
romanlutz: Right, but does this change precomputedBinary which is expected not to have error bars (?). Just directly assign it to the variable at the end that you're exporting (?)
Or deep copy precomputedBinary and then assign to the copy, but I wouldn't change precomputedBinary.

apps/dashboard-e2e/src/describer/fairness/IFairnessMetadata.ts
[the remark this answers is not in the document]
romanlutz: What do you mean by 
> retrieve the right URL
?
Where is that happening? It really seems like this should just be a separate bool rather than making this enum more complicated.

libs/fairness/src/lib/Controls/OutcomePlot.tsx
alexquach: It should be a number. Ke recommended to do type checking instead of type casting

libs/fairness/src/lib/FairnessWizard.tsx
alexquach: Yes I think the only case you wouldn't want them to be enabled would be if it's legacy data (in which you would explicitly disabled them)

libs/fairness/src/lib/Controls/CalloutHelpBar.tsx
alexquach: `Array<IBounds | undefined> | undefined` is not redundant because it allows for `IBounds[] | undefined[] | undefined`, where the key is the array of undefined items

libs/fairness/src/lib/Controls/CalloutHelpBar.tsx
alexquach: image
I thought so too, but the compiler still gives me an error

libs/fairness/src/lib/Controls/CalloutHelpBar.tsx
alexquach: filters for truthy values

libs/fairness/src/lib/Shared.styles.ts
alexquach: yes

apps/dashboard/src/fairness/__mock_data__/precomputedBinaryWithError.ts
```diff
@@ -0,0 +1,165 @@
...
+const precomputedBinaryCopy = JSON.parse(JSON.stringify(precomputedBinary));
+
+Object.assign(precomputedBinaryCopy, {
+  ...precomputedBinary,
+  precomputedMetrics: [
+    // gender
+    [
+      // Model 1 metrics
+      {
+        accuracy_score: {
```
riedgar-ms, line 18: I'm not seeing one of these with both `bounds` and `binBounds`? Or am I being blind?

apps/dashboard/src/fairness/utils.ts
```diff
@@ -64,13 +68,26 @@ export function generateRandomMetrics(
   const binSize = Math.max(...request.binVector);
   const bins: number[] = new Array(binSize + 1)
     .fill(0)
-    .map(() => Math.random());
-  // bins[2] = undefined;
+    .map(() => Math.random() / 3 + 0.33);
+  const binBounds: IBounds[] = bins.map((bin) => {
+    return {
+      lower: bin - Math.pow(Math.random() / 3, 2),
```
riedgar-ms, line 74: This looks like it's in product code? Or am I just not familiar with how TS projects are laid out?

libs/fairness/src/lib/Controls/ModelComparisonChart.tsx
```diff
@@ -189,30 +231,94 @@ export class ModelComparisonChart extends React.Component<
...
+              lowerErrorX !== 0 &&
+              upperErrorX !== 0 &&
+              typeof lowerErrorX == "number" &&
+              typeof upperErrorX == "number" &&
+              x !== undefined
+                ? `[${(x - lowerErrorX).toFixed(digitsOfPrecision)}, ${(
+                    x + upperErrorX
+                  ).toFixed(digitsOfPrecision)}]`
+                : "";
+            const yBounds =
```
riedgar-ms, line 287: At first glance, this appears to be the same as the `xBounds` calculation, swapping out `x` for `y`? If so, can it be refactored into a function?

libs/fairness/src/lib/util/calculateFairnessMetric.test.ts
```diff
@@ -0,0 +1,197 @@
...
+      bins: [0.5, 0.2],
+      global: 0.5
+    };
+    const result = calculateFairnessMetric(mockValue, FairnessModes.Difference);
+    const expectedResult = {
+      overall: 0.3
+    };
+    expect(result).toEqual(expectedResult);
+  });
+  it("should calculate difference (with overlap) correctly", () => {
```
riedgar-ms, line 34: Great to see these tests. Since Miro's proposed formulae make assumptions that `x1 < y1`, can you repeat the tests with the order of bins reversed, making sure that you get the same overall answers?

libs/fairness/src/lib/util/calculateFairnessMetric.test.ts
```diff
@@ -0,0 +1,197 @@
...
+      bins: [0.5, 0.2],
+      global: 0.5
+    };
+    const result = calculateFairnessMetric(mockValue, FairnessModes.Ratio);
+    const expectedResult = {
+      overall: 0.4
+    };
+    expect(result).toEqual(expectedResult);
+  });
+  it("should calculate ratio (with overlap) correctly", () => {
```
riedgar-ms, line 94: Again, please make sure that these work with the bin order reversed as well.

riedgar-ms: Also, when it comes to overlap, there are two possible - full overlap (i.e. x1 < y1 and x2>y2) and partial (where the order of points is x1, y1, x2, y2). Those both need to be checked as well.

notebooks/fairness-dashboard-loan-allocation.ipynb
```diff
@@ -11,39 +10,40 @@
     "For the purposes of this notebook, we shall treat this as a loan decision problem. We will pretend that the label indicates whether or not each individual repaid a loan in the past. We will use the data to train a predictor to predict whether previously unseen individuals will repay a loan or not. The assumption is that the model predictions are used to decide whether an individual should be offered a loan.\n",
     "\n",
     "We will first train a fairness-unaware predictor and use the fairness dashboard to demonstrate how this model leads to unfair decisions (under a specific notion of fairness called *demographic parity*). We then mitigate unfairness by applying the `GridSearch` algorithm from `Fairlearn` package."
-   ]
+   ],
+   "metadata": {}
```
riedgar-ms, line 14: It doesn't look like there are significant changes to this notebook? If so, can it just be removed from the PR?

raiwidgets/raiwidgets/fairness_metric_calculation.py
```diff
@@ -13,6 +15,108 @@
...
+def mae_standard_normal(y_true, y_pred):
+    assert len(y_true) == len(y_pred)
+    absolute_error = np.abs(y_true - y_pred)
+    sample_std = np.std(absolute_error)
+    mae = np.mean(absolute_error)
+    standard_error = z_score * sample_std / np.sqrt(len(y_true))
+    return mae - standard_error, mae + standard_error
+
+
+# TODO: Logarithmic loss implemented, but should be logistic loss
```
riedgar-ms, line 111: Is this still TODO?

riedgar-ms (commented):
Great to see progress!

apps/dashboard/src/fairness/__mock_data__/precomputedBinaryWithError.ts
alexquach: Performance metrics use bounds directly for the multi-model view, while Disparity Metrics use the binBounds to calculate the bounds for the multi-model view. binBounds are not necessary for performance metrics since they're not being displayed in the single model-view in this PR

apps/dashboard/src/fairness/utils.ts
alexquach: This function was already generating random points for local testing and testing on Azure Blob. I just expanded it to also generate bounds

apps/dashboard/src/fairness/__mock_data__/precomputedBinaryWithError.ts
riedgar-ms: But shouldn't having both `bounds` and `binBounds` be valid in general?

alexquach: Yes it's valid, but not necessary

libs/localization/src/lib/en.json
```diff
@@ -262,9 +262,14 @@
       "save": "Save"
     },
     "DropdownHeaders": {
-      "fairnessMetric": "Fairness metric",
+      "sensitiveFeature": "Sensitive feature",
```
xuke444, line 265: somehow prettier automatically sort whole file
looks like your version is kind of out-dated, do you mind merge with main branch?

raiwidgets/raiwidgets/fairness_metric_calculation.py
alexquach: removing it from this PR

libs/fairness/src/lib/util/calculateFairnessMetric.test.ts
```diff
@@ -0,0 +1,293 @@
...
+    const expectedResult = {
+      bounds: { lower: 0, upper: closeTo(0.5) },
+      overall: closeTo(0.4)
+    };
+    expect(result).toMatchObject(expectedResult);
+  });
+  it("should calculate difference (with complete overlap) correctly", () => {
+    const mockValue = {
+      binBounds: [
+        { lower: 0.2, upper: 0.8 },
```
riedgar-ms, line 69: Am I missing a variant of this with the reverse order as well?

libs/fairness/src/lib/util/calculateFairnessMetric.test.ts
```diff
@@ -0,0 +1,293 @@
...
+      global: 0.5
+    };
+    const result = calculateFairnessMetric(mockValue, FairnessModes.Ratio);
+    const expectedResult = {
+      bounds: { lower: closeTo(0.33), upper: 1 },
+      overall: closeTo(0.5)
+    };
+    expect(result).toMatchObject(expectedResult);
+  });
+  it("should calculate ratio (with complete overlap) correctly", () => {
```
riedgar-ms, line 174: Again, please can you add a 'complete overlap and reversed order' check

notebooks/fairness-dashboard-loan-allocation.ipynb
```diff
@@ -1,332 +1,1366 @@
-{
```
riedgar-ms, line 1: Is this supposed to have changed?

alexquach: Yes, reverting to no net changes for the entirety of the PR

libs/fairness/src/lib/util/calculateFairnessMetric.ts
```diff
@@ -0,0 +1,179 @@
...
+              binBounds[j].upper - binBounds[i].lower
+            );
+            const maxCandidate2 = Math.abs(
+              binBounds[i].upper - binBounds[j].lower
+            );
+            maxCandidate =
+              maxCandidate1 > maxCandidate2 ? maxCandidate1 : maxCandidate2;
+
+            minDiff = 0;
+            maxDiff = maxCandidate > maxDiff ? maxCandidate : maxDiff;
```
riedgar-ms, line 150: A comment.... doesn't doing things this way mean that the bound which contribute to the bound on the difference might not be the point which contributes to the maximum difference itself?

I'm thinking of a case where you have:
- 0.2 +/- 0.1
- 0.8 +/- 0.1
- 0.5 +/ 0.45

The difference will be 0.8-0.2=0.6, but the difference bound will be dominated by the bound on 0.5 (if I'm thinking things through correctly).

I think that your behaviour is reasonable, but it should be documented - both with a comment here and a test case or two below (with their own comments, highlighting what's going on).

libs/localization/src/lib/en.json
```diff
@@ -262,10 +262,15 @@
...
     },
     "DropdownHeaders": {
+      "errorMetric": "Enable error bars",
       "fairnessMetric": "Fairness metric",
       "performanceMetric": "Performance metric",
       "sensitiveFeature": "Sensitive feature"
     },
+    "ErrorBounds": {
+      "howToRead": "How to read these error bars",
+      "introModalText": "Error bars show the uncertainty in the metric values due to the size of the evaluation data set. We show 95% confidence intervals calculated using Wilson scores (for binary classification) or normal approximation (for other tasks)."
```
riedgar-ms, line 272: Is this the final version agreed with the PMs? Weren't we going to put in some caveat about 'These are only approximate, and you data may violate the assumptions made' ?

libs/fairness/src/lib/util/calculateFairnessMetric.ts
alexquach: I think this behavior naturally follows the technique of coming up with the most conservative bounds. What do you think should be specifically commented on? I don't think there is anything special from the fact that the bounds on 0.5 contributed to the upper and lower bounds of the difference, but 0.5 did not contribute to the single point/aggregate difference

riedgar-ms: Just a note that when computing the bounds on the difference, all bounds are considered, not just those at the nominal extremes. I agree that this is an entirely reasonable approach.

libs/localization/src/lib/en.json
riedgar-ms: And I've just seen the email showing the current text came from Miro. I'll resolve this.

libs/fairness/src/lib/util/calculateFairnessMetric.test.ts
```diff
@@ -0,0 +1,459 @@
...
+      overall: closeTo(0.4)
+    };
+    expect(result).toMatchObject(expectedResult);
+  });
+  it("should handle difference case with large binBound that provides the largest and smallest bounds", () => {
+    const mockValue = {
+      binBounds: [
+        { lower: 0.1, upper: 0.3 },
+        { lower: 0.7, upper: 0.9 },
+        { lower: 0.08, upper: 0.95 }
```
riedgar-ms, line 212: This is subtler.... and I haven't quite figured out the answer yet.

However, I would argue that you can only get to a minimum difference of zero for cases like:
```json
        { lower: 0.1, upper: 0.3 },
        { lower: 0.7, upper: 0.9 },
        { lower: 0.08, upper: 0.95 },
        { lower: 0.08, upper: 0.95 }
```
where the repetition allows the max and mins to fully overlap

riedgar-ms (commented):
I'm taking a look at the detailed bounds calculations, and I'm not sure about the cases where n>2

libs/fairness/src/lib/util/calculateFairnessMetric.test.ts
```diff
@@ -0,0 +1,459 @@
...
+        { lower: 0.25, upper: 0.35 },
+        { lower: 0.65, upper: 0.75 },
+        { lower: 0.45, upper: 0.55 }
+      ],
+      bins: [0.3, 0.7, 0.5],
+      global: 0.5
+    };
+    const result = calculateFairnessMetric(mockValue, FairnessModes.Difference);
+    const expectedResult = {
+      bounds: { lower: closeTo(0.1), upper: closeTo(0.5) },
```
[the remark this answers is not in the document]
riedgar-ms: Based on conversation with @MiroDudik , the bound here should indeed be (0.3, 0.5)

libs/fairness/src/lib/util/calculateFairnessMetric.test.ts
```diff
@@ -0,0 +1,459 @@
...
+        { lower: 0.45, upper: 0.7 },
+        { lower: 0.25, upper: 0.35 },
+        { lower: 0.65, upper: 0.75 }
+      ],
+      bins: [0.5, 0.3, 0.7],
+      global: 0.5
+    };
+    const result = calculateFairnessMetric(mockValue, FairnessModes.Difference);
+    const expectedResult = {
+      bounds: { lower: closeTo(0), upper: closeTo(0.5) },
```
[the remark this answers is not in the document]
riedgar-ms: Have been discussing with @MiroDudik . Please use what we settled on there

libs/fairness/src/lib/util/calculateFairnessMetric.test.ts
alexquach: I believe the case should be when all bounds overlap

libs/fairness/src/lib/util/calculateFairnessMetric.test.ts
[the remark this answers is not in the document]
alexquach: agree this should be 0.3 with the new way to calculate

libs/fairness/src/lib/util/calculateFairnessMetric.test.ts
```diff
@@ -0,0 +1,493 @@
...
+      global: 0.5
+    };
+    const result = calculateFairnessMetric(mockValue, FairnessModes.Ratio);
+    const expectedResult = {
+      bounds: { lower: closeTo(0.33), upper: closeTo(0.67) },
+      overall: closeTo(0.5)
+    };
+    expect(result).toMatchObject(expectedResult);
+  });
+  it("should calculate ratio with 3+ groups correctly", () => {
```
riedgar-ms, line 368: Again, maybe mention that you're permuting the order

libs/fairness/src/lib/util/calculateFairnessMetric.test.ts
```diff
@@ -0,0 +1,493 @@
...
+      ],
+      bins: [0.2, 0.8, 0.5],
+      global: 0.5
+    };
+    const result = calculateFairnessMetric(mockValue, FairnessModes.Ratio);
+    const expectedResult = {
+      // from the 1st and 3rd point: the lower bound = 0.1 / 0.95
+      // from the 1st and 2nd point: the upper bound = 0.3 / 0.7
+      // the smallest ratio between nominal bins: overall = 0.2 / 0.8
+      bounds: { lower: closeTo(0.09), upper: closeTo(0.43) },
```
riedgar-ms, line 434: This comment is confusing me - isn't the critical one 0.08/0.9 = 0.09?

riedgar-ms (approved):
Some minor comments, but this looks good

libs/fairness/src/lib/util/calculateFairnessMetric.test.ts
alexquach: good catch

libs/fairness/src/lib/util/calculateFairnessMetric.ts
```diff
@@ -0,0 +1,198 @@
...
+        for (let j = i + 1; j < binBounds.length; j++) {
+          let lowerCandidate;
+          let upperCandidate;
+
+          // if there is overlap
+          if (
+            binBounds[i].upper > binBounds[j].lower &&
+            binBounds[j].upper > binBounds[i].lower
+          ) {
+            const lowerCandidate1 = binBounds[i].lower / binBounds[j].upper;
```
riedgar-ms, line 96: Might want an extra comment or two to outline what's going on through these sections (and through the difference calculations as well). Not too much, just something to make clear what's happening at each stage (e.g. go through all pairs, calculating the max and min ratios, noting the special case for overlap).

It's obvious now, but come back in three months, and a little guidance as to what's going on will be much appreciated.

xuke444 (approved):

PR #769, opened 2021-08-10
Bug Fix

zhb000:
Fix scroll bar should be inside details list in CF flyout
Fix UI cards are too zoomed in


xuke444 (approved):

libs/counterfactuals/src/lib/CounterfactualPanel.tsx
```diff
@@ -77,7 +77,7 @@ export class CounterfactualPanel extends React.Component<
               onChange={this.setFilterText.bind(this)}
             />
           </Stack.Item>
-          <Stack.Item>
+          <Stack.Item className={classes.detailsList}>
```
romanlutz, line 80: Perhaps naming this something other than details list would be preferable since it's actually used on a Stack.

PR #777, opened 2021-08-11
Fix No-op error analysis coverage bubble hover

zhb000:
Fix No-op error analysis coverage bubble hover


libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/TreeLegend/TreeLegend.tsx
```diff
@@ -60,35 +59,6 @@ export class TreeLegend extends React.Component<ITreeLegendProps> {
                 </div>
               </Stack>
             </Stack>
-            <svg
```
xuke444, line 63: There is a mis communication we still need that big grey dot.
but just change the mouse cursor style, from that hand pointer to regular mouse arrow since it is not clickable.

PR #784, opened 2021-08-12
fix the x and y axes on the heatmap to be consistent with generated matrix

imatiach-msft:
Change the axes to be consistent in the error analysis heatmap: the y-axis dropdown only leads to a 1d vertical heatmap, and x-axis dropdown only leads to a 1d horizontal heatmap.  When both are selected, a 2d heatmap is creates with the y-axis feature on the vertical side and the x-axis feature on the horizontal side.  Note x and y axis order in dropdowns were flipped to be more consistent with matrix notation (rows, columns, where y-axis corresponds to rows and x-axis to columns)


xuke444:
when I select feature 2 only it still renders a matrix instead of a single row
is this caused by test data or code issue?
image

imatiach-msft:
@xuke444 this is just bad test data, it doesn't happen in the actual widget

xuke444 (approved):

gregorybchris (approved):

PR #785, opened 2021-08-12
Separate previously concatenated train/test in causal analysis

gregorybchris:
Causal analysis does not allow categories of categorical variables to be missing from the training dataset while existing in the test dataset. This was originally mitigated by merging the train and test data before fitting the causal models.

This has become a problem because it is unintuitive what train and test mean when setting up model analysis. This PR removes the merge of train and test and also adds new validation to fail fast under scenarios with missing categories.

Original merge/concat was done in this PR: https://github.com/microsoft/responsible-ai-widgets/pull/566

### DIFF
--- a/responsibleai/responsibleai/_managers/causal_manager.py
+++ b/responsibleai/responsibleai/_managers/causal_manager.py
@@ -143,17 +143,16 @@ class CausalManager(BaseManager):
                        f"got {nuisance_model}")
             raise UserConfigValidationException(message)
 
-        # Update X and y to contain both train and test
-        # This solves issues with categorical features missing some
-        # categories in the test set and causing transformers to fail
-        X = pd.concat([self._train, self._test], ignore_index=True)\
-            .drop([self._target_column], axis=1)
-        y = pd.concat([self._train, self._test], ignore_index=True)[
-            self._target_column].values.ravel()
-
         categoricals = self._categorical_features
         if categoricals is None:
             categoricals = []
+
+        X_train = self._train.drop([self._target_column], axis=1)
+        X_test = self._test.drop([self._target_column], axis=1)
+        y_train = self._train[self._target_column].values.ravel()
+
+        self._validate_train_test_categories(
+            X_train, X_test, categoricals)
 
         is_classification = self._task_type == ModelTask.CLASSIFICATION
         analysis = CausalAnalysis(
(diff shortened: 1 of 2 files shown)


responsibleai/responsibleai/_managers/causal_manager.py
```diff
@@ -234,6 +231,27 @@ def add(
...
+                test_unique = np.unique(X_test[column])
+                difference = np.setdiff1d(test_unique, train_unique)
+                if difference.shape[0] != 0:
+                    discovered[column] = difference.tolist()
+        if len(discovered) > 0:
+            message = ("Causal analysis requires that every category of "
+                       "categorical features present in the test data "
+                       "be also present in the train data. "
+                       "Categories missing from train data: {}")
+            raise UserConfigValidationException(message.format(discovered))
```
gaugup, line 253: Should this function be called in the init time of model analysis class. Looks like a generic validation that canbenefit all components.

gregorybchris: If my understanding is correct, explanations handle this case by dropping categories missing in the training data, so by moving this to the top level we would fail explanations with a user config error unnecessarily.

Whether counterfactuals and error analysis handle this or not is a valid question that I think @xuke444 already validated and found that they both can handle this case.

gregorybchris: So I think we want to keep this validation causal-specific, but I'm open to moving it if any of my assumptions above are wrong!

gaugup: I don't think this is specific to causal per say. Sure CFE and EA don't result in errors but that doesn't indicate whether these components handle this situation correctly.  To me this feels like generic enough and a good case for users to know that something is amiss in their validation data.

gregorybchris: Having categories (or classes) in the validation/test data unseen in the training data is a very common scenario that is often handled by simply dropping unknown categories. sklearn has this built in: https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OneHotEncoder.html

I'm not sure I understand the concern regarding "whether these components handle this situation correctly"

gregorybchris: Tagging @imatiach-msft since I know he is very familiar with ignoring unknown categoricals from work on interpret transformations

imatiach-msft (approved):

responsibleai/responsibleai/_managers/causal_manager.py
imatiach-msft: discussed on teams, seems fine to me

gregorybchris: I will check this PR in to unblock causal analysis scenarios where train and test data need to be disjoint sets, however we can add this same validation for other components in a future PR if we can agree that it is required.

PR #786, opened 2021-08-12
Add CODEOWNERS for pull request reviews

gregorybchris:


gaugup (approved):

CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+/responsibleai @gaugup @gregorybchris @imatiach-msft @riedgar-ms @romanlutz
```
xuke444, line 1: can you add me here as well?

CODEOWNERS
```diff
@@ -0,0 +1,2 @@
...
+/raiwidgets @zhb000 @xuke444
```
xuke444, line 2: probably add roman to here as well.

CODEOWNERS
imatiach-msft: can you add /erroranalysis too?

imatiach-msft (approved):

CODEOWNERS
imatiach-msft: I think we can make the format a bit better.  Check out the code owners page in mmlspark:

https://github.com/Azure/mmlspark/blob/master/CODEOWNERS

It has really nice formatting with some comments

CODEOWNERS
```diff
@@ -0,0 +1,8 @@
+# Error analysis package
+/erroranalysis @imatiach-msft
```
xuke444, line 2: looks like we need a backup here, otherwise Ilya may not be able to check in. LOL

PR #789, opened 2021-08-13
Fix CF UI: "See prediction delta" should be on hover and not be underlined

zhb000:
Fix CF UI: "See prediction delta" should be on hover and not be underlined
Causal Recommendation Policy Table: Scroll bar appears


libs/counterfactuals/src/lib/CounterfactualPanelStyles.ts
```diff
@@ -77,6 +78,11 @@ export const counterfactualPanelStyles: () => IProcessedStyleSet<ICounterfactual
         flexDirection: "row",
         paddingTop: "5px"
       },
+      predictedLink: {
+        color: "#0078D4",
```
xuke444, line 82: use theme sematic colors

libs/counterfactuals/src/lib/CounterfactualPanelStyles.ts
```diff
@@ -77,6 +78,11 @@ export const counterfactualPanelStyles: () => IProcessedStyleSet<ICounterfactual
...
+        fontSize: "14px",
```
xuke444, line 83: do we have to override fontSize/line height?

PR #793, opened 2021-08-13
add min child samples parameter to error analysis

imatiach-msft:
add min child samples parameter to error analysis to control the minimal number of data points required to create one leaf in the tree view

### DIFF
--- a/erroranalysis/tests/test_surrogate_error_tree.py
+++ b/erroranalysis/tests/test_surrogate_error_tree.py
@@ -57,11 +57,13 @@ class TestSurrogateErrorTree(object):
         diff = pred_y != y_test
         max_depth = 3
         num_leaves = 31
+        min_child_samples = 20
         surrogate = create_surrogate_model(error_analyzer,
                                            X_test,
                                            diff,
                                            max_depth,
                                            num_leaves,
+                                           min_child_samples,
                                            cat_ind_reindexed)
         model_json = surrogate._Booster.dump_model()
         tree_structure = model_json["tree_info"][0]['tree_structure']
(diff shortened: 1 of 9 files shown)


erroranalysis/tests/test_surrogate_error_tree.py
```diff
@@ -57,11 +57,13 @@ def test_traverse_tree(self):
         diff = pred_y != y_test
         max_depth = 3
         num_leaves = 31
+        min_child_samples = 20
```
gregorybchris, line 60: In EconML this param is called `min_tree_leaf_samples` ([code](https://github.com/microsoft/responsible-ai-widgets/blob/main/responsibleai/responsibleai/_managers/causal_manager.py#L61)) and in LightGBM it's called `min_data_in_leaf` ([docs](https://lightgbm.readthedocs.io/en/latest/Parameters-Tuning.html)).
Not sure if we want to standardize with some convention?

gregorybchris (approved):

responsibleai/responsibleai/_managers/error_analysis_manager.py
```diff
@@ -179,13 +196,17 @@ def __init__(self, model, dataset, target_column,
                                        self._feature_names,
                                        self._categorical_features)
 
-    def add(self, max_depth=3, num_leaves=31, filter_features=None):
+    def add(self, max_depth=3, num_leaves=31,
+            min_child_samples=20, filter_features=None):
```
gaugup, line 200: Can you add a unit test in model analysis to test for min_child_samples?

gaugup (changes requested):

responsibleai/responsibleai/_managers/error_analysis_manager.py
imatiach-msft: done, added some basic validation logic for any trees constructed in all model analysis tests

gaugup (approved):

PR #799, opened 2021-08-17
add quantile binning to heatmap (erroranalysis python backend only)

imatiach-msft:
- add quantile binning option to heatmap in error analysis python backend
- add num bins parameter to allow the user to control how many bins should be used for continuous real-valued features
- add tests to validate new functionality

Note this is for python backend only.  UI frontend will be updated in future PR.

### DIFF
--- a/erroranalysis/erroranalysis/_internal/error_analyzer.py
+++ b/erroranalysis/erroranalysis/_internal/error_analyzer.py
@@ -13,8 +13,10 @@ from erroranalysis._internal.matrix_filter import (
 from erroranalysis._internal.surrogate_error_tree import (
     compute_error_tree as _compute_error_tree)
 from erroranalysis._internal.error_report import ErrorReport
-from erroranalysis._internal.constants import ModelTask, Metrics
+from erroranalysis._internal.constants import ModelTask, Metrics, MatrixParams
 from erroranalysis._internal.version_checker import check_pandas_version
+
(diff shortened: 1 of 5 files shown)


erroranalysis/erroranalysis/_internal/error_analyzer.py
```diff
@@ -13,9 +13,11 @@
 from erroranalysis._internal.surrogate_error_tree import (
     compute_error_tree as _compute_error_tree)
 from erroranalysis._internal.error_report import ErrorReport
-from erroranalysis._internal.constants import ModelTask, Metrics
+from erroranalysis._internal.constants import ModelTask, Metrics, MatrixParams
 from erroranalysis._internal.version_checker import check_pandas_version
 
+BIN_THRESHOLD = MatrixParams.BIN_THRESHOLD
```
gregorybchris, line 19: Can `compute_matrix` just use `MatrixParams.BIN_THRESHOLD` directly? Looks like this is a common pattern in this PR. Is this just done to make the code look cleaner when fitting in 80 columns?

gregorybchris (approved):

erroranalysis/erroranalysis/_internal/error_analyzer.py
imatiach-msft: "Is this just done to make the code look cleaner when fitting in 80 columns?"
yes exactly, it's to reduce line counts since I keep hitting the 79 chars limit all the time

PR #804, opened 2021-08-18
fix a message in release pipeline

xuke444:


TitanArmy:
Adding continuous delivery to the YAML definition

PR #805, opened 2021-08-18
Upgrade dice-ml to 0.7

gaugup:


xuke444 (approved):

responsibleai/requirements.txt
```diff
@@ -1,4 +1,4 @@
-dice-ml>=0.6.1,<0.7
+dice-ml>=0.7.1
```
xuke444, line 1: should we also assign an upper bound?
major version upgrade may end up with breaking changes.

gaugup: Good point, Let me upper bound it at 0.8.

PR #809, opened 2021-08-19
Format policy tree categorical features

gregorybchris:
The causal policy tree works for both continuous and categorical features. However categorical features are always represented by one-hot features. For example, `Fruit_apple < 0.5` would be equivalent to `Fruit != apple`.

This PR formats the categorical tree splits so that they can be displayed by the UI with `==` and `!=` rather than `<` and `>`

Main algorithm for formatting is in `responsibleai/responsibleai/_tools/causal/causal_result.py`

The comparisons used are based on common comparison abbreviations. One example of their use is in MongoDB query comparisons: https://docs.mongodb.com/manual/reference/operator/query-comparison/

### DIFF
--- a/responsibleai/responsibleai/_interfaces.py
+++ b/responsibleai/responsibleai/_interfaces.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 
 from enum import Enum
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List, Optional, Union
 
 
 class TaskType(str, Enum):
@@ -18,7 +18,7 @@ class Dataset:
     probability_y: List
     true_y: List
     class_names: List[str]
-    categorical_features: [str]
+    categorical_features: List[str]
     target_column: str
 
 
@@ -94,7 +94,8 @@ class CausalPolicyTreeLeaf:
 class CausalPolicyTreeInternal:
     leaf: False
     feature: str
-    threshold: Union[float, str]  # TODO: Categorical features
+    category: Optional[str]
+    threshold: Union[float, str]
     left: Union['CausalPolicyTreeInternal', CausalPolicyTreeLeaf]
     right: Union['CausalPolicyTreeInternal', CausalPolicyTreeLeaf]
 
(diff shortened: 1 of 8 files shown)


responsibleai/responsibleai/_interfaces.py
```diff
@@ -94,7 +94,8 @@ class CausalPolicyTreeLeaf:
 class CausalPolicyTreeInternal:
     leaf: False
     feature: str
-    threshold: Union[float, str]  # TODO: Categorical features
+    category: Optional[str]
```
xuke444, line 97: so category and threshold are exclusive here, correct?

I was think of adding a sort of comparation type field and keep threshold as is.
For current regression, it will be type = "SmallerThan" threshold = 123
Categorical feature: type = "Equal" threshold="class 1"

So that we can expand in future to support other type, eg: in multiple category,

gregorybchris: @xuke444 I've addressed your comment. Please help me consider the back-compat ramifications that changing this might have on the UI.

xuke444: There should not be any issue, ui will still shows same thing as before, we will need to make code change to utilize new field.

responsibleai/responsibleai/_interfaces.py
```diff
@@ -91,10 +91,22 @@ class CausalPolicyTreeLeaf:
     treatment: str
 
 
+class ComparisonTypes:
```
xuke444, line 94: Let's check with @imatiach-msft , current ui and error analysis line up with different set of enum.
We need to standardize to one set, I am not 100% sure about impact.

gregorybchris: Discussed offline, agreed that the causal changes can go in first with changes to error analysis to standardize later if necessary

responsibleai/responsibleai/_interfaces.py
gregorybchris: Great, thanks!

PR #813, opened 2021-08-20
Extend unseen categorical check to counterfactuals

gaugup:


responsibleai/responsibleai/_managers/causal_manager.py
```diff
@@ -3,13 +3,13 @@
...
 """Manager for causal analysis."""
 import numpy as np
-import pandas as pd
 
 from typing import Any, List, Optional, Union
 
 from econml.solutions.causal_analysis import CausalAnalysis
 from pathlib import Path
 
+from responsibleai._data_validations import _validate_train_test_categories
```
gregorybchris, line 12: If we're going to use `_validate_train_test_categories` outside of `_data_validations.py`, then we should make it `validate_train_test_categories` with no underscore

gregorybchris (approved):

gregorybchris:
@gaugup will you explain why counterfactual should validate this? Which scenario are we in?

1. DiceML fails on unknown categories -> need to validate
2. DiceML does not fail on unknown categories, but the results are invalid for some reason -> need to validate
3. DiceML appropriately ignores/drops unknown categories -> no validation needed

responsibleai/responsibleai/_managers/counterfactual_manager.py
```diff
@@ -7,6 +7,7 @@
 from dice_ml import Dice
 import numpy as np
 
+from responsibleai._data_validations import _validate_train_test_categories
```
gregorybchris, line 10: FYI Looks like maybe you forgot to `git add` this file

responsibleai/responsibleai/_managers/causal_manager.py
gaugup: Renamed it.

responsibleai/responsibleai/_managers/counterfactual_manager.py
gaugup: Done.

gaugup:
> @gaugup will you explain why counterfactual should validate this? Which scenario are we in?
> 
> 1. DiceML fails on unknown categories -> need to validate
> 2. DiceML does not fail on unknown categories, but the results are invalid for some reason -> need to validate
> 3. DiceML appropriately ignores/drops unknown categories -> no validation needed

Yes, the recent version of dice-ml 0.7 added a check which would fail the CFE computation if the query instance carries a categorical value which is not present in training data. So the first one is True.

PR #834, opened 2021-08-30
refactor matrix code and fix display issue for categoricals

imatiach-msft:


erroranalysis/erroranalysis/_internal/matrix_filter.py
```diff
@@ -229,6 +229,7 @@ def compute_matrix(analyzer, features, filters, composite_filters,
             counts_err = counts_err.values.ravel()
         matrix = matrix_1d(categories, val_err, counts,
                            counts_err, metric)
+        print(matrix)
```
xuke444, line 232: seems for debug, remove?

imatiach-msft: ah yes I will remove

imatiach-msft: done!

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Matrix/MatrixFooter/MatrixFooter.tsx
```diff
@@ -32,13 +34,13 @@ export class MatrixFooter extends React.PureComponent<IMatrixFooterProps> {
                 key={`${this.props.matrixLength}_${0}category1`}
                 className={classNames.matrixCellPivot1Categories}
               />
-              {this.props.category1Values.map((category: any, i: number) => {
+              {this.props.category1Values.map((_: any, i: number) => {
```
xuke444, line 37: this part seems making more sense to extract as a react component.
so it can be used everywhere formatCategory is referenced.

formatCategory is more like a displaying processor, which should be better to sits in an tsx file

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Matrix/MatrixUtils.ts
```diff
@@ -0,0 +1,33 @@
...
+  intervalMax?: never;
+  intervalMin?: never;
+};
+export type IMatrixSingleRegressionCategory = {
+  value: number;
+  intervalMax: number;
+  intervalMin: number;
+};
+
+export function formatCategory(
```
xuke444, line 19: this looks like a display process function, extract as a react component could be better than define a ts function.

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Matrix/MatrixFooter/MatrixFooter.tsx
imatiach-msft: done, made formatCategory a class/tsx file

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Matrix/MatrixUtils.ts
imatiach-msft: done, made formatCategory a class/tsx file

xuke444 (approved):

PR #838, opened 2021-08-31
Fix click on info button shifts graphic

zhb000:
Causal UI Bug Bash: click on info button shifts graphic


libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalCallouts/causalCalloutDictionary.ts
```diff
@@ -0,0 +1,26 @@
...
+
+interface ICausalInfo {
+  title: string;
+  description: string;
+  linkUrl?: string;
+}
+
+const whyConfounding: ICausalInfo = {
+  description: localization.CausalAnalysis.AggregateView.confoundingFeature,
+  linkUrl: "https://www.microsoft.com/en-us/research/project/econml/#!how-to",
```
xuke444, line 20: This probably good to put in localization as well. so when other languages added there we can onboard

xuke444 (approved):

PR #845, opened 2021-09-07
Bug Fixes

yongjiaaaa:
Fixed 4 bugs:
1. renamed the component names in the Dashboard setting to have spaces in between words and also not have the word “Tab” in it since that’s confusing
image
image
2. removed the grey line at the top of the nav ba
image
3. fixed the issue of cohort selector dropdown text cutting
image
4. changed the font sizes of headers to size 28 font, such as "error analysis" etc.
image


libs/model-assessment/src/lib/ModelAssessmentDashboard/ModelAssessmentEnums.ts
```diff
@@ -2,11 +2,11 @@
...
 
 export enum GlobalTabKeys {
-  DataExplorerTab = "DataExplorerTab",
-  FeatureImportancesTab = "FeatureImportancesTab",
-  ErrorAnalysisTab = "ErrorAnalysisTab",
-  FairnessTab = "FairnessTab",
-  ModelStatisticsTab = "ModelStatisticsTab",
-  CausalAnalysisTab = "CausalAnalysisTab",
-  CounterfactualsTab = "CounterfactualsTab"
+  DataExplorerTab = "Data Explorer",
```
xuke444, line 5: Are we display these values anythere?
If yes, must be a bug.
We should not change the value, but display different work from loalization.

libs/interpret/src/lib/MLIDashboard/Controls/WhatIfTab/WhatIfTab.tsx
```diff
@@ -355,7 +355,18 @@ export class WhatIfTab extends React.PureComponent<
                   {localization.Interpret.WhatIfTab.cohortPickerLabel}
                 </Text>
                 <Dropdown
-                  styles={{ dropdown: { width: 150 } }}
+                  styles={{
```
xuke444, line 358: probably better to move to style file.

libs/model-assessment/src/lib/ModelAssessmentDashboard/ModelAssessmentEnums.ts
yongjiaaaa: please check my newest push

xuke444 (approved):

PR #853, opened 2021-09-09
Version causal results to enable back-compat checks

gregorybchris:
This PR adds versioning and schema validation to the `CausalResult` object so that back-compat can be tracked as the format of `CausalResult` changes over time.

- Factor all common result serialization code out of `CausalResult` into `BaseResult`.
- Create new `id.json` and `version.json` files when serializing RAI Result objects with common utilities for serializing any future metadata to disk.
- Read in `dashboard.json` file from disk when rehydrating a `CausalResult` so that the format does not change if a new SDK is used to load an old result.
- Add JSON schema validation when loading `CausalResult` objects from disk.

### DIFF
--- a/responsibleai/requirements.txt
+++ b/responsibleai/requirements.txt
@@ -7,6 +7,7 @@ numpy>=1.17.2
 pandas>=0.25.1
 scikit-learn>=0.22.1
 scipy>=1.4.1
+semver~=2.13.0
 
 # Pinned dependencies
 networkx<=2.5
--- a/responsibleai/responsibleai/_tools/causal/causal_result.py
+++ b/responsibleai/responsibleai/_tools/causal/causal_result.py
@@ -17,21 +17,19 @@ from responsibleai._interfaces import (
     CausalPolicyGains, CausalPolicyTreeInternal, CausalPolicyTreeLeaf,
     ComparisonTypes)
 from responsibleai._tools.causal.causal_config import CausalConfig
-from responsibleai._tools.shared.attribute_serialization import (
-    SerializationFormat)
 from responsibleai._tools.shared.base_result import BaseResult
 
 
 class CausalResult(BaseResult['CausalResult']):
     """Result of causal analysis."""
 
-    _ATTRIBUTES = {
-        'config': SerializationFormat.PICKLE,
-        'causal_analysis': SerializationFormat.PICKLE,
-        'global_effects': SerializationFormat.PICKLE,
-        'local_effects': SerializationFormat.PICKLE,
-        'policies': SerializationFormat.PICKLE,
-    }
+    _ATTRIBUTES = [
+        'config',
+        'causal_analysis',
+        'global_effects',
+        'local_effects',
+        'policies',
+    ]
 
     def __init__(
         self,
--- a/responsibleai/responsibleai/_tools/causal/causal_result.py
+++ b/responsibleai/responsibleai/_tools/causal/causal_result.py
@@ -3,44 +3,45 @@
 """Result of causal analysis."""
 
 import json
+import pandas as pd
 import re
 import uuid
 
 from pathlib import Path
-from typing import Any, Optional, Tuple, Union
+from typing import Any, List, Optional, Tuple, Union
 
 from responsibleai._tools.causal.causal_constants import (
-    ResultAttributes, SerializationAttributes)
+    ResultAttributes, Versions)
 from responsibleai._interfaces import (
     CausalConfig as CausalConfigInterface, CausalData, CausalPolicy,
     CausalPolicyGains, CausalPolicyTreeInternal, CausalPolicyTreeLeaf,
     ComparisonTypes)
 from responsibleai._tools.causal.causal_config import CausalConfig
 from responsibleai._tools.shared.attribute_serialization import (
-    load_attributes, save_attributes)
-from responsibleai.serialization_utilities import serialize_json_safe
+    SerializationFormat)
+from responsibleai._tools.shared.base_result import BaseResult
 
 
-class CausalResult:
+class CausalResult(BaseResult['CausalResult']):
     """Result of causal analysis."""
 
-    _ATTRIBUTES = [
-        'config',
-        'causal_analysis',
-        'global_effects',
-        'local_effects',
-        'policies'
-    ]
+    _ATTRIBUTES = {
+        'config': SerializationFormat.PICKLE,
+        'causal_analysis': SerializationFormat.PICKLE,
+        'global_effects': SerializationFormat.PICKLE,
+        'local_effects': SerializationFormat.PICKLE,
+        'policies': SerializationFormat.PICKLE,
+    }
 
     def __init__(
         self,
         config: Optional[CausalConfig] = None,
         causal_analysis: Optional[Any] = None,
-        global_effects: Optional[Any] = None,
-        local_effects: Optional[Any] = None,
+        global_effects: Optional[pd.DataFrame] = None,
+        local_effects: Optional[pd.DataFrame] = None,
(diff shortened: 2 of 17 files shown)


responsibleai/requirements.txt
```diff
@@ -7,6 +7,7 @@ numpy>=1.17.2
 pandas>=0.25.1
 scikit-learn>=0.22.1
 scipy>=1.4.1
+semver~=2.13.0
```
riedgar-ms, line 10: Looks like this is only used in test, not product code? If so, shouldn't it be in requirements-dev?

responsibleai/responsibleai/_tools/shared/base_result.py
```diff
@@ -0,0 +1,149 @@
...
+
+        # Save dashboard data
+        self._save_dashboard(result_dir)
+
+    def _save_metadata(self, result_dir: Path):
+        """Save result metadata to disk.
+
+        :param path: Path to result directory on disk.
+        """
+        id_path = result_dir / SerializationAttributes.ID_FILENAME
```
riedgar-ms, line 57: Just curious: why make the ID and version files separate? This isn't a particular objection for now, but in future, might they be combined in an `MLCausal` YAML file?

gregorybchris: This will have back-compat implications, so ideas for combining them are totally welcome. What else would the MLCausal YAML include @riedgar-ms?

riedgar-ms: I've not thought that far yet. `MLCausal` `MLExplainer` etc. are still rather vague ideas

responsibleai/requirements.txt
gregorybchris: My original intention was to do full power semver ranging checking in the package code, not just the tests. Looks like I stopped short of that, will keep this comment open until I add that

responsibleai/responsibleai/_tools/shared/base_result.py
riedgar-ms: I don't think it's worth holding up this PR for that. As you say, there are potential back-compat implications, but I don't think thoughts are sufficiently fixed to avoid them.

The `MLModel` file is shown on this page:
https://www.mlflow.org/docs/latest/models.html

responsibleai/responsibleai/_tools/causal/causal_result.py
```diff
@@ -3,44 +3,44 @@
...
     ComparisonTypes)
 from responsibleai._tools.causal.causal_config import CausalConfig
-from responsibleai._tools.shared.attribute_serialization import (
-    load_attributes, save_attributes)
-from responsibleai.serialization_utilities import serialize_json_safe
+from responsibleai._tools.shared.base_result import BaseResult
 
 
-class CausalResult:
+class CausalResult(BaseResult['CausalResult']):
```
imatiach-msft, line 24: what does this do:
```BaseResult['CausalResult']```
I'm not familiar with this syntax

File: responsibleai/responsibleai/_tools/causal/dashboard_schemas/schema_0.1.0.json [hunk 4]
```diff
@@ -0,0 +1,220 @@
...
+          "description": "Number of local sample (only for local effects)",
+          "type": "number"
+        },
+        "feature": {
+          "description": "Treatment feature name",
+          "type": ["string", "number"]
+        },
+        "feature_value": {
+          "description": "Treatment feature value",
+          "type": "string"
```
imatiach-msft, line 68: it can only be a string value?

responsibleai/responsibleai/_tools/causal/causal_result.py
gregorybchris: It's type generics: https://docs.python.org/3/library/typing.html#generics, allows the Python type system to understand the inheritance model of your code with type parameters.

gregorybchris: I found the mypy docs really helpful for understanding generic type variance: https://mypy.readthedocs.io/en/stable/generics.html#variance-of-generic-types

responsibleai/responsibleai/_tools/shared/base_result.py
```diff
@@ -0,0 +1,149 @@
...
+        See responsibleai/_interfaces.py for existing interfaces.
+        """
+        pass
+
+    def _get_dashboard_data(self):
+        """Get the Python dict representation of the dashboard object."""
+        if self._dashboard_data is None:
+            dashboard_object = self._get_dashboard_object()
+            self._dashboard_data = serialize_json_safe(dashboard_object)
+            return self._dashboard_data
```
imatiach-msft, line 138: tiny nit: remove return self._dashboard_data here, you are already do that in the not-if condition anyway

responsibleai/responsibleai/_tools/causal/dashboard_schemas/schema_0.1.0.json
```diff
@@ -0,0 +1,220 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
imatiach-msft, line 2: will we be removing this schema from azureml-responsibleai package now?  The schema checks will always be done in OSS now?

responsibleai/responsibleai/_tools/causal/dashboard_schemas/schema_0.1.0.json
```diff
@@ -0,0 +1,220 @@
...
+  "title": "AzureML Responsible AI causal analysis dashboard schema",
```
imatiach-msft, line 3: probably should remove "AzureML Responsible AI" since this is OSS and not tied to AzureML in any way

imatiach-msft (approved):

responsibleai/responsibleai/_tools/causal/dashboard_schemas/schema_0.1.0.json
gregorybchris: Oh definitely!! Thanks for catching this

responsibleai/responsibleai/_tools/causal/dashboard_schemas/schema_0.1.0.json
gregorybchris: Yep! Currently working on another branch to remove it

responsibleai/responsibleai/_tools/shared/base_result.py
gregorybchris: Good call! Done

responsibleai/responsibleai/_tools/causal/dashboard_schemas/schema_0.1.0.json
gregorybchris: You're right, it can be numeric as well. Thanks for catching that

responsibleai/responsibleai/_tools/causal/causal_result.py
imatiach-msft: shouldn't that be BaseResult[CausalResult] and not BaseResult['CausalResult'] according to that doc, eg the quotes should be removed?

imatiach-msft (approved):

responsibleai/responsibleai/_tools/causal/causal_result.py
gregorybchris: That's only if the class has already been defined. If a type is being used in the class that defined it you need to use quotes

responsibleai/requirements.txt
gregorybchris: Added semver checking

PR #859, opened 2021-09-10
Add soft limit to test data in ModelAnalysis

riedgar-ms:
Performance of `ModelAnalysis` tends to fall off past 5k rows in the test data, so add a soft limit on this.


responsibleai/responsibleai/modelanalysis/model_analysis.py
```diff
@@ -216,6 +223,12 @@ def _validate_model_analysis_input_parameters(
                     'The serializer should be serializable via pickle')
 
         if isinstance(train, pd.DataFrame) and isinstance(test, pd.DataFrame):
+            if test.shape[0] > maximum_rows_for_test:
+                raise UserConfigValidationException(
```
gaugup, line 227: Can you add the number of rows in the test as well?

responsibleai/responsibleai/modelanalysis/model_analysis.py
```diff
@@ -72,7 +72,8 @@ class ModelAnalysis(object):
 
     def __init__(self, model, train, test, target_column,
                  task_type, categorical_features=None, train_labels=None,
-                 serializer=None):
+                 serializer=None,
+                 maximum_rows_for_test: int = 5000):
```
gaugup, line 76: Can you do away with the spaces around the '='?

responsibleai/responsibleai/modelanalysis/model_analysis.py
```diff
@@ -156,8 +160,9 @@ def _get_classes(task_type, train, target_column, train_labels):
 
     def _validate_model_analysis_input_parameters(
             self, model, train, test, target_column,
-            task_type, categorical_features=None, train_labels=None,
-            serializer=None):
+            task_type, categorical_features, train_labels,
```
gaugup, line 163: Why have you changed the optional parameters to this function?

responsibleai/tests/test_model_analysis_validations.py
```diff
@@ -35,6 +35,25 @@ def test_validate_unsupported_task_type(self):
...
+    def test_validate_test_data_size(self):
+        X_train, X_test, y_train, y_test, _, _ = \
+            create_iris_data()
+
+        model = create_lightgbm_classifier(X_train, y_train)
+        X_train['target'] = y_train
+        X_test['target'] = y_test
+
+        with pytest.raises(UserConfigValidationException) as ucve:
+            ModelAnalysis(
```
gaugup, line 47: Can you check the message in the exception as well here?

gaugup (changes requested):

responsibleai/responsibleai/modelanalysis/model_analysis.py
gregorybchris: I think spaces around the `=` is standard when adding type hints?

gregorybchris: Might be more clear if there are type hints on all params

riedgar-ms: The space around the `=` is indeed a `flake8`ism. I agree that the type hints would be nice.... which is why I added one for the parameter I've added. The rest are.... out of scope for this.

responsibleai/responsibleai/modelanalysis/model_analysis.py
riedgar-ms: Because it's internal and only called from the constructor (which already has the defaults making them pointless here). Having defaults just makes it harder to add arguments (adding on the end means they need defaults, adding them earlier would change the order from the constructor).

responsibleai/responsibleai/modelanalysis/model_analysis.py
riedgar-ms: Good idea. Done.

responsibleai/tests/test_model_analysis_validations.py
riedgar-ms: I do below.

responsibleai/responsibleai/modelanalysis/model_analysis.py
```diff
@@ -216,6 +225,14 @@ def _validate_model_analysis_input_parameters(
                     'The serializer should be serializable via pickle')
 
         if isinstance(train, pd.DataFrame) and isinstance(test, pd.DataFrame):
+            if test.shape[0] > maximum_rows_for_test:
+                msg_fmt = 'The test data has {0} rows, ' +\
```
imatiach-msft, line 229: can you make the message more actionable, specifically by stating something like:
"Please downsample your dataset to have fewer than 5000 rows or increase maximum_rows_for_test parameter value"

imatiach-msft (approved):

responsibleai/responsibleai/modelanalysis/model_analysis.py
riedgar-ms: Done.

gaugup (approved):

PR #864, opened 2021-09-13
add more metrics to error analysis

imatiach-msft:
- add precision and recall metrics to error analysis for classification scenario, including macro and micro versions for multiclass case (macro averaged precision, micro averaged prevision, macro averaged recall, micro averaged recall)
- add mean absolute error metric to error analysis for regression scenario
image

image


erroranalysis/erroranalysis/_internal/matrix_filter.py
```diff
@@ -292,6 +376,24 @@ def matrix_2d(categories1, categories2, matrix_counts,
...
                 if index_exists_err and col_exists_err:
                     false_count = int(matrix_err_counts.loc[cat1, cat2])
+            elif is_precision:
+                tp_sum = []
+                pred_sum = []
+                metric_value = 0
+                if index_exists_err and col_exists_err:
+                    metric_value = float(matrix_err_counts.loc[cat1, cat2][0])
+                    tp_sum = matrix_err_counts.loc[cat1, cat2][1]
+                    pred_sum = matrix_err_counts.loc[cat1, cat2][2]
```
imatiach-msft, line 386: I'm not sure if it is better to do it this way or just to pass the tp/tn/fn/fp (always?) in the matrix and then have the UI use that for calculating the metrics

gregorybchris: We've dealt with this in AutoML and made the choice to keep all 4 values for ease of use. You can get away with storing very few values if you store aggregates as well, but it just makes it harder to reason about, especially if users have direct access to the data.

gregorybchris (approved):

erroranalysis/erroranalysis/_internal/matrix_filter.py
```diff
@@ -268,13 +306,59 @@ def _agg_func_triplet(self, pair):
...
+        return 0
+
+
+class _MultiMetricAggFunc(_BaseAggFunc):
+    def __init__(self, aggfunc, metric, labels):
+        super(_MultiMetricAggFunc, self).__init__(aggfunc)
+        self.metric = metric
+        if len(labels) == 2:
+            # for binary classification case, choose positive class label
+            self.labels = [labels[1]]
```
gregorybchris, line 319: In the future we may want to make it possible to select the positive class. This is fine for now

erroranalysis/erroranalysis/_internal/matrix_filter.py
imatiach-msft: done, store all tp, fp, fn, tn now

PR #870, opened 2021-09-14
Add widgets extra to responsibleai

gregorybchris:


responsibleai/setup.py
```diff
@@ -20,6 +20,12 @@
 with open('requirements.txt') as f:
     install_requires = [line.strip() for line in f]
 
+EXTRAS = {
```
gaugup, line 23: I thought this extras was for azureml only? Are we taking dependency on visualization package from a purely compute package?

gregorybchris: Don't we want users to be able to install compute and viz in one command to make sure they get the right versions? The same problem that happens for AML notebooks also happens for OSS notebooks

gregorybchris: **OSS notebook requirements:**
```python
old = {
  responsibleai==0.10.0,
  raiwidgets==0.10.0
}

new = {
  responsibleai[widgets]==0.10.0
}
```

**AML notebook requirements:**
```python
old = {
  azureml-responsibleai==1.34.0,
  raiwidgets==0.10.0
}

new = {
  azureml-responsibleai[widgets]==1.34.0
}
```

gregorybchris: In both cases you don't need to know about the correct version of widgets at install time anymore since our setup.py handles it

riedgar-ms: This does feel a bit odd, although I take the point about ensuring versions are synced. What has driven this, beyond a general desire to streamline the flow? You could argue that the widgets should pull in `responsibleai`

gaugup: I dont what the rule on the extras is. I thought that the parent package has some code to light up scenario on the extras dependency. Just feels really weird to take some dependency under extras. 

Even I think about AML, it doesn't feel correct to introduce an extra to bypass the notebook infra issue.

gregorybchris:
Closing this PR since it's not required at this moment

PR #895, opened 2021-09-21
Log warning when regression model has predict_proba()

gaugup:


gregorybchris (approved):

responsibleai/responsibleai/modelanalysis/model_analysis.py
```diff
@@ -309,6 +309,13 @@ def _validate_model_analysis_input_parameters(
                             ' getting predictions via predict_proba()'
                         )
 
+                if task_type == ModelTask.REGRESSION:
```
gregorybchris, line 312: Might consider making this an `elif` of the condition above. Otherwise looks good

gaugup: Keeping this as a spearate if.

responsibleai/responsibleai/modelanalysis/model_analysis.py
```diff
@@ -309,6 +309,13 @@ def _validate_model_analysis_input_parameters(
...
+                    if hasattr(model, 'predict_proba'):
```
imatiach-msft, line 313: can you use the constant from here:

https://github.com/microsoft/responsible-ai-widgets/blob/850bf429f20b5401b7907e249d219dbd5d97f6c9/responsibleai/responsibleai/_internal/constants.py#L56

responsibleai/responsibleai/modelanalysis/model_analysis.py
imatiach-msft: nit: I think using and is better than two ifs, but please ignore if you prefer this way

```if task_type == ModelTask.REGRESSION and hasattr(model, PREDICT_PROBA):```

imatiach-msft (approved):

PR #907, opened 2021-09-24
Raise exception when users configures less than required counterfactuals

gaugup:
dice-ml needs at least ten counterfactuals to be computed to compute feature importance. Hence, the exception.

### DIFF
--- a/responsibleai/responsibleai/_managers/counterfactual_manager.py
+++ b/responsibleai/responsibleai/_managers/counterfactual_manager.py
@@ -159,6 +159,14 @@ class CounterfactualManager(BaseManager):
                     'The desired_range should not be None'
                     ' for regression scenarios.')
 
+        if new_counterfactual_config.feature_importance and\
+                new_counterfactual_config.total_CFs < 10:
+            raise UserConfigValidationException(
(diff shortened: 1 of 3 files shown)


responsibleai/responsibleai/_managers/counterfactual_manager.py
```diff
@@ -159,6 +159,14 @@ def _add_counterfactual_config(self, new_counterfactual_config):
                     'The desired_range should not be None'
                     ' for regression scenarios.')
 
+        if new_counterfactual_config.feature_importance and\
```
imatiach-msft, line 162: maybe instead of using \ you can break it into two variables:
```
has_feat_imps = new_counterfactual_config.feature_importance
total_CFs_check = new_counterfactual_config.total_CFs < 10
if has_feat_imps  and total_CFs_check:
  ...
```

imatiach-msft (approved):

responsibleai/responsibleai/_managers/counterfactual_manager.py
gaugup: would probably keep the check the same.

PR #909, opened 2021-09-27
fix handling of numeric features names in pandas for error analysis tree view

imatiach-msft:
A customer who did not pass their feature names to the explainer and generated an explanation without features hit an exception in error analysis dashboard due to numeric feature names.  This PR allows error analysis dashboard to work with numeric feature names and adds an end to end test.

### DIFF
--- a/erroranalysis/erroranalysis/_internal/surrogate_error_tree.py
+++ b/erroranalysis/erroranalysis/_internal/surrogate_error_tree.py
@@ -3,6 +3,7 @@
 
 import numpy as np
 import pandas as pd
+import numbers
 from lightgbm import LGBMClassifier, LGBMRegressor
 from enum import Enum
 from erroranalysis._internal.cohort_filter import filter_from_cohort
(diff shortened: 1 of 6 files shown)


imatiach-msft:
note: tests will fail on the PR until I release a new version of error analysis with the fixes in the PR

erroranalysis/erroranalysis/_internal/surrogate_error_tree.py
```diff
@@ -326,6 +328,15 @@ def node_to_dict(df, tree, nodeid, categories, json,
...
         p_node_name = feature_names[parent[SPLIT_FEATURE]]
+        # use number.Integral to check for any numpy or python number type
+        if isinstance(p_node_name, numbers.Integral):
+            # for numeric column names, we can use @df[numeric_colname] syntax
+            p_node_query = "@df[" + str(p_node_name) + "]"
+        else:
+            # for string column names, we can just use column name directly
+            # with backticks
+            p_node_query = "`" + str(p_node_name) + "`"
+        p_node_name = str(p_node_name)
```
gaugup, line 339: IS this something we should support? We could also catch this condition and raise a user error if we think that is something a user can fix in their data.

imatiach-msft: yes, we definitely should, it's valid to have numeric feature names - both this customer and actually Ke at some point earlier hit this issue.  I don't think we should throw (that's what we did earlier) since it would block customers from running a valid scenario?  What do you think?

gaugup: Sure sounds good.

gaugup (approved):

PR #912, opened 2021-09-27
Rename wrapped-flask to rai_core_flask

gaugup:


.github/workflows/CI-python.yml
```diff
@@ -13,20 +13,20 @@ jobs:
     strategy:
```
imatiach-msft, line 13: can you add the version bump as well to the PR?

imatiach-msft (approved):

.github/workflows/CI-python.yml
gaugup: Version bump for rai_core_flask?

xuke444 (approved):

.github/workflows/CI-python.yml
gaugup: Spoke offline to Ilya. We need not update the version for this PR as there are no namepsace changes.

PR #914, opened 2021-09-28
update cohort stats

xuke444:


imatiach-msft:
Could you add a description and maybe some screenshots?  It's hard to review code without understanding the context and the changes.

imatiach-msft (approved):

PR #915, opened 2021-09-29
Change regression current class label

zhb000:
Change regression current class label

### DIFF
--- a/libs/counterfactuals/src/lib/CounterfactualChart.tsx
+++ b/libs/counterfactuals/src/lib/CounterfactualChart.tsx
@@ -326,7 +326,7 @@ export class CounterfactualChart extends React.PureComponent<
                 />
                 <div className={classNames.legendLabel}>
                   <b>{`${localization.Counterfactuals.currentClass}: `}</b>
-                  {this.getCurrentClass()}
+                  {this.getCurrentLabel()}
                 </div>
                 <PrimaryButton
                   className={classNames.legendLabel}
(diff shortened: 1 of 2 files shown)


libs/counterfactuals/src/lib/CounterfactualChart.tsx
```diff
@@ -379,13 +379,14 @@ export class CounterfactualChart extends React.PureComponent<
   }
 
   private getCurrentClass(): string {
```
imatiach-msft, line 381: minor nit: maybe getCurrentClass should be renamed to something else, like getCurrentLabel, since in the regression case classes don't apply like in the classification scenario?

libs/counterfactuals/src/lib/CounterfactualChart.tsx
```diff
@@ -379,13 +379,14 @@ export class CounterfactualChart extends React.PureComponent<
...
-    if (this.props.data.desired_class) {
-      return this.props.data.desired_class;
-    }
-    if (this.props.data.desired_range) {
-      return this.props.data.desired_range.join("->");
+    if (this.context.dataset.task_type === "regression") {
+      return (
+        (this.props.data.desired_range &&
```
imatiach-msft, line 384: is it possible to add some kind of tooltip as well to explain what this means in the what if tool?  that can be a separate PR and maybe might need some PM input

zhb000: It could cause inconsistent if just add tooltip for regression cases, 
Also, the prefix "Current class" maybe need to rephrased, as in regression it's a range instead of specific class?
@minthigpen any idea?

imatiach-msft (approved):

PR #983, opened 2021-10-28
Add Cognitive Service fairness assessment notebooks 

LeJit:
Adds two notebooks for fairness assessment of face validation systems and speech-to-text systems.


imatiach-msft:
@LeJit it looks like there are still some flake8 errors:

image

imatiach-msft:
@LeJit I've merged the PR, but we will need to move the datasets out at some point

PR #499, opened 2021-05-06
Causal analysis manager for responsibleai package

gregorybchris:
This PR adds causal analysis to the responsibleai package. It uses EconML to compute causal effects and implements the method to add a new config and compute the causal effects from an existing config.

PR #994, opened 2021-11-03
Remove setup_explainer() after load() call

gaugup:
Looks like the setup_explainer() is not needed after ModelAnalysis.load().


imatiach-msft (approved):
It looks like it was added as part of this PR:
https://github.com/microsoft/responsible-ai-widgets/pull/499
I have no idea why.  I thought it might be a bugfix.  Maybe it's good to keep it regardless just to make sure on deserialize those methods can still be called and have the correct behavior.

gaugup:
> It looks like it was added as part of this PR: #499 I have no idea why. I thought it might be part of a bugfix, to validate some functionality. Maybe it's good to keep it regardless just to make sure on deserialize those methods can still be called and have the correct behavior.

Ah makes sense now. For explainer manager since we don't allow the any other config so this looked redundant. I will move the check inside the explainer manager if clause in that case as it makes more sense there.

PR #1005, opened 2021-11-10
display message bar after saving a cohort in error analysis dashboard

imatiach-msft:
Fixes issue: https://github.com/microsoft/responsible-ai-widgets/issues/242
Display message bar after saving a cohort in error analysis dashboard.

Sample screenshot in test app:
image

### DIFF
--- a/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Navigation/Navigation.styles.ts
+++ b/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Navigation/Navigation.styles.ts
@@ -9,16 +9,36 @@ import {
 } from "office-ui-fabric-react";
 
 export interface INavigationStyles {
+  breadcrumbItemWidth: IStyle;
+  content: IStyle;
+  dismissal: IStyle;
+  dismissSingleLine: IStyle;
+  iconContainer: IStyle;
   navigation: IStyle;
-  breadcrumb: IStyle;
+  root: IStyle;
+  text: IStyle;
 }
 
 export const navigationStyles: () => IProcessedStyleSet<INavigationStyles> =
   () => {
     const theme = getTheme();
     return mergeStyleSets<INavigationStyles>({
-      breadcrumb: {
-        padding: "0px 0px 0px 10px"
+      breadcrumbItemWidth: {
+        width: "500px"
+      },
+      content: {
+        margin: "0"
+      },
+      dismissal: {
+        height: "24px",
+        padding: "0"
+      },
+      dismissSingleLine: {
+        margin: "0"
+      },
+      iconContainer: {
+        height: "20px",
+        margin: "4px 8px 0 12px"
       },
       navigation: {
         backgroundColor: theme.palette.white,
@@ -27,6 +47,15 @@ export const navigationStyles: () => IProcessedStyleSet<INavigationStyles> =
         color: theme.palette.black,
         height: "25px",
         width: "100%"
+      },
+      root: {
+        height: "24px",
+        margin: "0",
+        minHeight: "24px"
+      },
+      text: {
+        height: "24px",
+        margin: "4px 0 0"
       }
     });
   };
--- a/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Navigation/Navigation.tsx
+++ b/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Navigation/Navigation.tsx
@@ -7,9 +7,13 @@ import {
   IBreadcrumbItem,
   IBreadcrumbStyleProps,
   IBreadcrumbStyles,
+  IMessageBarStyles,
+  IRenderFunction,
   IStyleFunctionOrObject,
   Link,
-  IRenderFunction
+  MessageBar,
+  MessageBarType,
+  Stack
 } from "office-ui-fabric-react";
 import React from "react";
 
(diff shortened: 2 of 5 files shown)


zhb000 (approved):
Nit: for 0 value, you don't need add px there, could just make margin: 0
margin: "0px 8px 0px 12px" => margin: "0 8px 0 12px" to save some bits

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Navigation/Navigation.styles.ts
```diff
@@ -20,13 +26,34 @@ export const navigationStyles: () => IProcessedStyleSet<INavigationStyles> =
       breadcrumb: {
         padding: "0px 0px 0px 10px"
       },
+      content: {
+        margin: "0px 0px 0px 0px"
```
RubyZ10, line 30: nit: if all four margins have same value, can just do margin: 0px

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Navigation/Navigation.tsx
```diff
@@ -102,18 +108,42 @@ export class Navigation extends React.Component<INavigationProps> {
...
+              />
+            </div>
+          </Stack.Item>
+          {this.props.showMessageBar && (
+            <Stack.Item>
+              <MessageBar
+                messageBarType={MessageBarType.success}
+                isMultiline={false}
+                onDismiss={() => this.props.closeMessageBar()}
+                className={classNames.messageBar}
```
RubyZ10, line 139: why do we need both className and styles? can we merge them to one?

libs/error-analysis/src/lib/ErrorAnalysisDashboard/ErrorAnalysisDashboard.tsx
```diff
@@ -805,7 +812,8 @@ export class ErrorAnalysisDashboard extends React.PureComponent<
     this.setState((prevState) => ({
       baseCohort: switchNew ? savedCohort : prevState.baseCohort,
       cohorts: newCohorts,
-      selectedCohort: switchNew ? savedCohort : prevState.selectedCohort
+      selectedCohort: switchNew ? savedCohort : prevState.selectedCohort,
+      showMessageBar: true
```
RubyZ10, line 816: why we don't need to check if it is switchNew or not? just for my understanding

RubyZ10 (approved):

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Navigation/Navigation.tsx
vinuthakaranth: +1 on this

vinuthakaranth (approved):

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Navigation/Navigation.styles.ts
imatiach-msft: done!

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Navigation/Navigation.tsx
```diff
@@ -102,18 +108,42 @@ export class Navigation extends React.Component<INavigationProps> {
...
-            maxDisplayedItems={10}
-            ariaLabel="Navigation"
-            overflowAriaLabel="More links"
-            onRenderItem={this._onRenderItem}
-            styles={breadcrumbStyle}
-          />
-        </div>
+        <Stack horizontal horizontalAlign="space-between">
+          <Stack.Item>
+            <div className={classNames.breadcrumb}>
```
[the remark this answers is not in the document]
imatiach-msft: I removed the div but I also added another breadcrumbItemWidth to specify width of breadcrumb, otherwise it gets minized with the space-between horizontalAlign property.

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Navigation/Navigation.tsx
imatiach-msft: done!  Moved the className style values to be the root of the styles property.  Now className is removed.

libs/error-analysis/src/lib/ErrorAnalysisDashboard/ErrorAnalysisDashboard.tsx
imatiach-msft: Ah, so you can save a cohort but not switch to it.  If you just save and don't switch, we keep the current base/selected cohorts the same.  If you save and switch to the new one, then we set the base and selected cohort to be the new one.

libs/error-analysis/src/lib/ErrorAnalysisDashboard/ErrorAnalysisDashboard.tsx
```diff
@@ -380,6 +381,8 @@ export class ErrorAnalysisDashboard extends React.PureComponent<
             viewType={this.state.viewType}
             activeGlobalTab={this.state.activeGlobalTab}
             activePredictionTab={this.state.predictionTab}
+            showMessageBar={this.state.showMessageBar}
+            closeMessageBar={this.closeMessageBar.bind(this)}
```
xuke444, line 385: we should avoid using bind in render, it will create a new instance everytime render get called.
instead change the function defination to be arrow function could be a good practice.

xuke444:
is this only in error analysis or it also impact the mad as well?

imatiach-msft:
@xuke444 this is only in error analysis dashboard

libs/error-analysis/src/lib/ErrorAnalysisDashboard/ErrorAnalysisDashboard.tsx
imatiach-msft: done!

PR #1010, opened 2021-11-11
update readme links to be absolute instead of relative path and update raiwidgets to latest error analysis

imatiach-msft:
- update readme links to be absolute instead of relative path for the pypi page (https://pypi.org/project/raiwidgets/)
- update error analysis package dependency to latest
- update list of maintainers


xuke444 (approved):
seem error-analysis version also increased, probably better to include it in description as wel.

PR #1014, opened 2021-11-12
add error analysis error importances (aka correlation of features with error) to model analyzer API

imatiach-msft:
- add importances to the error analysis manager in model analyzer, so importances (aka correlation of features with error) will be computed

### DIFF
--- a/responsibleai/responsibleai/_interfaces.py
+++ b/responsibleai/responsibleai/_interfaces.py
@@ -69,6 +69,7 @@ class ErrorAnalysisData:
     tree_features: List[str]
     matrix_features: List[str]
     metric: str
+    importances: list
 
 
 class CausalMetric:
--- a/responsibleai/responsibleai/_managers/error_analysis_manager.py
+++ b/responsibleai/responsibleai/_managers/error_analysis_manager.py
@@ -262,7 +262,8 @@ class ErrorAnalysisManager(BaseManager):
             report = self._analyzer.create_error_report(
                 filter_features, max_depth=max_depth,
                 min_child_samples=min_child_samples,
-                num_leaves=num_leaves)
+                num_leaves=num_leaves,
+                compute_importances=True)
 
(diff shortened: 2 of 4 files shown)


gaugup:
Can you update the schema file for error analysis as a part of this PR?

responsibleai/responsibleai/_managers/error_analysis_manager.py
```diff
@@ -262,7 +262,8 @@ def compute(self):
             report = self._analyzer.create_error_report(
                 filter_features, max_depth=max_depth,
                 min_child_samples=min_child_samples,
-                num_leaves=num_leaves)
+                num_leaves=num_leaves,
+                compute_importances=True)
```
gaugup, line 266: Is compute_importance not a user defined parameter like num_leaves?

imatiach-msft:
"Can you update the schema file for error analysis as a part of this PR?"
done!

gaugup (approved):

responsibleai/responsibleai/_interfaces.py
```diff
@@ -69,6 +69,7 @@ class ErrorAnalysisData:
     tree_features: List[str]
     matrix_features: List[str]
     metric: str
+    importances: list
```
xuke444, line 72: do we have to use word "importance" it is sort of confusing.

imatiach-msft: That's a good point, this word is used in the UI directly so that's why it's also used in the backend.  We can change it but 1.) it probably requires some approval and discussion with PM, as well as decision on what new name should be 2.) it seems like it would require a lot of renames that are outside the scope of this PR.  I'll create a task to track this.

imatiach-msft: @xuke444 I created an issue to track this:
https://github.com/microsoft/responsible-ai-widgets/issues/1019

PR #1020, opened 2021-11-16
Bug fix: Feature importance for node with 0 error

tongyu-microsoft:
This PR fixes the bug when the tree node has 0 errors and N correct, the raw data table shows the opposite with N+1 errors and -1 correct. 
We should consider the case when the  firstIncorrectItemIndex is -1 (no incorrect item) separately.

### DIFF
--- a/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/IndividualFeatureImportanceView.tsx
+++ b/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/IndividualFeatureImportanceView.tsx
@@ -250,10 +250,13 @@ export class IndividualFeatureImportanceView extends React.Component<
           (row) =>
             row[JointDataset.TrueYLabel] !== row[JointDataset.PredictedYLabel]
         );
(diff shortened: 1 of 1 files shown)


libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/IndividualFeatureImportanceView.tsx
```diff
@@ -250,20 +250,24 @@ export class IndividualFeatureImportanceView extends React.Component<
...
             .CorrectPredictions,
           startIndex: 0
         },
         {
-          count:
-            this.props.selectedCohort.cohort.filteredData.length -
-            firstIncorrectItemIndex,
+          count: noIncorrectItem
+            ? 0
+            : this.props.selectedCohort.cohort.filteredData.length -
```
imatiach-msft, line 269: interesting that they are sorted by correct/incorrect

imatiach-msft (approved):
nice fix!  LGTM!

zhb000 (approved):

PR #1021, opened 2021-11-16
rename model analysis to rai insights

imatiach-msft:
Main change: Rename ModelAnalysis to RAIInsights, but keep the ModelAnalysis API for backwards compatibility.

Also, with the new API, I changed the "train_labels" parameter to be called "classes", since PM was confused about the name and this is more similar to scikit-learn (eg see classes_ attribute https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.LabelEncoder.html?highlight=labelencoder#sklearn.preprocessing.LabelEncoder) and InterpretML APIs.

The example notebook code will be updated once the dashboard is renamed as well to ResponsibleAIDashboard from ModelAnalysisDashboard in a follow-up PR.

For tests, I kept both the old tests and added the new tests for the new API.  The old tests can be removed once the API is removed.


RubyZ10 (approved):

vinuthakaranth:
may be a nit: Do we leave files with the name model_analysis_dashboard.py as is?

raiwidgets/raiwidgets/model_analysis_dashboard.py
```diff
@@ -5,7 +5,7 @@
 
 from flask import jsonify, request
 
-from responsibleai import ModelAnalysis
+from responsibleai import RAIInsights
 
 from .dashboard import Dashboard
 from .model_analysis_dashboard_input import ModelAnalysisDashboardInput
```
vinuthakaranth, line 11: Is this unrelated to this change?

vinuthakaranth (approved):

raiwidgets/raiwidgets/model_analysis_dashboard.py
imatiach-msft: yes that will be changed in follow-up PRs as well:
1.) change ModelAnalysisDashboard to ResponsibleAIDashboard
2.) change notebooks and any internal constants that are still left

PR #1022, opened 2021-11-16
Performing schema validations during compute() and _load()

gaugup:
It seems that time when schema validation varies for each manager. For instance in case of causal amanger, it is done at load() but for error analysis manager and counterfactual manager it was being done only at the compute time. Also, since load() operation may happen at any time so schema validation need to happen at load() time.

Hence, making schema validation uniform for all three managers by doing this check both at compute() and load() time.

### DIFF
--- a/responsibleai/responsibleai/_managers/causal_manager.py
+++ b/responsibleai/responsibleai/_managers/causal_manager.py
@@ -240,6 +240,8 @@ class CausalManager(BaseManager):
                 treatment_features[i], treatment_cost[i],
                 alpha, max_tree_depth, min_tree_leaf_samples)
             result.policies.append(policy)
+
+        result._validate_schema()
         self._results.append(result)
         return result
 
--- a/responsibleai/responsibleai/_managers/counterfactual_manager.py
+++ b/responsibleai/responsibleai/_managers/counterfactual_manager.py
@@ -336,7 +336,7 @@ class CounterfactualManager(BaseManager):
                                 permitted_range=cf_config.permitted_range)
 
                     # Validate the serialized output against schema
-                    schema = self._get_counterfactual_schema(
+                    schema = CounterfactualManager._get_counterfactual_schema(
                         version=counterfactual_obj.metadata['version'])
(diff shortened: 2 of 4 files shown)


responsibleai/responsibleai/_managers/causal_manager.py
```diff
@@ -240,6 +240,8 @@ def add(
                 treatment_features[i], treatment_cost[i],
                 alpha, max_tree_depth, min_tree_leaf_samples)
             result.policies.append(policy)
+
+        result._validate_schema()
```
imatiach-msft, line 244: It seems in causal it's done during add and load, while for counterfactual/error analysis it's now done for both compute and load?  It seems a bit inconsistent.  I also wonder if we need to move to a similar object to 'CausalResult' for error analysis and counterfactual?

imatiach-msft (approved):

responsibleai/responsibleai/_managers/causal_manager.py
gaugup: I don't know when/why was the implementation for causal changed from add/compute to just compute. I think Chris changed it for some reason. I don't think it is trivial to change to add()/compute() as azureml depends on this. 

>> I also wonder if we need to move to a similar object to 'CausalResult' for error analysis and counterfactual?
For causal, there was no object based interface that was provided by econml. Counterfactuals have a self-explanatiory interface like explanations so I don't think it makes a whole lot of sense to migrate Counterfactuals to that interface. For ErrorAnalysis, we can discuss this if we need to move to CausalResult based object.

riedgar-ms: We had been thinking about removing the `compute()` call, since it's a little pointless when you have live compute. Syncing the removal with the AzureML code would obviously be required, but it could be made a noop in the meantime.

gaugup: We need to decide on the name of the API that does both add and compute. Removing **compute()** is outside of the scope of this PR.

PR #1026, opened 2021-11-19
Directory based state management for model analysis tools

gaugup:
This PR has the following changes:-

- Move state of counterfactual, explainer, causal and erroranalysis manager to directory based state management as below.
- Add directory based state manager
- Add the unit tests for directory based state manager
- Add tests for checking the directory based storage paths for each manager state.

model_analysis_state/
|----explanations/
|----error_analysis/
|----causal/
|----counterfactual/
|--------guid1/
|------------config/
|----------------config.json
|------------data/
|----------------f1.json
|----------------f2.json
|----------------f3.json
|------------explainer/
|----------------model1.pkl
|----------------model2.pkl
|--------guid2/
|------------config/
|----------------config.json
|------------data/
|----------------f1.json
|----------------f2.json
|----------------schema.json
|------------explainer/
|----------------model1.pkl
|----------------conda_env.yml
|--------guid3/
|----model.pkl
|----train.json
|----test.json

### DIFF
--- a/responsibleai/responsibleai/_tools/shared/state_directory_management.py
+++ b/responsibleai/responsibleai/_tools/shared/state_directory_management.py
@@ -1,0 +1,124 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+"""Directory management of state of ResponsibleAI insights."""
+
+import os
+import uuid
+from pathlib import Path
+
+
+class DirectoryManager:
+
+    CONFIGURATION = 'configuration'
+    DATA = 'data'
+    EXPLAINER = 'explainer'
+
+    def __init__(self, parent_directory_path, sub_directory_name=None):
+        """Directory manager for managing the configuration,
+           dashboard data and explainers need to recompose the state
+           in the various managers.
+
+        :param parent_directory_path: Directory path of where the state data
+                                      directory will be stored.
+        :type parent_directory_path: Path
+        :param sub_directory_name: The name where the state needs to be stored
+                                   or the directory where the state can be
+                                   found.
+        :type sub_directory_name: str
+        """
+        self.parent_directory_path = parent_directory_path
+
+        if sub_directory_name is None:
+            self.sub_directory_name = str(uuid.uuid4())
+        else:
+            self.sub_directory_name = sub_directory_name
+
+        parent_directory = Path(self.parent_directory_path)
+        if not parent_directory.exists():
+            parent_directory.mkdir(parents=True, exist_ok=True)
+
+        sub_directory = (Path(self.parent_directory_path) /
+                         self.sub_directory_name)
+        if not sub_directory.exists():
+            sub_directory.mkdir(parents=True, exist_ok=True)
+
+    def create_config_directory(self):
+        """Create the configuration directory and return its path.
+
+        :returns: Path to the configuration directory.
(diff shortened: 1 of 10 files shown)


riedgar-ms (approved):
Thanks!

responsibleai/responsibleai/_tools/shared/state_directory_management.py
```diff
@@ -0,0 +1,124 @@
...
+        data_directory = (Path(self.parent_directory_path) /
+                          self.sub_directory_name /
+                          DirectoryManager.DATA)
+        if not data_directory.exists():
+            data_directory.mkdir(parents=True, exist_ok=True)
+
+        return data_directory
+
+    def create_explainer_directory(self):
+        """Create the explainer directory and return its path.
```
imatiach-msft, line 72: this seems something too specific to explanations to be in this class?  Just seems a little weird to put it in a more common RAI class.  I wonder if there is a better way to structure the DirectoryManager.

responsibleai/tests/test_model_analysis.py
```diff
@@ -252,6 +261,36 @@ def run_model_analysis(model, train_data, test_data, target_column,
...
+        dm = DirectoryManager(path / manager_type, component_path)
+
+        config_path = dm.get_config_directory()
+        data_path = dm.get_data_directory()
+        explainer_path = dm.get_explainer_directory()
+
+        if manager_type == ManagerNames.EXPLAINER:
+            assert not config_path.exists()
+            assert data_path.exists()
+            assert not explainer_path.exists()
```
imatiach-msft, line 279: It seems none of them are using the explainer_path right now?  Perhaps at this point it shouldn't exist then?

responsibleai/responsibleai/_tools/shared/state_directory_management.py
imatiach-msft: Will this also be used for causal model in the future?  Maybe the name could just be made more generic somehow.

imatiach-msft: It looks like it's not used currently by the managers if I understand the code correctly right now.

imatiach-msft (approved):

responsibleai/responsibleai/_tools/shared/state_directory_management.py
imatiach-msft: How about this: we have different directory managers for different components, and the explanation and counterfactual components can use the ExplainerDirectoryManager which includes this explainer directory.

responsibleai/tests/test_model_analysis.py
gaugup: I plan to move causal to explainer path. The causal is state management is a little bit tricky. So keeping it in data directory for now. Will address that in up coming PR.

responsibleai/responsibleai/_tools/shared/state_directory_management.py
gaugup: As per the design dicussion we decided on maintaining the same directory structure for all the components. Do you have any name suggestion for explainer_path? 

For causal manager, I will move the causal models to explainer path in a follow up PR. The state management for causal is alittle bit tricky.

PR #1030, opened 2021-11-19
Add .json extension to train and test dataset

gaugup:


responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -503,12 +503,12 @@ def save(self, path):
         dtypes = self.train.dtypes.astype(str).to_dict()
         self._write_to_file(data_directory / (_TRAIN + _DTYPES),
                             json.dumps(dtypes))
-        self._write_to_file(data_directory / _TRAIN,
+        self._write_to_file(data_directory / (_TRAIN + '.json'),
```
imatiach-msft, line 506: maybe move the .json to a constant at top of file like JSON = '.json' and just re-use here?

imatiach-msft (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
gaugup: Added the string constant.

PR #1031, opened 2021-11-19
when classes are not passed, sort them automatically

imatiach-msft:
When classes are not passed to RAIInsights, sort them automatically.
Currently, if the user has classes like "0", "1", they may get arranged in any order after calling unique(), hence it is common for classes to be the array [1,0].  This can give results in ResponsibleAIDashboard which look difficult to interpret, since the probabilities for negative class now correspond to positive 1.0 probability.

### DIFF
--- a/responsibleai/tests/test_rai_insights_validations.py
+++ b/responsibleai/tests/test_rai_insights_validations.py
@@ -6,6 +6,7 @@ from unittest.mock import MagicMock
 
 import pandas as pd
 import pytest
+import numpy as np
 
 from responsibleai.exceptions import UserConfigValidationException
 from responsibleai import RAIInsights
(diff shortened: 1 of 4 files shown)


responsibleai/tests/test_rai_insights_validations.py
```diff
@@ -311,6 +312,16 @@ def test_classes(self):
 
         assert 'The train labels and distinct values in target ' + \
             '(test data) do not match' in str(ucve.value)
+        
+        rai = RAIInsights(
```
gaugup, line 316: Can you create a seprate test case for this validation?

gaugup (changes requested):

responsibleai/tests/test_rai_insights_validations.py
imatiach-msft: done!

gaugup (approved):

xuke444 (approved):

PR #1034, opened 2021-11-22
Use spilt orientation to save train and test dataframe to address duplicate indices condition in them

gaugup:
The to_json() call expects non-duplicate indexes in the pandas Dataframe. If saving the Dataframe to JSON with duplicate indexes you may get the following exception:-

>>> new_df.to_json()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\gaugup\Anaconda3\envs\rai\lib\site-packages\pandas\core\generic.py", line 2424, in to_json
    index=index,
  File "C:\Users\gaugup\Anaconda3\envs\rai\lib\site-packages\pandas\io\json\_json.py", line 78, in to_json
    index=index,
  File "C:\Users\gaugup\Anaconda3\envs\rai\lib\site-packages\pandas\io\json\_json.py", line 122, in __init__
    self._format_axes()
  File "C:\Users\gaugup\Anaconda3\envs\rai\lib\site-packages\pandas\io\json\_json.py", line 202, in _format_axes
    "'{orient}'.".format(orient=self.orient)
ValueError: DataFrame index must be unique for orient='columns'.

Solution in the PR is to set orient='split' in to_json() and read_json() so that the index is saved and read as a separate dictionary.


responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -504,13 +521,16 @@ def save(self, path):
...
                             json.dumps(dtypes))
+        reset_index_train = self._reset_data_frame_index(self.train)
         self._write_to_file(data_directory / (_TRAIN + _JSON_EXTENSION),
-                            self.train.to_json())
+                            reset_index_train.to_json())
+
         dtypes = self.test.dtypes.astype(str).to_dict()
         self._write_to_file(data_directory / (_TEST + _DTYPES),
                             json.dumps(dtypes))
+        reset_index_test = self._reset_data_frame_index(self.test)
```
imatiach-msft, line 531: don't we then need to set the index back when calling load?  I don't quite understand where we undo the changes to reset index?

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -489,6 +489,23 @@ def _write_to_file(self, file_path, content):
...
+        :return: Dataframe without duplicate indices.
+        :rtype: pd.DataFrame
+        """
+        num_indices = len(some_dataframe.index.tolist())
+        num_distinct_indices = len(np.unique(some_dataframe.index).tolist())
+
+        if num_indices == num_distinct_indices:
+            return some_dataframe
+
+        return some_dataframe.reset_index(drop=True)
```
imatiach-msft, line 507: I don't think we can drop the index like this:
```drop=True```
I recall there were many cases where this is needed by the learner itself to make predictions, especially in time series scenarios?

Comment (gaugup) [reply responsibleai/responsibleai/rai_insights/rai_insights.py:507]: The reset_index call doesn't drop the index but resets it to be uniue. Documentation is here https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.reset_index.html.

The timeseries models should have unique indexes right?

File: responsibleai/responsibleai/rai_insights/rai_insights.py (hunk 1 above)
Comment (gaugup) [reply responsibleai/responsibleai/rai_insights/rai_insights.py:531]: As responded in the previous comment, we are not dropping the index but resetting it.

File: responsibleai/responsibleai/rai_insights/rai_insights.py (hunk 2 above)
Comment (imatiach-msft) [reply responsibleai/responsibleai/rai_insights/rai_insights.py:515]: discussed offline - when drop=True, the index is reset to default monotonically increasing values, which is effectively dropping that information.  However, there are some weird models/scenarios we've encountered in past where this index is actually used.  There may also be tricky cases with MultiIndex scenarios.  If possible, the ideal solution would be to make the index a regular column through reset_index without drop=True, and then call set_index on deserialize to rehydrate the index column and recreate the exact same dataset as the original one.

Comment (gaugup) [reply responsibleai/responsibleai/rai_insights/rai_insights.py]: the option orient='split' in to_json() saves index as a separate entity in the json file so even if duplicate index exists in the dataframe.

**Review by imatiach-msft — APPROVED** [review]

File: responsibleai/tests/test_model_analysis_validations.py [hunk 3]
```diff
@@ -52,7 +52,7 @@ def test_validate_test_data_size(self):
                 target_column='bad_target',
                 task_type='classification',
                 maximum_rows_for_test=len(y_test) - 1)
-        assert "The test data has 30 rows, but limit is set to 29 rows" in \
+        assert "The test data has 31 rows, but limit is set to 30 rows" in \
```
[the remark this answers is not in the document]
gaugup: This is because added another row to X_test in create_iris_data() so the values changed.

PR #1035, opened 2021-11-22
Raise user config exception when train/test data is not pandas dataframe

gaugup:
It looks like downstream code in RAI Insights doesn't support anything apart from pandas DataFrame. Hence, raising the user exception to prevent unexpected errors.


responsibleai/tests/test_model_analysis_validations.py
```diff
@@ -263,6 +263,24 @@ def test_mismatch_train_test_features(self):
...
+    def test_unsupported_train_test_types(self):
+        X_train, X_test, y_train, y_test, _, _ = \
+            create_cancer_data()
+        model = create_lightgbm_classifier(X_train, y_train)
+
+        X_train['target'] = y_train
+        X_test['bad_target'] = y_test
+
+        with pytest.raises(UserConfigValidationException) as ucve:
+            ModelAnalysis(
```
imatiach-msft, line 275: I don't think we need to keep adding tests for new functionality/code cases to the old model analysis API, we can just add new tests to RAIInsights, what do you think?  The other alternative is to change the tests to use pytest.mark.parametrize or run both classes in a loop and have a single test file but I feel it can be error prone, and will make it difficult to add new functionality to the new class while keeping the same functionality on the old class.

imatiach-msft (approved):

responsibleai/tests/test_model_analysis_validations.py
gaugup: I am also finding it increasingly annopying having to add the same test in two files. Since ModelAnalysis is just a wrapper over RAIInsights and new functionaility goes in RAIInsights, one set of tests should help remedy this. I don't understand why would that be a problem.

riedgar-ms: Have we marked ModelAnalysis as deprecated? If so, I think it makes sense not to bother adding tests for it.

imatiach-msft: Exactly, it now prints a warning and says it is deprecated.  We've done deprecations this way in the past and usually we just stop modifying the test files on the deprecated classes moving forward but keep them around just to validate they can still run.  Eventually when we remove the class we remove the tests as well.

gaugup: Sure sounds good. I will let this merge but going forward will only add tests for the new class only.

PR #1037, opened 2021-11-23
Add separate github workflow for python linting

gaugup:
Creating one github workflow for all python linting. Currently, same flake8 linter runs in all workflows which is needless.

### DIFF
--- a/.github/workflows/CI-python.yml
+++ b/.github/workflows/CI-python.yml
@@ -84,9 +84,6 @@ jobs:
           pip install -v .
         working-directory: ${{ matrix.packageDirectory }}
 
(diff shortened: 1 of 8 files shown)


imatiach-msft (approved):
nice fix!

gaugup:
> Two thoughts for the future:
> 
> * Add in `mypy` as well
> * See if the trigger can be made specific to Python files. No point running this if only the TS changed

Yes, mypy can be added once we add type hints in all python filles. 

I think flake8 is good linter for any python code. Keeps code clean and very readable.

xuke444 (approved):

PR #1038, opened 2021-11-24
[E2E] What-if Counterfactuals - common functionalities & Y Axis Flyout 

vinuthakaranth:


apps/widget-e2e/src/describer/modelAssessment/whatIfCounterfactuals/describeSubBarChart.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { BarChart } from "../../../util/BarChart";
+import { ScatterChart } from "../../../util/ScatterChart";
+import { IModelAssessmentData } from "../IModelAssessmentData";
+
+export function describeSubBarChart(dataShape: IModelAssessmentData): void {
+  const props = {
+    chart: undefined as unknown as ScatterChart,
```
tongyu-microsoft, line 10: how is the different from `ScatterChart | undefined`?

tongyu-microsoft (approved):

apps/widget-e2e/src/describer/modelAssessment/whatIfCounterfactuals/describeSubBarChart.ts
vinuthakaranth: I just checked locally, these doesn't work in same way. Here undefined is assigned as the value for chart rather than Type.

PR #1040, opened 2021-11-25
Split cf output into individual json files

gaugup:
Trying to create the same set of json files for CF output as what the UI expects. 

Ouput below:-

image

### DIFF
--- a/responsibleai/responsibleai/_managers/counterfactual_manager.py
+++ b/responsibleai/responsibleai/_managers/counterfactual_manager.py
@@ -46,6 +46,9 @@ class CounterfactualConfig(BaseConfig):
     HAS_COMPUTATION_FAILED = 'has_computation_failed'
     FAILURE_REASON = 'failure_reason'
 
+    CONFIG_FILE_NAME = 'config.json'
+    RESULT_FILE_NAME = 'result.json'
+
     def __init__(self, method, continuous_features, total_CFs,
                  desired_class=CounterfactualConstants.OPPOSITE,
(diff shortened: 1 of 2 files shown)


xuke444 (approved):

responsibleai/responsibleai/_managers/counterfactual_manager.py
```diff
@@ -119,6 +228,90 @@ def get_result(self):
...
+                    counterfactual_examples_dict[
+                        counterfactual_examples_key] = json.load(result_file)
+
+            counterfactuals_json_str = json.dumps(counterfactual_examples_dict)
+            self.counterfactual_obj = \
+                CounterfactualExplanations.from_json(counterfactuals_json_str)
+        else:
+            self.counterfactual_obj = None
+
+        result_path = (data_directory_path /
```
imatiach-msft, line 300: I think this kind of logic needs to live in a common place instead of being replicated in each manager?
HAS_COMPUTATION_FAILED
FAILURE_REASON
IS_COMPUTED
these are all common to all of the managers and seems strange to have this logic just here?

imatiach-msft (approved):

responsibleai/responsibleai/_managers/counterfactual_manager.py
gaugup: Yes, I agree that we should consolidate the logic as per common code. But the major trouble is that causal is using a completely different logic to manage the config. So we might need to refactor the causal manager first before addressing the consolidation of serilization logic.

PR #1041, opened 2021-11-25
Change empty list to None for categorical_features in RAI tests

gaugup:
It seems we don't need to pass empty lists for categorical_features. The underlying code handles None.


responsibleai/tests/test_model_analysis.py
```diff
@@ -61,7 +61,7 @@ def test_model_analysis_iris(self, manager_type):
         }
 
         for model in models:
-            run_model_analysis(model, X_train, X_test, LABELS, [],
+            run_model_analysis(model, X_train, X_test, LABELS, None,
```
riedgar-ms, line 64: Should we test both ways, though (at least once, if not on all of these tests)?

imatiach-msft (approved):

responsibleai/tests/test_model_analysis.py
gaugup: Is there a use case when user might pass an empty list for categorical_features?

PR #1042, opened 2021-11-29
Rename the explainer directory to generators

gaugup:
@imatiach-msft pointed out that the directory name "explainer" for storing all the generators is more specific to explanations. So changing this to be "models" for more generic terminology.


xuke444 (approved):

gaugup:
> If these are for what we're calling generators, why not call the directory that? When I see 'model' I think of the ML model we're analysing.

Changed to generators

PR #1050, opened 2021-12-01
counterfactual list style 

zhb000:
highlight counterfactual examples columns  with different color, and bold size


imatiach-msft (approved):

libs/counterfactuals/src/lib/CounterfactualList.tsx
```diff
@@ -285,6 +286,16 @@ export class CounterfactualList extends React.Component<
     });
   };
 
+  private renderItemColumn(item: any, index?: number, column?: IColumn) {
```
xuke444, line 289: use arrow function here can avoid bind above.

xuke444 (approved):

PR #1051, opened 2021-12-01
fix mutating predict function scenario for model analysis

imatiach-msft:
A customer had an issue where RAIInsights was failing, and after investigating the issue it was discovered that the predict function was mutating the input dataset and adding back the label column.  This PR adds a warning to auto-discover such mutations on behalf of the user but still intelligently allows them to occur and run RAIInsights successfully regardless.

### DIFF
--- a/responsibleai/responsibleai/rai_insights/rai_insights.py
+++ b/responsibleai/responsibleai/rai_insights/rai_insights.py
@@ -409,6 +409,8 @@ class RAIInsights(object):
         dataset: pd.DataFrame = self.test.drop(
             [self.target_column], axis=1)
 
+        features = dataset.columns
+
         if isinstance(dataset, pd.DataFrame) and hasattr(dataset, 'columns'):
             self._dataframeColumns = dataset.columns
         try:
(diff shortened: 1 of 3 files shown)


responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -423,6 +425,14 @@ def _get_dataset(self):
                 msg = "Model does not support predict method for given"
                 "dataset type"
                 raise ValueError(msg) from ex
+            post_pred_features = dataset.columns
```
gaugup, line 428: We should not address error scenarios here but raise an error in the input validations for rai_insights. It really complicates our package design. 

The customer should wrap there models to adhere to our package.

imatiach-msft: done, created a separate PR: https://github.com/microsoft/responsible-ai-toolbox/pull/1055

imatiach-msft:
closing this PR in favor of:
https://github.com/microsoft/responsible-ai-toolbox/pull/1055
which throws an error instead

PR #1055, opened 2021-12-01
throw error on mutating predict function in RAIInsights

imatiach-msft:
A customer had an issue where RAIInsights was failing, and after investigating the issue it was discovered that the predict function was mutating the input dataset and adding back the label column. This PR throws an error if it discovers such mutations on behalf of the user.

### DIFF
--- a/responsibleai/responsibleai/rai_insights/rai_insights.py
+++ b/responsibleai/responsibleai/rai_insights/rai_insights.py
@@ -291,6 +291,8 @@ class RAIInsights(object):
                 small_test_data = test.iloc[0:1].drop(
                     [target_column], axis=1)
 
+                small_train_features_before = list(small_train_data.columns)
+
                 # Run predict() of the model
                 try:
                     model.predict(small_train_data)
@@ -300,6 +302,9 @@ class RAIInsights(object):
                         'The model passed cannot be used for'
                         ' getting predictions via predict()'
                     )
+                self._validate_features_same(small_train_features_before,
+                                             small_train_data,
+                                             SKLearn.PREDICT)
 
                 # Run predict_proba() of the model
                 if task_type == ModelTask.CLASSIFICATION:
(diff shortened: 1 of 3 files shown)


responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -311,6 +313,14 @@ def _validate_model_analysis_input_parameters(
                             'The model passed cannot be used for'
                             ' getting predictions via predict_proba()'
                         )
+                small_train_features_after = list(small_train_data.columns)
+
+                if small_train_features_before != small_train_features_after:
```
gaugup, line 318: should we also have this check after predict_proba() test as well (just below). It could be that predict() doesn't change the input data but predict_proba() changes the input data.

imatiach-msft: oh the check is after both the predict and predict_proba

imatiach-msft: I suppose I could add a separate check for predict_proba so as to have a custom error message though

imatiach-msft: done, added separate checks both after predict and predict_proba methods, and added another test to validate each case

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -324,6 +332,16 @@ def _validate_model_analysis_input_parameters(
                 "Expecting pandas Dataframe for train and test."
             )
 
+    def _validate_features_same(self, small_train_features_before,
+                                small_train_data, function):
```
gaugup, line 336: 1. Can this function take the first two parameter as list of features? Looks a weird function signature with one being list and other being dataframe.

2. Could you add a function description and parameters descriptions and types?

gaugup (changes requested):

responsibleai/responsibleai/rai_insights/rai_insights.py
imatiach-msft: 1.) I could, but I would have the same duplicate code twice though, so I think this is preferred as it reduces code duplication?
2.) done!  This is an internal function though and I see other internal functions which don't have docs, but probably good to add it anyway.

gaugup (approved):

PR #1059, opened 2021-12-02
Fix SubBarChart flaky test

vinuthakaranth:


zhb000 (approved):

apps/widget-e2e/src/util/Table.ts
```diff
@@ -29,3 +29,23 @@ export function getTableRowCount(srPrefix?: string): number {
...
+  columnValue: string,
+  srPrefix?: string
+): void {
+  const selectorPrefix = srPrefix ? `${srPrefix}` : "";
+  const columnsHeaderColumn = `${selectorPrefix} [data-automationid='ColumnsHeaderColumn']`;
+  cy.get(columnsHeaderColumn)
+    .contains(columnName)
+    .parents(columnsHeaderColumn)
+    .invoke("attr", "data-item-key")
+    .then((columnIndex) => {
```
xuke444, line 44: This way may be an issue.
The async statement in side "then" may not triggered before test finish.

Probably better to convert to async + await

imatiach-msft (approved):

apps/widget-e2e/src/util/Table.ts
vinuthakaranth: The statements inside will be async, but Cypress will manage async and await by default.

PR #1060, opened 2021-12-03
Refactor load() method in RAIInsights

gaugup:
Breaking up operations in load() into separate functions.


responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -649,4 +676,24 @@ def load(path):
...
+        :type path: str
+        :return: The RAIInsights object after loading.
+        :rtype: RAIInsights
+        """
+        # create the RAIInsights without any properties using the __new__
+        # function, similar to pickle
+        inst = RAIInsights.__new__(RAIInsights)
+
+        # load current state
+        RAIInsights._load_data(inst, path)
```
imatiach-msft, line 694: maybe at some point we should even move these load methods to a separate file or file(s), since this rai_insights file is already very large

imatiach-msft (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
gaugup: Sure. We might need to think about the interfaces for these methods to be exported to another file.

PR #1061, opened 2021-12-03
default categorical feature to array

xuke444:
Signed-off-by: Ke xu <xuke@microsoft.com>


gaugup:
I remember we had a discussion around keeping it as None if not specified. Why is this being changed back?

imatiach-msft:
+1 confused why this is being changed back

xuke444:
it breaks ui and there will be a lot places to change on ui instead of fix in sdk.

imatiach-msft (approved):

imatiach-msft:
@xuke444 I'm actually seeing this issue now with the notebook I'm adding in this PR:
https://github.com/microsoft/responsible-ai-toolbox/pull/1105
the UX just seems to break because no categorical features are specified, which is really bad.  I think this should be merged ASAP, otherwise non-categorical scenarios are broken.

PR #1073, opened 2021-12-08
[E2E] Dataset explorer - whisker plot

vinuthakaranth:


apps/widget-e2e/src/describer/modelAssessment/dataExplorer/describeAxisConfigDialog.ts
```diff
@@ -0,0 +1,135 @@
...
+              cy.get("#AxisConfigPanel")
+                .find("button")
+                .contains("Select")
+                .click();
+              cy.get(
+                '#DatasetExplorerChart div[class*="rotatedVerticalBox"] button:eq(0)'
+              ).contains(text1);
+            });
+          cy.get(
+            '#DatasetExplorerChart div[class*="rotatedVerticalBox"] button'
```
imatiach-msft, line 52: I see some of these strings repeated, like:
```
#DatasetExplorerChart div[class*="rotatedVerticalBox"]
```

do you think it might be better to move the repeating strings to a constants file, to make it easier to maintain/update tests in the future?

imatiach-msft: or even just at top of the file and not a separate constants file if they only repeat within the test

imatiach-msft (approved):

apps/widget-e2e/src/describer/modelAssessment/dataExplorer/describeAxisConfigDialog.ts
vinuthakaranth: Thanks for the comment, moved all repeating strings to constant file.

Issue #1072, opened 2021-12-08
Scroll is hidden in Heat Map


Diyarmo:
Hi there,

When I use the heat map only the first 13 rows of the "Feature 1" are shown and the rest of the rows are hidden.
![image](https://user-images.githubusercontent.com/43572673/145205628-49c91d1f-9249-4fa6-a741-4695edb1bcac.png)

in the above image 7 rows are hidden and normally there is no way I can see them.

I tried some changes in the CSS of the webpage and noticed that the overflow attribute of cohortsLayerHost2 div is set to "hidden".
![image](https://user-images.githubusercontent.com/43572673/145206296-4b27db00-9821-4c52-858a-42a234493a2d.png)

So, I changed the overflow to "auto" and the problem was fixed.

I did some search in the repository but I couldn't find the right file to change. can you please fix this bug?

imatiach-msft:
Great point. Maybe we can add a horizontal and vertical scroll bar.  I've heard this issue mentioned before by internal users actually, but I think there was some hesitancy in adding scroll bars to the UX and discussion around solving this through some sort of aggregation.  But the issue is probable most easily solved by adding the scroll bars.

imatiach-msft:
closing as fixed with merged pr #1074

PR #1074, opened 2021-12-08
add scrollbar area around matrix cells when there are many categories or bins

imatiach-msft:
Resolves issue https://github.com/microsoft/responsible-ai-toolbox/issues/1072
With high number of bins or categorical features users will see a scrollbar now instead of the heatmap cells being cut off.

High number of bins:
image

Categorical features with many categories (scrolled to bottom):
image

Note that user will not see scrollbar if matrix area is small.

### DIFF
--- a/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Matrix/MatrixArea/MatrixArea.styles.ts
+++ b/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Matrix/MatrixArea/MatrixArea.styles.ts
@@ -2,13 +2,14 @@
 // Licensed under the MIT License.
 
 import {
+  IProcessedStyleSet,
   IStyle,
-  mergeStyleSets,
-  IProcessedStyleSet
+  mergeStyleSets
 } from "office-ui-fabric-react";
 
 export interface IMatrixAreaStyles {
   emptyLabelPadding: IStyle;
+  layerHost: IStyle;
   matrixLabelBottom: IStyle;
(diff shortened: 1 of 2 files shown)


libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Matrix/MatrixArea/MatrixArea.styles.ts
```diff
@@ -21,6 +22,11 @@ export const matrixAreaStyles: () => IProcessedStyleSet<IMatrixAreaStyles> =
       emptyLabelPadding: {
         paddingTop: "60px"
       },
+      layerHost: {
+        height: "500px",
```
vinuthakaranth, line 26: Since height is fixed, does this work in zoom-in conditions?

vinuthakaranth (approved):

tongyu-microsoft (approved):

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Matrix/MatrixArea/MatrixArea.styles.ts
imatiach-msft: Great point, but we specify the height in pixels in many other places as well, so if it's a problem we need a more general fix probably.  However, so far I haven't heard that this is a problem in either the widget or full-screen view.  Perhaps we may need to change this eventually.  I'm not quite sure what the ideal fix is - do we want to provide percentages everywhere instead?

PR #1075, opened 2021-12-08
Use getattr() to reference predict_proba()

gaugup:


vinuthakaranth (approved):

imatiach-msft (approved):

raiwidgets/raiwidgets/error_analysis_dashboard_input.py
```diff
@@ -136,7 +136,7 @@ class and for the regression case a method of predict()
         self._categorical_indexes = []
         self._is_classifier = model is not None\
             and hasattr(model, SKLearn.PREDICT_PROBA) and \
-            model.predict_proba is not None
+            getattr(model, SKLearn.PREDICT_PROBA) is not None
```
imatiach-msft, line 139: why do we need to call both hasattr and getattr?  I think checking just hasattr should be enough (or just getattr)?

https://stackoverflow.com/questions/24971061/python-hasattr-vs-getattr

hasattr just calls getattr in the implementation

gaugup: Yes, changed the logic to previous one. getattr() has some flake8 related issues. Refatored the code to make these big if conditions into a function call.

PR #1078, opened 2021-12-09
add flake8-bugbear for better python code contructs

gaugup:


imatiach-msft (approved):
Since the build is green, and I don't see any other changes, does this mean our code is already perfect?

imatiach-msft:
I guess it just writes warnings but doesn't trigger failures?

gaugup:
> Since the build is green, and I don't see any other changes, does this mean our code is already perfect?

I think there are no error pertaining to flake8-bugbear as yet in our code. But I found it's errors reporting on code really useful.

gaugup:
> I guess it just writes warnings but doesn't trigger failures?

No errors in our code as yet. But any future additions that has such issues dhould trigger a failure.

PR #1083, opened 2021-12-11
Gracefully handle failure in loading model.pkl from disk

gaugup:
There could be a situation where pickle.load() cannot load the model. In such scenarios, the load() of the RAIInsights will fail. Adding additional logic to cover this scenario.

### DIFF
--- a/responsibleai/responsibleai/rai_insights/rai_insights.py
+++ b/responsibleai/responsibleai/rai_insights/rai_insights.py
@@ -714,8 +714,11 @@ class RAIInsights(object):
             inst.__dict__[_MODEL] = serializer.load(top_dir)
         else:
             inst.__dict__['_' + _SERIALIZER] = None
-            with open(top_dir / _MODEL_PKL, 'rb') as file:
-                inst.__dict__[_MODEL] = pickle.load(file)
+            try:
+                with open(top_dir / _MODEL_PKL, 'rb') as file:
+                    inst.__dict__[_MODEL] = pickle.load(file)
+            except Exception:
+                inst.__dict__[_MODEL] = None
 
     @staticmethod
     def _load_managers(inst, path):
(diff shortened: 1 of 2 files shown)


responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -714,8 +714,11 @@ def _load_model(inst, path):
             inst.__dict__[_MODEL] = serializer.load(top_dir)
         else:
             inst.__dict__['_' + _SERIALIZER] = None
-            with open(top_dir / _MODEL_PKL, 'rb') as file:
-                inst.__dict__[_MODEL] = pickle.load(file)
+            try:
+                with open(top_dir / _MODEL_PKL, 'rb') as file:
+                    inst.__dict__[_MODEL] = pickle.load(file)
+            except Exception:
+                inst.__dict__[_MODEL] = None
```
riedgar-ms, line 721: Perhaps log a warning?

imatiach-msft (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -714,8 +714,11 @@ def _load_model(inst, path):
             inst.__dict__[_MODEL] = serializer.load(top_dir)
         else:
             inst.__dict__['_' + _SERIALIZER] = None
-            with open(top_dir / _MODEL_PKL, 'rb') as file:
-                inst.__dict__[_MODEL] = pickle.load(file)
+            try:
```
imatiach-msft, line 717: not necessary but might be interesting to add a test for this scenario

responsibleai/responsibleai/rai_insights/rai_insights.py
gaugup: Added the requested warning.

responsibleai/responsibleai/rai_insights/rai_insights.py
gaugup: Added the test case.

PR #1085, opened 2021-12-14
Fix issue with missing directories

riedgar-ms:
When saving to AzureML, directories are only implicit. This means that empty directories don't exist, which is a problem when we download an `RAIInsights` directory where one or more of the tools wasn't run.

This change tweaks the causal, counterfactual and error analysis managers to be tolerant of this. It is not applied to the explainer manager because that does create files in its manager directory even if no explanation is produced.

Signed-off-by: Richard Edgar <riedgar@microsoft.com>


responsibleai/responsibleai/_managers/counterfactual_manager.py
```diff
@@ -675,6 +675,7 @@ def _load(path, rai_insights):
 
         inst.__dict__[CounterfactualManager._COUNTERFACTUAL_CONFIG_LIST] = []
 
+        DirectoryManager.ensure_dir_exists(path)
         all_cf_dirs = DirectoryManager.list_sub_directories(path)
```
gaugup, line 679: Ideally the call to list_sub_directories() should return an empty list. Why do you need to create a directory?

riedgar-ms: This seemed the lowest-risk thing to do - normally there would be a directory there. Also suppose a user then adds an insight via the manager - will that automatically create the directory if it isn't there?

riedgar-ms: Although, a new save would presumably be a whole new directory tree. Let me see if adjusting the directory list to be empty works.

riedgar-ms: It works, and on reflection is probably a better solution. Have adjusted the PR.

vinuthakaranth (approved):

riedgar-ms:
@gaugup are you OK with this now?

gaugup (approved):

PR #1082, opened 2021-12-10
Test for Save/load/save bug in Explanations

riedgar-ms:
There was a bug in the way `interpret-community` loaded explanations which prevented them being saved again. This test ensures that this round trip remains working (the fix was made to `interpret-community` itself).

Closes #1081

PR #1096, opened 2021-12-27
Separate out save() and load() tests

gaugup:


riedgar-ms (approved):
What about the case covered by #1082 which you fixed in `interpret-community`? Or have we not done the release of `interpret-community` yet to pick that up?

imatiach-msft (approved):

gaugup:
> What about the case covered by #1082 which you fixed in `interpret-community`? Or have we not done the release of `interpret-community` yet to pick that up?

Yes, the fix is checked in interpret-community but there is no release as yet.

Issue #1091, opened 2021-12-21
[ResponsibleAIDashboard]: Heatmap cannot get generated for int64 features


nushib:
Notebook: http://localhost:8888/notebooks/responsibleaidashboard-housing-classification-model-debugging.ipynb

Generate heatmap using "OverallCond" as feature1

Error: 
ERROR in app: Exception on /matrix [POST]
...
TypeError: Object of type 'int64' is not JSON serializable

imatiach-msft:
closing as fixed with PR https://github.com/microsoft/responsible-ai-toolbox/pull/1097

PR #1097, opened 2021-12-27
fix handling numpy types in heatmap for categories

imatiach-msft:
Resolves issue:
https://github.com/microsoft/responsible-ai-toolbox/issues/1091
Error:
TypeError: Object of type 'int64' is not JSON serializable
Fixes the error by ensuring categories are converted to python types if they are the corresponding numpy type in the matrix json.

### DIFF
--- a/erroranalysis/erroranalysis/_internal/matrix_filter.py
+++ b/erroranalysis/erroranalysis/_internal/matrix_filter.py
@@ -389,7 +389,7 @@ def matrix_2d(categories1, categories2, matrix_counts,
             category1_max_interval.append(cat1.right)
             category1.append(str(cat1))
         else:
-            category1.append(cat1)
+            category1.append(get_py_value(cat1))
         for col_index in range(len(categories2)):
             cat2 = categories2[col_index]
             index_exists_err = cat1 in matrix_err_counts.index
@@ -454,7 +454,7 @@ def matrix_2d(categories1, categories2, matrix_counts,
             category2_max_interval.append(cat2.right)
             category2.append(str(cat2))
         else:
-            category2.append(cat2)
+            category2.append(get_py_value(cat2))
     category1 = {VALUES: category1,
                  INTERVAL_MIN: category1_min_interval,
                  INTERVAL_MAX: category1_max_interval}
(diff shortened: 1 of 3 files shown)


erroranalysis/erroranalysis/_internal/matrix_filter.py
```diff
@@ -540,3 +540,9 @@ def fill_matrix_nulls(matrix, null_value):
     idx_tuples = list(zip(idx_arrays[0], idx_arrays[1]))
     for tuple in idx_tuples:
         matrix.iloc[tuple] = null_value
+
+
+def get_py_value(value):
+    if isinstance(value, np.generic):
```
romanlutz, line 546: Could it be anything other than `np.generic`? Just wondering since numpy has many types. I'm sure you've considered that (hence more of a question), but the reason why it's only `np.generic` isn't obvious to me from this PR.

romanlutz (approved):

erroranalysis/erroranalysis/_internal/matrix_filter.py
imatiach-msft: this was the best I could come up with based on these searches:

https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.generic

https://stackoverflow.com/questions/12569452/how-to-identify-numpy-types-in-python#:~:text=To%20get%20the%20type%2C%20use%20the%20builtin%20type,%27numpy%27%20in%20str%20%28type%20%28a%29%29%20Out%20%3A%20True

https://stackoverflow.com/questions/25432523/how-to-check-if-a-value-is-of-a-numpy-type

riedgar-ms: Could you add a docstring explaining why this is needed? It's obvious now, but it may look a little odd in three months time (the `Value '0' is not JSON-serializable` thing has wasted my time in the past as well)

romanlutz: This is indeed very interesting and thanks for sharing these links! I agree with @riedgar-ms. Documenting this may turn out to be useful later on.

Issue #279, opened 2021-01-24
Error Analysis and Interpretability Scalability Design Decisions


mesameki:
Issue: Error analysis heatmap and treemap views can be calculated on very large datasets. Interpretabiliy, however, downsamples to 5K and only generates that many local explanations and render them in the UI. This mismatch of scalability capabilities needs to be navigated and addressed in the UI to not limit the error analysis functionalities and features. We discussed the following solution: We will calculate Error analysis tree map and heat map on ALL DATA, but only generate explanations for 5K subsampled dataset. Now, users going thorugh the dashboard can see the following options:
 
 
-- User selects a cohort on the decision tree or selects one of more grids on the heatmap and clicks on “explanation”. Downsampling for explanation can lead to that cohort not having any explanations attached (In an unfortunate situation none of the 5K subsampled dataset covers the cohort datapoints). Solution we chose:

-------> [CHOSEN OPTION] Allow them to click on the explanation tab. But, disable only global and local explanation tabs and let them explore data explorer and ONLY see aggregate insights on all data. Aggregate insights calculations will be kicked off from UI and happen on ALL Data not subsampled data.

 
-- User selects a cohort on the decision tree or selects one of more grids on the heatmap and clicks on “explanation”. Downsampling for explanation can lead to that cohort not having any explanations attached (In an unfortunate situation none of the 5K subsampled dataset covers the cohort datapoints). Solution we chose:

-------> [CHOSEN OPTION] Allow them to click on the explanation tab. But, disable only global and local explanation tabs and let them explore data explorer and ONLY see aggregate insights on all data. Aggregate insights calculations will be kicked off from UI and happen on ALL Data not subsampled data.

-- User selects a cohort on the decision tree or selects one of more grids on the heatmap and clicks on “explanation”. Downsampling for explanation can lead to that cohort having some explanations attached (explanations generated for the downsampled dataset has some overlap with that cohort datapoints). Solution we chose:

-------> [CHOSEN OPTION] 1. Data explorer will show aggregate insights of all cohort data points of that cohort.
-------> [CHOSEN OPTION] 2. Global explanation will have a warning on top stating that this is overall explanations of the subsampled dataset. (if the explanations are for less than X datapoints, maybe we go with the previous approach of assuming no explanations are there…)
-------> [CHOSEN OPTION] 3. [For end of Jan] local explanation tab only shows the cohorts data points that have explanations.

[...]

PR #1098, opened 2021-12-28
add warning message for downsampled explanation data in ErrorAnalysis UI

imatiach-msft:
Resolves last remaining part of issue https://github.com/microsoft/responsible-ai-toolbox/issues/279

The last remaining piece was to add a warning message when the UI has fewer rows loaded than the tree view or heatmap.

image

image

### DIFF
--- a/libs/localization/src/lib/en.json
+++ b/libs/localization/src/lib/en.json
@@ -345,6 +345,7 @@
     "instanceView": "Instance View",
     "localExplanationView": "Local explanation",
     "noFeature": "No feature",
+    "scaleWarning": "For scale reasons of the UI explanations are randomly downsampled",
     "whatIfDatapoints": "What-If datapoints"
   },
   "Fairness": {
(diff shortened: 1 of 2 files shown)


libs/localization/src/lib/en.json
```diff
@@ -345,6 +345,7 @@
     "instanceView": "Instance View",
     "localExplanationView": "Local explanation",
     "noFeature": "No feature",
+    "scaleWarning": "For scale reasons of the UI, explanations are randomly downsampled",
```
romanlutz, line 348: nit: is the comma meant to be there? I'm actually not entirely sure...

romanlutz (approved):
This warning is very useful! Thanks for adding.

libs/localization/src/lib/en.json
imatiach-msft: oh maybe I can remove it

PR #1099, opened 2021-12-28
remove csv files from source for cognitive services example notebooks

imatiach-msft:
CSV files should not live in git/github because git isn't designed to handle them well and they also make it really slow to clone a repository and sometimes even to perform git operations.  This PR removes two datasets that were recently contributed and moves them to public blob storage instead.


romanlutz (approved):
Thanks for cleaning this up!

PR #1116, opened 2022-01-06
Update PR template to be more intuitive

romanlutz:
## Description

Updates the PR template so that each set of checkboxes has at least one that should be checked. I went through this myself once with a PR and found that I always expect to check at least one, so this change makes this a bit more obvious.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] raiwidgets
- [ ] responsibleai
- [ ] erroranalysis
- [ ] rai_core_flask

## Tests

- [x] No new tests required.

## Documentation:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -42,5 +44,6 @@ Python packages changed:
 
 <!--- Put an `x` in all the boxes that apply. -->
 
-- [ ] My change requires a change to the documentation.
-- [ ] I have updated the documentation accordingly.
+- [ ] My change does not require any changes to the documentation.
+- [ ] My change requires an update to the documentation which does not happen in this PR. <!--- Elaborate below on when the update will happen. -->
+- [ ] My change requires an update to the documentation which is part of this PR.
```
gaugup, line 49: For the documentation section, we should have a third option as well. Lots of PR as bug fixes and may not always have a doc change.

 My change requires a change to the documentation.
 I have updated the documentation accordingly.
 **My change doesn't change any documentation**

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -20,13 +20,15 @@ npm packages changed:
 - [ ] responsibleai/localization
 - [ ] responsibleai/mlchartlib
 - [ ] responsibleai/model-assessment
+- [ ] No npm packages were changed.
```
imatiach-msft, line 23: is there a better type of checkbox we can use?
this is kind of annoying to see on PRs:
image

.github/PULL_REQUEST_TEMPLATE.md
romanlutz: Isn't that the first option? "My change does not require any changes to the documentation."

As far as I can tell there's only 3 paths: no changes, changes required but I'm not doing them in this PR, changes required and I did them.

.github/PULL_REQUEST_TEMPLATE.md
romanlutz: Oh I hadn't seen that before! This is kind of annoying. Let me check! I believe ✔️ `:heavy_check_mark:` should work, but I'll have to try!

.github/PULL_REQUEST_TEMPLATE.md
gaugup: sorry, I probably missed this.

gaugup (approved):

.github/PULL_REQUEST_TEMPLATE.md
romanlutz: No problem! I totally agree and didn't like the way I had originally written it. That's basically the motivation for the PR, so I'm glad we're on the same page 🙂

.github/PULL_REQUEST_TEMPLATE.md
romanlutz: ❌  for "no" and ✔️ for "yes" should be good! Updating this right now! LMK what you think.

imatiach-msft: hmm that's kind of annoying because you can't just check the "viewed" version, you have to modify the underlying text, but maybe that's the best we can do on github

imatiach-msft (approved):

.github/PULL_REQUEST_TEMPLATE.md
imatiach-msft: huh, maybe what we should do is keep the checkboxes, but rephrase them to be tasks, like:

But then for the "npm packages changed" I guess it doesn't really make sense.  Actually, is there a reason we should keep track of the npm packages that were changed?  The release process doesn't really care about this, does it?
Similarly, I'm not sure about python packages changed - would it help to keep track of that?

imatiach-msft: I don't know, maybe this PR is fine.  Maybe we just need to see how contributors use it and get feedback from them.

romanlutz: True, but the "viewed" version is identical with the task list, so you can't have both that and then not the task list in the PR overview. 🙁 

To be honest, the most important part of this PR template for me is to get all of us to write nice descriptions for PRs and include screenshots where useful. I know we haven't been great on this in the past, but it just helps so much and isn't a lot of work for the PR author. [I would add that your descriptions are typically top notch @imatiach-msft ! If all of us did it like you we wouldn't need a template as a reminder at all...]

In addition, we've added lots of features without tests in the past. The corresponding section is there as a reminder to make sure we don't do that anymore. A simple checkbox like yours might accomplish the same. Same for documentation.

The section about packages is there with the hope to eventually be able to auto-extract information on which changes happened to which packages. Perhaps that's not as important, at least not urgently or immediately important... (?)

So perhaps let me minimize it to 
- description
- screenshots
- the two checkboxes you mentioned above.
I'll update the PR later today.

romanlutz:
Since there were quite substantial changes since your reviews I'll reset your approvals @imatiach-msft and @gaugup. I reduced the sections significantly based on @imatiach-msft 's feedback (see corresponding thread). Let me know what you think!

.github/PULL_REQUEST_TEMPLATE.md
gaugup: I kind of like the checkbox format better. It gives a very good high level view of the scope of changes in the PR. 

I feel the description writing varies with author to author and also with the time pressure. It is just really hard to enforce on every body. With the checkbox system, at least there is some control over the descriptions.

imatiach-msft (approved):
looks good to me!

.github/PULL_REQUEST_TEMPLATE.md
romanlutz: Couldn't agree more (in fact that's my motivation for the PR). However, the fact that it displays it as tasks (like in Ilya's screenshot) worries me. Could be interpreted as "there's a lot of work to be done" rather than the actual purpose of the checkboxes.

One alternative I could think of: add the various packages as a simple list (no checkboxes) and the author has to delete all the ones they're not editing so that only the right ones remain. That's easy enough, just like clicking the boxes. 

I do like the simplified single box for tests and documentation. They're really just a nudge in the right direction after all.

I'll definitely wait to hear back from @gaugup 🙂

romanlutz:
@gaugup any thoughts? This is waiting on your input.

.github/PULL_REQUEST_TEMPLATE.md
gaugup: @romanlutz I like "
One alternative I could think of: add the various packages as a simple list (no checkboxes) and the author has to delete all the ones they're not editing so that only the right ones remain. That's easy enough, just like clicking the boxes.". 

I don't however see the corresponding change in this PR.

romanlutz: Yes, that was only an idea, but I didn't put it in yet because I'm not entirely sure of the value proposition myself. Initially I had thought perhaps we can simplify release notes by generating them off of the checkboxes, but if people just delete items off of this list and perhaps even add new ones that makes it tricky. Perhaps it's best to keep it simple and without the section describing the areas changed. The description of the PR can take care of that after all. Do you see value in the programmatic generation of release notes?

romanlutz: Since there was no more input I'm going to assume it's fine. It doesn't seem like anyone particularly like the additional work the current template caused, so let's see how the new one does and we can always change it.

PR #1156, opened 2022-01-20
add a perf test for validating training the surrogate error tree on large data and a minor optimization for traversing the tree

imatiach-msft:
## Description

- Add a performance test for validating training the surrogate error tree on large data (>1 million rows)
- In profiling, about half of the time is actually spent in traversal.  This PR adds a minor optimization for traversing the tree.

Note: I believe another optimization can be done for traversal by reducing the number of columns in the traversed dataset to only the features that the model splits on - or, possibly, even changing the dataset to numpy and doing the queries there, if that is more efficient.

There was another issue I found which was more concerning though.  It seems that we are calling the /tree API three times for some reason on initial load and twice when changing features when running on large data.  I will need to look into UX logic more to see why that is, my suspicion is that it is due to retry logic since the result only comes back after several seconds.

Note the optimization in this PR was very small.  On a 2.5 million row dataset the change in time was:
before:
5.826886177062988
after:
4.795517444610596
To traverse the model.  I still think the minor optimization is worth having though.  Removing not-needed columns will probably reduce this time much more though (hopefully!).

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] raiwidgets
- [ ] responsibleai
- [x] erroranalysis
- [ ] rai_core_flask

## Tests

- [ ] No new tests required.

## Documentation:

### DIFF
--- a/erroranalysis/tests/test_surrogate_error_tree.py
+++ b/erroranalysis/tests/test_surrogate_error_tree.py
@@ -1,10 +1,15 @@
 # Copyright (c) Microsoft Corporation
 # Licensed under the MIT License.
 
+import time
+
 import pytest
-from common_utils import (create_adult_census_data, create_iris_data,
-                          create_kneighbors_classifier,
-                          create_models_classification)
+from common_utils import (create_adult_census_data,
+                          create_binary_classification_dataset,
+                          create_iris_data, create_kneighbors_classifier,
+                          create_models_classification,
+                          create_sklearn_random_forest_regressor,
+                          replicate_dataset)
 
 from erroranalysis._internal.constants import (DIFF, LEAF_INDEX, PRED_Y,
                                                SPLIT_FEATURE, SPLIT_INDEX,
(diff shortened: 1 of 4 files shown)


gaugup:
Where is the perf improvement coming from in the traversing of the tree? It is not apparent from the diff for me.

imatiach-msft:
@gaugup it comes from query being slower, and using comparison directly:
df = df[df[p_node_name_val] <= parent_threshold]
instead of:
query = p_node_query + " <= " + str(parent_threshold)
df.query(query)
but the improvement is still really minor.  I think the most useful thing about this PR is the test validating training is indeed very fast.

erroranalysis/tests/test_surrogate_error_tree.py
```diff
@@ -41,6 +47,44 @@ def test_surrogate_error_tree_int_categorical(self):
...
+    def test_large_data_surrogate_error_tree(self):
+        # validate tree trains quickly for large data
+        X_train, y_train, X_test, y_test, _ = \
+            create_binary_classification_dataset(100)
+        feature_names = list(X_train.columns)
+        model = create_sklearn_random_forest_regressor(X_train, y_train)
+        for _ in range(16):
+            X_test = pd.concat([X_test, X_test], ignore_index=True)
+            y_test = np.concatenate([y_test, y_test])
+        assert X_test.shape[0] > 1000000
```
gaugup, line 59: Can you move the above logic to create a large test set into a utility function in common_utils?

erroranalysis/tests/test_surrogate_error_tree.py
```diff
@@ -41,6 +47,44 @@ def test_surrogate_error_tree_int_categorical(self):
...
+                                           num_leaves,
+                                           min_child_samples,
+                                           cat_ind_reindexed)
+        t1 = time.time()
+        execution_time = t1 - t0
+        print(execution_time)
+        # assert we don't take too long to train the tree on 1 million rows
+        # note we train on >1 million rows in ~1 second
+        assert execution_time < 5
+        model_json = surrogate._Booster.dump_model()
```
gaugup, line 83: In my experience assert on execution time can be really flaky. The execution time depends on a lot of factors like avaible CPU, memory and may not be consistently less than 5 seconds here. Maybe the above print is suffice.

gaugup (changes requested):

erroranalysis/tests/test_surrogate_error_tree.py
imatiach-msft: done!

erroranalysis/tests/test_surrogate_error_tree.py
imatiach-msft: I expanded it to 20 seconds then - certainly, if it only takes a second to train on >1 million rows, and suddenly it takes 20, then something must be wrong.  We can always increase the time more later if we see it failing.

gaugup (approved):

PR #1171, opened 2022-01-25
Add excludes cohort filter test in erroranalysis

gaugup:
## Description

It seems that this test was missing. Also moved the cohort filter methods under a class and changed the references.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] raiwidgets
- [ ] responsibleai
- [x] erroranalysis
- [ ] rai_core_flask

## Tests

- [ ] No new tests required.

## Documentation:


erroranalysis/erroranalysis/_internal/constants.py
```diff
@@ -11,8 +11,20 @@
...
+    """Cohort filter methods.
+    """
+
+    METHOD_INCLUDES = 'includes'
+    METHOD_EXCLUDES = 'excludes'
+    METHOD_EQUAL = 'equal'
+    METHOD_GREATER = 'greater'
+    METHOD_LESS = 'less'
+    METHOD_LESS_AND_EQUAL = 'less and equal'
+    METHOD_GREATER_AND_EQUAL = 'greater and equal'
```
romanlutz, line 26: This should technically be less than or equal, or at least less or equal (stressing on the OR rather than AND). I know that you're just moving this around from another file, but I'm just wondering whether we're adjusting those at some point.

romanlutz (approved):
Thanks for adding the new test case!

erroranalysis/erroranalysis/_internal/constants.py
imatiach-msft: this would need to be fixed everywhere, ui and backend, and wouldn't be backwards compatible, so I would be careful when making this change.  Chris also had a proposal around changing these to be more like "lt", "gt", etc, but it was not implemented, probably because it would be such a large change.

imatiach-msft (approved):

erroranalysis/erroranalysis/_internal/constants.py
romanlutz: I remember @xuke444 mentioning this last week. However, "lt" etc. aren't really readable. Not sure if that's a concern but it's worth thinking about. In any case, all of this is out of scope for this PR. Would just be nice to know what the long-term plan is on this.

PR #1174, opened 2022-01-25
Add metrics to StatisticsUtils - F1 score, selection rate, mean absolute error

romanlutz:
## Description

This PR adds additional metrics to `StatisticsUtils` that will be required for the fairness integration. This includes
- F1 score
- selection rate
- mean absolute error

The rendering was slightly off due to the long strings, which happened even before (see long ones like FPR), but was exacerbated by adding more metrics which then exceeded the box size:
image
For this reason I've extended the box width slightly to 255px and validated that it doesn't affect the experience:
image

Additionally, I added a `key` field to `ILabeledStatistic`. Before, it had only `label` and `stat`. However, `label` is localized. Since we'll need to search through a list of `ILabeledStatistic` in a follow-up PR, it seemed wrong to do that on localized strings as it could result in various kinds of issues. Instead, I added a `key` field and created two enums to list all calculated classification and regression metrics. These will be used in the UI code in the follow-up PR to decide which metrics to render.

Note: for multiclass classification we don't currently show the same metric box.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [x] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] raiwidgets
- [ ] responsibleai
- [ ] erroranalysis
- [ ] rai_core_flask

## Tests

- [ ] No new tests required.
- [x] I validated the changes manually.

classification:
image

regression:
image

## Documentation:


imatiach-msft (approved):

libs/core-ui/src/lib/util/StatisticsUtils.ts
```diff
@@ -8,10 +8,28 @@ import { ModelTypes } from "../Interfaces/IExplanationContext";
...
+  FalsePositiveRate = "falsePositiveRate",
+  SelectionRate = "selectionRate",
+  F1Score = "f1Score"
+}
+
+export enum RegressionMetrics {
+  MeanSquaredError = "meanSquaredError",
+  MeanAbsoluteError = "meanAbsoluteError",
+  MeanPrediction = "meanPrediction",
+  RSquared = "rSquared"
```
imatiach-msft, line 30: I sometimes see this called as coefficient of determination:
https://github.com/microsoft/responsible-ai-toolbox/blob/775c2f352ece316a3c575016b00d16d6a680f7b5/libs/localization/src/lib/en.json#L1044
I wonder if we can make the ^2 look like squared using unicode, or use the other term instead, any thoughts about that?

imatiach-msft: maybe you can use this ^2 unicode character: R²

imatiach-msft: I think this comment applies to the en.json instead

romanlutz: I'll see if I can make it work! Thanks for the suggestion

romanlutz: added superscript 2. Using the unicode encoding somehow didn't work, but just copying the character in did.

romanlutz: image

PR #1179, opened 2022-01-26
refactor matrix filter and area state to be private static

imatiach-msft:
## Description

This PR is a refactor of the MatrixFilter and MatrixArea, there is no new functionality.
The PR moves the state to be within the MatrixFilter and MatrixArea as private static properties, and hence the parent error analysis and model analysis dashboards don't need to keep track of it.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret
- [x] responsibleai/error-analysis

Python packages changed:

- [ ] raiwidgets
- [ ] responsibleai
- [ ] erroranalysis
- [ ] rai_core_flask

## Tests

- [x] No new tests required.

## Documentation:


libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/Matrix/MatrixArea/MatrixArea.tsx
```diff
@@ -56,24 +59,25 @@ export class MatrixArea extends React.PureComponent<
   IMatrixAreaState
 > {
   public static contextType = ModelAssessmentContext;
+  private static savedState: IMatrixAreaState | undefined;
```
romanlutz, line 62: It seems very uncommon to add state on the class level while at the same time having the state of the React component have the same `IMatrixAreaState` type. Why not just use it as state on `MatrixArea`? 

I agree that passing in the state from the class that contains this as `this.props.state....` is not at all a good pattern, but I'm not sure we're getting to the best solution this way either.

Perhaps someone with more React experience than me should chime in 🙂  @xuke444 or others?

xuke444: Yes, it is not typical.
But this one is for store state for unloaded component, so that user would not lost their selection/input when they come back.
Using private property here will make this component be isolated, compare to pass back to props (which creates a long dependency chain) it is much cleaner.

romanlutz: Fair enough. Hopefully we'll find a better solution at some point, but I don't know of one either...

zhb000 (approved):

PR #1181, opened 2022-01-28
Make causal manager add() and compute() behavior similar to other managers

gaugup:
## Description

It seems causal manager is different than other managers where causal manager computes causal effects in the add() function and doesn't do anything in the compute() function. This is different in comparison to other managers where other managers seem to validate and add config in add() and compute the RAI insight in compute(). Hence, making the causal manager validate and add the config in add() and computing the causal effects in compute().

This change will break the back compat for azureml where the expectation is add() will also compute the causal effects. A SDK upgrade in azureml will ensure that we call compute() and get() to get the causal effect.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] raiwidgets
- [x] responsibleai
- [ ] erroranalysis
- [ ] rai_core_flask

## Tests

- [x] No new tests required.

## Documentation:


responsibleai/responsibleai/managers/causal_manager.py
```diff
@@ -321,8 +257,83 @@ def _whatif(self, id, X, X_feature_new, feature_name, y, alpha=0.1):
...
 
     def compute(self):
-        """No-op function to comply with model analysis design."""
-        pass
+        """Computes the causal effects by running the causal
+           configuration."""
+        is_classification = self._task_type == ModelTask.CLASSIFICATION
+        for result in self._results:
+            causal_config = result.config
+            if not result.is_computed:
```
imatiach-msft, line 265: where is result.is_computed set to true?

imatiach-msft (approved):

responsibleai/responsibleai/managers/causal_manager.py
```diff
@@ -321,8 +257,83 @@ def _whatif(self, id, X, X_feature_new, feature_name, y, alpha=0.1):
                               y, alpha=alpha).to_dict(orient="records")
 
     def compute(self):
-        """No-op function to comply with model analysis design."""
-        pass
+        """Computes the causal effects by running the causal
```
imatiach-msft, line 260: it looks like the documentation on the add() function needs to be updated, as it no longer can return a result, so that doc needs to be removed I think?

responsibleai/responsibleai/_tools/causal/causal_result.py
```diff
@@ -50,6 +50,13 @@ def __init__(
         self.local_effects = local_effects
         self.policies = policies
 
+    @property
+    def is_computed(self):
```
imatiach-msft, line 54: not sure if you want to add a doc here, it's not necessary since it's a private function, but it seems like we've been trying to document even the private functions lately

responsibleai/responsibleai/managers/causal_manager.py
imatiach-msft: oh, nvm, I think I got confused - it looks like this isn't a bool property here but instead we are checking:
```
self.causal_analysis is not None or \
            self.global_effects is not None or \
            self.local_effects is not None or \
            self.policies is not None
```
I guess this is a little different from the other configs

vinuthakaranth (approved):

responsibleai/responsibleai/_tools/causal/causal_result.py
gaugup: Done.

responsibleai/responsibleai/managers/causal_manager.py
gaugup: Done.

PR #1186, opened 2022-01-31
Add cohort and filter definitions in raiwidgets SDK

gaugup:
## Description

Add classes for cohort definitions. The **Cohort** class is the actual class that will be defined by the user. **CohortFilter** class will help with the management of cohort filter values like the cohort filter method (among greater, greater and equal, less, less and equal, equal, includes, excludes and in the range of), the values for filtering and the user column name. The **CohortFilterMethods** defines the cohort filter methods.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [x] raiwidgets
- [ ] responsibleai
- [ ] erroranalysis
- [ ] rai_core_flask

## Tests

- [ ] No new tests required.

## Documentation:

- [x] I have updated the documentation accordingly.


tongyu-microsoft (approved):

raiwidgets/raiwidgets/_cohort.py
```diff
@@ -0,0 +1,186 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+"""Module for defining cohorts in raiwidgets package."""
```
imatiach-msft, line 4: I guess my biggest question here is whether this should live in raiwidgets package, responsibleai package, or be its own package.  The problem with it being in raiwidgets is that then it can't be re-used in, for example, erroranalysis, since raiwidgets depends on erroranalysis but not the other way around.  In case other packages (like RAI mitigations) want to re-use this, it will be more difficult if it is in raiwidgets package.

raiwidgets/raiwidgets/_cohort.py
```diff
@@ -0,0 +1,186 @@
...
+    """Defines different methods for cohort filters.
+    """
+    GreaterThan = "greater"  # Strictly greater than a value
+    # Greater than or equal to a value
+    GreaterThanEqualTo = "greater and equal"
+    LessThan = "less"  # Strictly lesser than a value
+    LessThanEqualTo = "less and equal"  # Lesser than or equal to a value
+    Equal = "equal"  # Equal to a value
+    Includes = "includes"  # Includes a set of values
+    Excludes = "excludes"  # Excludes a set of values
```
imatiach-msft, line 23: I think these constants should live in one python place.  I see them in error analysis code and now I see them here as well. I think it might be better to unify them somehow into one common constants file?

imatiach-msft: see C:\responsible-ai-widgets\erroranalysis\erroranalysis\_internal\constants.py and where they are used

```
class CohortFilterMethods:
    """Cohort filter methods.
    """

    METHOD_INCLUDES = 'includes'
    METHOD_EXCLUDES = 'excludes'
    METHOD_EQUAL = 'equal'
    METHOD_GREATER = 'greater'
    METHOD_LESS = 'less'
    METHOD_LESS_AND_EQUAL = 'less and equal'
    METHOD_GREATER_AND_EQUAL = 'greater and equal'
    METHOD_RANGE = 'in the range of'
```

raiwidgets/raiwidgets/_cohort.py
```diff
@@ -0,0 +1,186 @@
...
+
+    :param method: Cohort filter method from one of CohortFilterMethods.
+    :type method: str
+    :param arg: List of values to be used by the cohort filter.
+    :type arg: list
+    :param column: The column name from the dataset on which the filter
+                   will be applied.
+    :type column: str
+    """
+    def __init__(self, method: str, arg: List[Any], column: str):
```
imatiach-msft, line 71: it looks like this doesn't support composite filter, but maybe that's ok for the initial version

raiwidgets/raiwidgets/_cohort.py
```diff
@@ -0,0 +1,186 @@
...
+class Cohort:
+    """Defines the cohort which will be injected from SDK into the Dashboard.
+
+    :param method: Name of the cohort.
+    :type method: str
+    """
+    def __init__(self, name: str):
+        """Defines the cohort which will be injected from SDK into the Dashboard.
+
+        :param method: Name of the cohort.
```
imatiach-msft, line 165: is it "method" or "name" for the param here?

```
def __init__(self, name: str):
...
    :param method: Name of the cohort.
```

raiwidgets/raiwidgets/_cohort.py
```diff
@@ -0,0 +1,186 @@
...
+        :type: CohortFilter
+        """
+        if self.cohort_filter_list is None:
+            self.cohort_filter_list = [cohort_filter]
+        else:
+            self.cohort_filter_list.append(cohort_filter)
+
+    def _validate_cohort_with_test_data(
+            self, test_data: pd.DataFrame):
+        # TODO: Add validations for checking cohort and cohort filter against
```
imatiach-msft, line 184: maybe remove this for now, since it's not implemented anyway?

imatiach-msft (approved):

raiwidgets/raiwidgets/_cohort.py
gaugup: Removed it.

raiwidgets/raiwidgets/_cohort.py
gaugup: Changed it to name in doc strings.

raiwidgets/raiwidgets/_cohort.py
gaugup: The cohort defined from the UI (not from error analysis)  is not of type Composite filter. So on similar lines keeping the initial implementation simple to just include filter.

raiwidgets/raiwidgets/_cohort.py
gaugup: I agree with all of the above. I do see that in future the cohort and filter definitions will go in separate package. Currently, for the pre-built cohort feature it is ok to keep it in raiwidgets. When this feature is complete we can think about moving these class definitions into a separate package and take dependency on it.

raiwidgets/raiwidgets/_cohort.py
gaugup: I think this circles back to the question if we should have a common package for cohorts. In case we do, we can condense all the constants for the cohorts in a common place.

I don't want to take dependency on error analysis for this class. I am introducing the ALL and SINGLE_VALUE_METHODS which I think are not going to be used in error analysis. Hence, I intend to keep this class in _cohort.py. I would change the method variables to match the uppercase as above.

PR #1196, opened 2022-02-02
Return exception details to UI

tongyu-microsoft:
This PR returns detailed error strings to UI when hitting exceptions for on_predict, debug_ml, matrix, importances and causal_whatif.

## Description

Currently when we hit errors from UI side, we can only get a generic error message saying that "Failed to..." without detailed error message. This PR returns exception error string to UI, so we can handle errors properly on UI side.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [x] raiwidgets
- [ ] responsibleai
- [ ] erroranalysis
- [ ] rai_core_flask

## Tests

- [x] No new tests required.

## Documentation:

### DIFF
--- a/raiwidgets/raiwidgets/responsibleai_dashboard_input.py
+++ b/raiwidgets/raiwidgets/responsibleai_dashboard_input.py
@@ -51,9 +51,11 @@ class ResponsibleAIDashboardInput:
         except Exception as e:
             print(e)
             traceback.print_exc()
+            e_str = _format_exception(e)
             return {
                 WidgetRequestResponseConstants.error: "Model threw exception"
-                " while predicting...",
+                " while predicting..."
+                "inner error: {}".format(e_str),
                 WidgetRequestResponseConstants.data: []
             }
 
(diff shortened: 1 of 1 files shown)


gaugup (approved):

raiwidgets/raiwidgets/responsibleai_dashboard_input.py
```diff
@@ -101,9 +106,11 @@ def matrix(self, data):
         except Exception as e:
             print(e)
             traceback.print_exc()
+            e_str = _format_exception(e)
             return {
                 WidgetRequestResponseConstants.error:
-                    "Failed to generate json matrix representation",
+                    "Failed to generate json matrix representation,"
+                    "inner error: {}".format(e_str),
```
imatiach-msft, line 113: I'm a bit worried if we return these in some other platform that they may reveal private user information, like column names etc, so anything that calls these will need to sanitize them?

imatiach-msft (approved):

raiwidgets/raiwidgets/responsibleai_dashboard_input.py
imatiach-msft: it seems this is fine after discussion because it's not logged anywhere anyway

tongyu-microsoft: Talked offline, in AzureML, we will make sure we don't log any detailed error message which could contain user's private information.

PR #1207, opened 2022-02-09
RFC add metadata_columns to RAIInsights

romanlutz:
## Description

This PR includes changes to
- add a `metadata_columns` argument on `RAIInsights` (including input validation)
- passes only features (excluding `metadata_columns` using `drop`) to `predict` and `predict_proba` methods
- adjust managers:
  - causal needs no change since it doesn't use the model
  - error analysis needs to be aware of `metadata_columns` to use them for the error analysis calculations, but at the same time avoid passing `metadata_columns` to the model
  - explanations and counterfactuals just need to ignore `metadata_columns` so we don't pass them to that manager

Note that this is a DRAFT PR so I have deliberately not included the following yet since I want to work through potential comments:
- tests
- notebook adjustments and additions

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] raiwidgets
- [x] responsibleai
- [x] erroranalysis
- [ ] rai_core_flask

## Tests

- [ ] No new tests required.

## Documentation:


erroranalysis/erroranalysis/analyzer/error_analyzer.py
```diff
@@ -40,6 +40,10 @@ class BaseAnalyzer(ABC):
     :type feature_names: numpy.array or list[]
     :param categorical_features: The categorical feature names.
     :type categorical_features: list[str]
+    :param metadata_columns: The set of columns that are not passed
+        to the model or explainers. These columns can be used for
+        other analyses.
```
gaugup, line 45: Can you elaborate more on what other analysis is intended?

responsibleai/responsibleai/managers/error_analysis_manager.py
```diff
@@ -202,19 +172,25 @@ def __init__(self, model, dataset, target_column, classes=None,
         :type classes: list
         :param categorical_features: The categorical feature names.
         :type categorical_features: list[str]
+        :param metadata_columns: The set of columns that are not passed
```
gaugup, line 175: Can you elaborate more on what other analysis is intended?

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -96,12 +72,17 @@ def __init__(self, model, train, test, target_column,
         :type task_type: str
         :param categorical_features: The categorical feature names.
         :type categorical_features: list[str]
+        :param metadata_columns: The set of columns that are not passed
```
gaugup, line 75: Can you elaborate more on what other analysis is intended?

erroranalysis/erroranalysis/analyzer/error_analyzer.py
romanlutz: I don't necessarily want to keep a running list of what we do with this in 10 different places. It's metadata, i.e., it provides data on your data and you can use it to analyze your data with more information than what you just get from your data. One such application is disaggregated analysis, but you can also use it for error analysis, or in fact for causal analysis. In the future we may have others, too.

responsibleai/responsibleai/rai_insights/rai_insights.py
romanlutz: same as in the other comment

erroranalysis/erroranalysis/analyzer/error_analyzer.py
riedgar-ms: I think a 'such as' phrase could be useful here, but agree that it doesn't need to be length.

erroranalysis/erroranalysis/analyzer/error_analyzer.py
```diff
@@ -445,6 +451,10 @@ class and for the regression case a method of predict()
...
+    :type metadata_columns: list[str]
```
riedgar-ms, line 457: optional[list[str]] if `None` is valid?

riedgar-ms: Although we haven't done that on the others

responsibleai/responsibleai/_interfaces.py
```diff
@@ -20,6 +20,7 @@ class Dataset:
     class_names: List[str]
     categorical_features: List[str]
     target_column: str
+    metadata_columns: List[List]
```
riedgar-ms, line 23: This is perhaps a place where is should be `Optional[List[str]]` ?

responsibleai/responsibleai/managers/counterfactual_manager.py
```diff
@@ -664,8 +666,10 @@ def _load(path, rai_insights):
 
         # Rehydrate model analysis data
         inst.__dict__[CounterfactualManager._MODEL] = rai_insights.model
-        inst.__dict__[CounterfactualManager._TRAIN] = rai_insights.train
-        inst.__dict__[CounterfactualManager._TEST] = rai_insights.test
+        inst.__dict__[CounterfactualManager._TRAIN] = \
+            rai_insights.train.drop(columns=rai_insights.metadata_columns)
```
riedgar-ms, line 670: I'm not sure about what pandas does behind the scenes (or for that matter when Python does references or values) but.... we have this happening a lot, and it's going to be doing copies each time. Should we introduce `RAIInsights._X_test = RAIInsights.train.drop(columns=rai_insights.metadata_columns)` and just do it once at top level?

responsibleai/responsibleai/managers/error_analysis_manager.py
```diff
@@ -165,28 +152,11 @@ def from_json(json_str):
 
 
 class ErrorAnalysisManager(BaseManager):
-
-    """Defines the ErrorAnalysisManager for discovering errors in a model.
```
riedgar-ms, line 169: This change feels like it should be a different PR....

riedgar-ms (commented):
I like the basic idea, but should we just do one 'drop' in the top level `RAIInsights` object and then reference those?

erroranalysis/erroranalysis/analyzer/error_analyzer.py
romanlutz: I'm actually taking the unrelated doc adjustments out of this PR into a separate one. Perhaps I can do that there!

responsibleai/responsibleai/managers/counterfactual_manager.py
romanlutz: This does do a copy and it's a valid concern. I can try using a view like this:
```
all_cols = df.columns
cols=list(set(all_cols) - set(metadata_columns))
df[cols]  # this is a view
```
However, I have to try whether that works at all. You can't modify views without getting the following:
```
<stdin>:1: SettingWithCopyWarning:
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
```
and I don't know whether we're trying to do something anywhere. Will report back.

responsibleai/responsibleai/managers/error_analysis_manager.py
romanlutz: Good point. I am working on that right now and will do all unrelated doc changes there. Stay tuned!

romanlutz: #1209 exists now! Once it's merged the changes should vanish from here :-)

responsibleai/responsibleai/_interfaces.py
romanlutz: This inspired me to create #1214 . Once that's merged I'll add the same annotations here, too.

erroranalysis/erroranalysis/analyzer/error_analyzer.py
romanlutz: #1214 !

Also, the optional[list[str]] annotations are only used for type annotations on the args in code, not in the docstring, right? I may be wrong...

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -50,10 +51,12 @@ class RAIInsights(object):
     """
 
     def __init__(self, model, train, test, target_column,
-                 task_type, categorical_features=None, classes=None,
+                 task_type, categorical_features=None,
+                 metadata_columns=None, classes=None,
```
gaugup, line 55: Can you add this parameter in the end after maximum_rows_for_test for back compat reasons?

responsibleai/responsibleai/rai_insights/rai_insights.py
gaugup: should this be analysis instead of analyses?

responsibleai/responsibleai/managers/counterfactual_manager.py
romanlutz: I changed every occurrences of drop to a view through `df[cols]` and it didn't complain so I infer from that that we're not changing the underlying DataFrame at any later point which is great. This should also help performance. Thanks for the suggestion!

responsibleai/responsibleai/rai_insights/rai_insights.py
romanlutz: Logically, I felt like it makes most sense right after `categorical_features`, since it's also a list of column names. Realizing that these are positional arguments actually makes me want to make all but model, train, and test keyword-only:

```
func(self, model, train, test, *, target_column, task_type, categorical_features=None, ...)
```
In that case, the positioning after `*` is irrelevant since you have to call it with keyword
```
obj.func(model, train, test, target_column="y", categorical_features=["gender"], metadata_columns=["gender", "age"], task_type="classification", ...)
``` 

To me, it's pretty clear that there's no benefit at all to having them be positional. We've done something like this in Fairlearn a while ago: https://github.com/fairlearn/fairlearn/blob/862263f4352e23da088ae7a638b88b56580e5230/fairlearn/metrics/_metric_frame.py#L269
Since it's not a new class, but one that already exists, the situation is somewhat more complicated. Making any change might break code that would currently run fine. If I remember correctly, we ended up adding a warning that arguments will be keyword-only starting in the next released version and then made the change one version later. For this particular PR I won't get around moving it to the end it seems.

@xuke444 @riedgar-ms @gaugup @imatiach-msft

imatiach-msft: we only just very recently released this RAIInsights, so I think it's fine to change as long as we have a really really really good reason to.  I'm not quite sure if this one meets the bar.

responsibleai/responsibleai/_interfaces.py
romanlutz: Actually it's `List[str]` (updated just now). It doesn't need to be optional for this. It can be set as `[]`

erroranalysis/erroranalysis/analyzer/error_analyzer.py
romanlutz: added the annotations, but there will probably be a few merge conflicts to resolve once #1214  is merged.

erroranalysis/erroranalysis/analyzer/error_analyzer.py
romanlutz: Added!

responsibleai/responsibleai/rai_insights/rai_insights.py
romanlutz: No, intended to be plural. Added some more context.

responsibleai/responsibleai/managers/error_analysis_manager.py
romanlutz: Added a bit more context.

imatiach-msft:
@romanlutz not sure why these suddenly got included but we shouldn't be adding any datasets to git/source:
image
datasets make git slow, especially cloning, and git doesn't render/work with them well

romanlutz:
I'm closing this as we've decided to move in a somewhat different direction led by @gaugup. The RFC has therefore served its purpose well 🙂

PR #1218, opened 2022-02-10
make builds more reliable by adding retry logic to urlretrieve calls in notebooks

imatiach-msft:
…

## Description

Make builds more reliable by adding retry logic to urlretrieve calls in notebooks.
One recent build failed:

https://github.com/microsoft/responsible-ai-toolbox/runs/5146324656?check_suite_focus=true

This was due to urlretrieve calls in notebooks failing.
image
This PR adds retry logic to notebooks.
I think it may be useful to make it into a utility in some python package as well, perhaps in raiwidgets?  Then, after some time, we can migrate all of the notebooks to that utility.  Or, perhaps, we need to create some package just for notebooks and tests.  It seems it could be useful in other places as well (other repositories, maybe even internal code, etc).

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] raiwidgets
- [ ] responsibleai
- [ ] erroranalysis
- [ ] rai_core_flask

## Tests

- [ ] No new tests required.

## Documentation:


gaugup:
Can you add a utility file to the notebooks folder and add the retry there? Seems like a lot of repetitive code and harder to maintain if a fix needs to be made.

imatiach-msft:
@gaugup Yes the problem with the utility file is it tends to get lost when people copy the notebook, and it has to be just in the right folder structure.  I think a better fix might be to add this method to one of our packages or even to create a common "test/utils" pypi package.  This utility is actually reused in a ton of our notebooks across different repositories at this point.

vinuthakaranth (approved):

PR #1220, opened 2022-02-10
Refactor highchart defaultOptions

zhb000:
Refactor highchart defaultOptions

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [x] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] raiwidgets
- [ ] responsibleai
- [ ] erroranalysis
- [ ] rai_core_flask

## Tests

- [ ] No new tests required.

## Documentation:


libs/core-ui/src/lib/Highchart/HighchartWrapper.tsx
```diff
@@ -3,10 +3,10 @@
 
 import { ITheme } from "@fluentui/react";
 import _ from "lodash";
-import { mergeStyles } from "office-ui-fabric-react";
+import { getTheme, mergeStyles } from "office-ui-fabric-react";
```
imatiach-msft, line 6: why do we need both @fluentui/react and office-ui-fabric-react, I thought they were the same package?  Just for my understanding.

libs/core-ui/src/lib/Highchart/getDefaultHighchartOptions.ts
```diff
@@ -0,0 +1,98 @@
...
+        style: {
+          color: colorTheme.fontColor
+        }
+      },
+      title: {
+        text: undefined
+      }
+    },
+    yAxis: {
+      gridLineWidth: 1,
```
imatiach-msft, line 80: the x and y axis look similar and I see this code elsewhere, is there some way to refactor to some common function that returns this info?

imatiach-msft (approved):

libs/core-ui/src/lib/Highchart/HighchartWrapper.tsx
zhb000: They are, we are in the process of migrating from fabric ui to fluentui, eventually fluentui will be the one always imported from.

libs/core-ui/src/lib/Highchart/getDefaultHighchartOptions.ts
zhb000: this one is defaultOptions, each specific chart might have seperate data, labels passed in

PR #1227, opened 2022-02-15
Add data validations to SDK defined cohorts

gaugup:
## Description

The following validations need to be performed for cohort filter with
test data:-

High level validations
1. Validate if the filter column is present in the test data.
2. Validate if the filter column is present in the special column
   list.

"Index" Filter validations
1. The Index filter only takes integer arguments.
2. The Index filter doesn't take CohortFilterMethods.EXCLUDES
   filter method.

"Classification Outcome" Filter validations
1. Validate that "Classification Outcome" filter is not configure for
   multiclass classification and regression.
2. The "Classification Outcome" filter only contains values from set
   ClassificationOutcomes.
3. The "Classification Outcome" filter only takes
   CohortFilterMethods.INCLUDES filter method.

"Error" Filter validations
1. Validate that "Error" filter is not configure for
   multiclass classification and binary classification.
2. Only integer or floating points can be configured as arguments.
3. The CohortFilterMethods.INCLUDES and CohortFilterMethods.EXCLUDES
   filter methods cannot be configured for this filter.

"Predicted Y/True Y" Filter validations
1. The set of classes configured in case of classification is a
   superset of the classes available in the test data.
2. The CohortFilterMethods.INCLUDES is only allowed to be
   configured for "Predicted Y" filter in case of classification.
3. The CohortFilterMethods.INCLUDES and CohortFilterMethods.EXCLUDES
   filter methods cannot be configured for this filter for regression.

"Dataset" Filter validations
1. TODO:- For continuous features the allowed values that be configured
   should be within the range of minimum and maximum values available
   within the continuous feature column in the test data.
2. For categorical features only CohortFilterMethods.INCLUDES can be
   configured.
3. For categorical features the values allowed are a subset of the
   the values available in the categorical column in the test data.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [x] raiwidgets
- [ ] responsibleai
- [ ] erroranalysis
- [ ] rai_core_flask

## Tests

- [ ] No new tests required.

## Documentation:


raiwidgets/raiwidgets/_cohort.py
```diff
@@ -66,9 +79,20 @@ class CohortFilter:
                    will be applied.
     :type column: str
     """
+    PREDICTED_Y = 'Predicted Y'
+    TRUE_Y = 'True Y'
+    INDEX = 'Index'
+    CLASSIFICATION_OUTCOME = 'Classification Outcome'
```
imatiach-msft, line 85: 1.) it looks like the O should be lowercase based on what I see in error analysis code?  would be good to check.
2.) I think many of these are the same or similar to the constants in error analysis.  This really makes me wonder if something needs to be more "core" than error analysis, like a cohort package, and EA should take a dependency on this cohort package.

See:

https://github.com/microsoft/responsible-ai-toolbox/blob/main/erroranalysis/erroranalysis/_internal/cohort_filter.py#L14

raiwidgets/raiwidgets/_cohort.py
```diff
@@ -89,15 +113,34 @@ def __init__(self, method: str, arg: List[Any], column: str):
...
+        The following validations can be performed on the cohort filter:-
+
+        1. Verify the correct types for method (expected string), column
+           (expected string) and arg (expected list).
+        2. The method value should be one of the filter string from
+           CohortFilterMethods.ALL.
+        3. The arg shouldn't be an empty list.
+        4. For all cohort filter methods in
+           CohortFilterMethods.SINGLE_VALUE_METHODS, the value in the arg
+           should be integer or float and there should be utmost one value
```
imatiach-msft, line 133: "utmost one value"
Is this British English?  I'm more familiar with "at most", for example "at most one value", than the word "utmost".

raiwidgets/raiwidgets/_cohort.py
```diff
@@ -140,39 +183,286 @@ def _validate_cohort_filter_parameters(
...
+            raise UserConfigValidationException(
+                "The target_column {0} was not found in test_data.".format(
+                    target_column)
+            )
+
+        test_data_columns_set = set(test_data.columns) - set([target_column])
+        if not all(entry in test_data_columns_set
+                   for entry in categorical_features):
+            raise UserConfigValidationException(
+                "Found some categorical feature name which is not"
```
imatiach-msft, line 459: maybe specify via categorucal_features parameter?  so user knows where "categorical feature name" is coming from?

raiwidgets/tests/test_cohort.py
```diff
@@ -141,7 +150,343 @@ def test_cohort_filter_serialization_include_exclude_methods(self, method):
...
+                categorical_features=[]
+            )
+
+    def test_validate_with_test_data_classification_error_filter_validations(
+            self):
+        test_data_multiclass = pd.DataFrame(
+            data=[[23, 'X'], [25, 'Y'], [25, 'Z']],
+            columns=["age", "target"])
+
+        test_data_binary = pd.DataFrame(
```
imatiach-msft, line 202: I feel like these repeating test_data declarations should be refactored into some common method or utility

raiwidgets/tests/test_cohort.py
```diff
@@ -141,7 +150,343 @@ def test_cohort_filter_serialization_include_exclude_methods(self, method):
...
+            data=[[23, 'new', 'A'], [25, 'new, ''B'], [25, 'old', 'B']],
+            columns=["age", 'type', "target"])
+
+        with pytest.raises(
+                UserConfigValidationException,
+                match="{0} is a categorical feature and should be only "
+                      "configured with {1} cohort filter.".format(
+                          "type",
+                          CohortFilterMethods.METHOD_INCLUDES)):
+            cohort_filter = \
```
imatiach-msft, line 390: I think we should try to avoid using \
I wonder if the code could be less nested by refactoring some of the pytest.raises into helper functions, not sure.
Or maybe the cohort_filter could be defined prior to the pytest.raises.

raiwidgets/tests/test_cohort.py
```diff
@@ -141,7 +150,343 @@ def test_cohort_filter_serialization_include_exclude_methods(self, method):
...
+                CohortFilter(method=CohortFilterMethods.METHOD_EXCLUDES,
```
imatiach-msft, line 391: I don't understand why method excludes is a problem for categorical features, and only method includes is supported for that?  I thought both work for categorical features?

raiwidgets/tests/test_cohort.py
```diff
@@ -141,7 +150,343 @@ def test_cohort_filter_serialization_include_exclude_methods(self, method):
...
+                      "categorical columns."):
+            cohort_1._validate_with_test_data(
+                test_data=test_data,
+                target_column="target",
+                categorical_features={})
+
+        with pytest.raises(
+                UserConfigValidationException,
+                match="All entries in categorical_features "
+                      "need of string type."):
```
imatiach-msft, line 475: need of string type -> need to be of string type (?)

imatiach-msft (approved):

raiwidgets/tests/test_cohort.py
gaugup: Done

raiwidgets/tests/test_cohort.py
gaugup: I also thought the same. But it seems UI only allows to take "includes" filter. 
image

raiwidgets/tests/test_cohort.py
gaugup: reduced the use of \ throughout the this file.

raiwidgets/raiwidgets/_cohort.py
gaugup: Changed to say only one value.

raiwidgets/tests/test_cohort.py
gaugup: Looking to implement only the cases we support in UI currently.

raiwidgets/raiwidgets/_cohort.py
gaugup: Thanks fixed lowercase 'o' for 1.

For 2, we should look at the contract between UI and SDK and see if we can leverage the code in _cohort.py. Probably outside of scope of this PR but I am open to looking for consolidating opportunities and creating a separate package for cohorts.

raiwidgets/tests/test_cohort.py
gaugup: done. Made function for these toy datasets

raiwidgets/raiwidgets/_cohort.py
gaugup: Done. Added column name to exception to guide the user.

PR #1242, opened 2022-02-24
PreBuilt cohorts UX changes

gaugup:
## Description

The PR has the following changes:-

- Adding of interfaces `libs/core-ui/src/lib/Interfaces/IPreBuiltCohort.tsx` for user defined SDK cohorts and `libs/core-ui/src/lib/Interfaces/IPreBuiltFilter.ts` for user defined SDK cohort filters. The interfaces `IFilter` only allows for numeric value in `arg` so we need these new interfaces to handle the SDK defined cohorts.
- Added mock data in `apps/dashboard/src/model-assessment/__mock_data__/adultCensus.ts`, `apps/dashboard/src/model-assessment/__mock_data__/bostonData.ts` and `apps/dashboard/src/model-assessment/__mock_data__/wineData.ts` to be used in `apps/dashboard/src/app/applications.ts` for dashboard-e2e. 
- Added public APIs `getModelType()` and `getModelClasses()` in `libs/core-ui/src/lib/util/JointDataset.ts`.
- The translation logic of transforming SDK defined cohorts '`PreBuiltCohort` into UI `ErrorCohort`/`Cohort` is in `libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/ProcessPreBuiltCohort.ts`.
- The unit tests for translation of SDK defined cohorts to UI cohorts are in `libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/ProcessPreBuiltCohortBinaryClassification.test.ts` and `libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/ProcessPreBuiltCohortRegression.test.ts`. 
- The translated SDK cohorts are added in the global cohort list in file `libs/model-assessment/src/lib/ModelAssessmentDashboard/Context/buildModelAssessmentContext.ts`.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [x] responsibleai/core-ui
- [ ] responsibleai/counterfactuals
- [ ] responsibleai/dataset-explorer
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret
- [ ] responsibleai/localization
- [ ] responsibleai/mlchartlib
- [x] responsibleai/model-assessment

Python packages changed:

- [ ] raiwidgets
- [ ] responsibleai
- [ ] erroranalysis
- [ ] rai_core_flask

## Tests

- [ ] No new tests required.
- [x] New tests for the added feature are part of this PR.
- [ ] I validated the changes manually.

## Documentation:

[...]

PR #1255, opened 2022-03-03
Change procedural loops to functions map() and filter()

gaugup:
## Description

This is follow up to the comment in PR https://github.com/microsoft/responsible-ai-toolbox/pull/1242 to change the procedural loops to functional style in typescript.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] raiwidgets
- [ ] responsibleai
- [ ] erroranalysis
- [ ] rai_core_flask

## Tests

- [x] No new tests required.

## Documentation:


libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/ProcessPreBuiltCohort.ts
```diff
@@ -118,14 +118,9 @@ function translatePreBuiltCohortFilterForTarget(
...
-    const index: number[] = [];
-    for (const modelClass of preBuiltCohortFilter.arg) {
-      const indexModelClass = modelClasses.indexOf(modelClass);
-
-      if (indexModelClass !== -1) {
-        index.push(indexModelClass);
-      }
-    }
+    const index = preBuiltCohortFilter.arg
+      .map((modelClass) => modelClasses.indexOf(modelClass))
```
imatiach-msft, line 122: might using the reduce function be even better?  
https://stackoverflow.com/questions/16037049/why-does-javascript-map-function-return-undefined

like this example they had:
```
var arr = ['a','b',1];
var results = arr.reduce((results, item) => {
    if (typeof item === 'string') results.push(modify(item)) // modify is a fictitious function that would apply some change to the items in the array
    return results
}, [])
```
map and filter also looks fine to me though

imatiach-msft (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/ProcessPreBuiltCohort.ts
```diff
@@ -205,17 +197,13 @@ function translatePreBuiltCohortFilterForDataset(
...
       jointDataset.metaDict[jointDatasetFeatureName].sortedCategoricalValues;
     if (categorcialValues !== undefined) {
-      for (const categoricalValue of preBuiltCohortFilter.arg) {
-        const indexCategoricalValue =
-          categorcialValues.indexOf(categoricalValue);
-        if (indexCategoricalValue !== -1) {
-          index.push(indexCategoricalValue);
-        }
-      }
+      index = preBuiltCohortFilter.arg
```
imatiach-msft, line 204: I feel like all of these can just be refactored into one helper function.. although on the other hand I guess 3 lines of code isn't so much that it's really worth it to refactor perhaps?

gaugup: Yeah, map and filter are self explanatory constructs so may not need to define a function here.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/ProcessPreBuiltCohort.ts
```diff
@@ -205,17 +197,13 @@ function translatePreBuiltCohortFilterForDataset(
...
+        .map((categoricalValue) => categorcialValues.indexOf(categoricalValue))
```
romanlutz, line 205: nit: categorical instead of categorcial

romanlutz (approved):

PR #1305, opened 2022-03-31
Limit individual feature importance selection to up to 5

tongyu-microsoft:
This PR adds a info icon with callout to let users know that we limit maximum datapoint selection to 5. User now can only select up to 5 datapoints to view individual feature importances. We also disabled the 'Select all' button for 'Correct predictions' and 'Incorrect predictions'.

image
image

## Description

This PR is to resolve the service throttling issue in Azure ML. Originally if user selects hundreds of data points all at once, it will cause service throttling. By limiting number of datapoints to 5, we won't hit service unavailable issue.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [x] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] erroranalysis
- [ ] raiutils
- [ ] raiwidgets
- [ ] rai_core_flask
- [ ] responsibleai

## Tests

- [ ] No new tests required.

## Documentation:


imatiach-msft (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/IndividualFeatureImportanceView.tsx
```diff
@@ -70,9 +77,22 @@ export class IndividualFeatureImportanceView extends React.Component<
...
     onSelectionChanged: (): void => {
+      const c = this.selection.getSelectedCount();
+      const indices = this.selection.getSelectedIndices();
+      if (c === this.maxSelectable) {
+        this.setState({ selectedIndices: indices });
+      }
+      if (c > this.maxSelectable) {
+        for (const index of indices) {
+          if (!this.state.selectedIndices.includes(index)) {
+            this.setState({ indexToUnselect: index });
```
imatiach-msft, line 92: don't you need to set the selectedIndices still if c > this.maxSelectable? Referring to:
```
this.setState({ selectedIndices: indices });
```

imatiach-msft: Oh, you are basically not allowing the user to select more indices, by unselecting them after they are selected.  I assumed you would just remove the first point that was selected to replace the new point, but that's not what is happening here.  I think this is ok except it's a bit strange to unselect a point that was selected?  The user should just not be able to select the new point I would think?

libs/localization/src/lib/en.json
```diff
@@ -1300,8 +1300,9 @@
       "CorrectPredictions": "Correct predictions",
       "GlobalExplanation": "Aggregate feature importance",
       "IncorrectPredictions": "Incorrect predictions",
-      "IndividualFeature": "Select a datapoint by clicking on a datapoint (or multiple datapoints) in the table to view their local feature importance values (local explanation) and individual conditional expectation (ICE) plot below. For datasets with more than 5000 datapoints, the view is a random subsample to enable easy exploration.",
-      "LocalExplanation": "Individual feature importance"
+      "IndividualFeature": "Select a datapoint by clicking on a datapoint (up to 5 datapoints) in the table to view their local feature importance values (local explanation) and individual conditional expectation (ICE) plot below. For datasets with more than 5000 datapoints, the view is a random subsample to enable easy exploration.",
+      "LocalExplanation": "Individual feature importance",
+      "SelectionLimit": "Up to 5 datapoints can be selected at this time."
```
imatiach-msft, line 1305: We can't make it 10?  Did PM decide on the number 5?

tongyu-microsoft: Yes this number 5 is from PM. We can collect feedback and update the number. But I feel 5 is a good number for now because otherwise feel like ICE plot will contain too many lines
image

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/IndividualFeatureImportanceView.tsx
romanlutz: I faced a similar issue just yesterday (with Dropdown, not DetailsList, but the result is similar) and solved it by showing an error message and not adding the latest selected one.

I do agree with @imatiach-msft that it's kind of an uncommon pattern (or at least one I haven't seen yet in my limited time with TS 🙂 ) to put an index to unselect into the state, rather than just not adding the latest one to the state's selected indices. This would mean that you set that in the state, then set the state again, so a total of two state changes for this and all the associated overhead.

It's entirely possible I'm missing something about this, of course... Just in case I'm not I'll unresolve this. What do you think @tongyu-microsoft ?

tongyu-microsoft: Yeah by selecting and unselecting the 6th datapoint, it is a hack to limit user's selection. 

But for DetailsList component, there is no way for us to block user from selecting the 6th datapoint. `onSelectionChanged` is a callback after user selected the datapoint. 

I also tried the other way where we can set `disableSelectionZone` to false when selection count is 5. But `disableSelectionZone=true` will also block user from unselecting the selected datapoints.

tongyu-microsoft: Here is the only example I found to limit selection count, and this is same as Ilya's idea to unselect a point when select 6th point: 
https://stackoverflow.com/questions/66967516/is-there-a-way-to-limit-the-number-of-selectable-items-in-selection-component-of
Codepen Link: https://codepen.io/sakamati/pen/poRryqa

tongyu-microsoft: Agree with Ilya's point that it is strange to unselect user's selected datapoints.. So I just unselect the latest selected point

tongyu-microsoft: I reached out to Ke to see if there is a better solution. I will update in next PR if we find a solution for it

PR #1308, opened 2022-04-01
update code owner

xuke444:
update code owner


CODEOWNERS
```diff
@@ -2,7 +2,47 @@
 /erroranalysis @gaugup @imatiach-msft
 
 # Responsible AI widgets package
-/raiwidgets @imatiach-msft @xuke444 @zhb000 @vinuthakaranth @RubyZ10 @tongyu-microsoft @romanlutz
+/raiwidgets @imatiach-msft @xuke444 @romanlutz
```
imatiach-msft, line 5: Can we add more folks to raiwidgets?  Just seems like 3 is very few for the largest python package we have.

imatiach-msft (approved):

CODEOWNERS
```diff
@@ -2,7 +2,47 @@
...
 
 # Responsible AI SDK package
-/responsibleai @gaugup @imatiach-msft @riedgar-ms @xuke444 @kicha0
\ No newline at end of file
+/responsibleai @gaugup @imatiach-msft @riedgar-ms @xuke444 @kicha0
+
+# Responsible AI SDK package
+/responsibleai @gaugup @imatiach-msft @riedgar-ms @xuke444 @kicha0
```
gaugup, line 11: There seems to be two lines for responsibleai here. Is that needed?

CODEOWNERS
```diff
@@ -2,7 +2,47 @@
...
+
+# workspace settings
+#/* @xuke444
```
gaugup, line 14: Shall we also add code owners for raiutils?

CODEOWNERS
romanlutz: I don't particularly care about the number tbh. The people who are listed should know what the code does, though. I can only speak for myself, of course, and feel like this PR represents which parts I know enough about to review PRs. In some ways it's better to have tighter control while encouraging more reviews per PR. For example, even if I'm not required for the responsibleai package it's probably still useful if I provide feedback. At the same time, that alone shouldn't be sufficient to merge a PR. So in summary, to me this makes sense.

Regarding size: Is raiwidgets really that large? I feel like a good chunk of it is in other packages (responsible, rai-core-flask, etc.) so it's a fairly thin layer of things. It is the one that's most exposed to users, of course.

Wdyt @imatiach-msft?

romanlutz (approved):

CODEOWNERS
xuke444: my bad

CODEOWNERS
xuke444: I didn't add any there since it is small and I am not clear what it will end up with later yet.

Currently only our group can approve pr, so even without adding any approver there, only 10 of us can review and approve.

CODEOWNERS
imatiach-msft: Hmm maybe responsibleai package is actually the largest in terms of the number of lines of python code.  However, raiwidgets in some sense also contains all of the typescript code, so it is still the largest by total lines of python + typescript code, from that viewpoint.  Regardless it is the main package that depends on all of the others, so it's the most important one for us, and it often has a lot of code changes in PRs.

CODEOWNERS
gaugup: I still the duplicate line for responsibleai. Can you recheck?

CODEOWNERS
xuke444: I think you mix folder with package.
This code owner is based on file structure not pipy package.
raiwidget directory has 0 typescript
it is mainly handle how we plug flask to typescript.
That's why I only put limited amount of people there.
all the typescript ui code splitted into different folders have different approvers.

CODEOWNERS
xuke444: just pushed

gaugup (approved):

PR #1306, opened 2022-04-01
Add a feature flag for the new model overview experience

romanlutz:
## Description

Adding a feature flag and turning it off by default. This will be used to keep the old experience while the new experience is being built so that we can keep changes small and easy to review. Additionally, both versions will be testable using the feature flag.

This PR also removes custom styling that's unnecessary.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/counterfactuals
- [ ] responsibleai/dataset-explorer
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret
- [ ] responsibleai/localization
- [ ] responsibleai/mlchartlib
- [x] responsibleai/model-assessment

Python packages changed:

- [ ] erroranalysis
- [ ] raiutils
- [ ] raiwidgets
- [ ] rai_core_flask
- [ ] responsibleai

## Tests

- [x] No new tests required.
- [ ] New tests for the added feature are part of this PR.
- [ ] I validated the changes manually.

## Screenshots (if appropriate):

No changes to the existing version except for the removal of the custom styling. Previously, the left padding was 39px for the section except for the header which had 40px left padding. This is now uniformly 40px (but hardly recognizable unless you count pixels, of course).

image

## Documentation:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1314, opened 2022-04-04
Add dataset cohort table to new ModelOverview experience

romanlutz:
## Description

Adding a table/heatmap for viewing statistics on the user-defined dataset cohorts. Note that this isn't exposed to users yet due to the feature flag added in #1306. 

The coloring is custom per column/metric. Metrics are configurable. Cohort names are wrapped after 40 characters with a max of 2 lines to avoid overlapping.

Tests will be added separately since the feature isn't exposed yet and to keep the PRs small.

Expect several follow-up PRs to this one in the next days. The `TODO` for multiclass is intentional.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] erroranalysis
- [ ] raiutils
- [ ] raiwidgets
- [ ] rai_core_flask
- [ ] responsibleai

## Tests

Tests to follow in a separate PR as mentioned above.

- [ ] No new tests required.
- [x] I validated the changes manually.

image

and a GIF!

https://user-images.githubusercontent.com/10245648/161852286-690f45bc-de5e-4d4d-8626-1ddd0e193010.mp4

## Documentation:

- [x] My change requires a change to the documentation.

Doc changes will happen once the feature flag is removed.


libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/CohortStatsHeatmap.tsx
```diff
@@ -0,0 +1,145 @@
...
+        .filter((element) =>
+          this.props.selectedMetrics.includes(element.key.toString())
+        )
+        .map((element) => {
+          return element.text;
+        })
+    );
+
+    // generate table contents
+    const cohortLabeledStatistics = generateMetrics(
```
romanlutz, line 49: I'm not excited about using `JointDataset` here but I don't see this elsewhere. Any idea @xuke444 ?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/CohortStatsHeatmap.tsx
```diff
@@ -0,0 +1,147 @@
...
+                return localization.formatString(
+                  localization.ModelAssessment.ModelOverview.tableCountTooltip,
+                  this.series.yAxis.categories[this.point.y],
+                  value
+                );
+              }
+              // Metric columns
+              return localization.formatString(
+                localization.ModelAssessment.ModelOverview.tableMetricTooltip,
+                this.series.xAxis.categories[this.point.x].toLowerCase(),
```
tongyu-microsoft, line 119: why do we only do toLowerCase() for xAxis, not yAxis?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -0,0 +1,195 @@
...
+      if (
+        !item.selected &&
+        this.state.selectedMetrics.includes(item.key.toString())
+      ) {
+        const selectedMetrics = this.state.selectedMetrics;
+        const unselectedMetricIndex = selectedMetrics.findIndex(
+          (key) => key === item.key.toString()
+        );
+        // remove unselected metric
+        selectedMetrics.splice(unselectedMetricIndex, 1);
```
tongyu-microsoft, line 188: I was commenting to see if we want to use .filter()? `const newSelectedMetrics = selectedMetrics.filter(key => key !== item.key.toString())`  
After research online, looks like .splice() wins. So no change needed here :D
https://stackoverflow.com/questions/44435141/remove-object-in-array-using-filter-and-splice-which-one-is-best-approach-in-jav

tongyu-microsoft (approved):

libs/localization/src/lib/en.json
```diff
@@ -1316,6 +1316,25 @@
     },
     "Navigation": {
       "modelStatistics": "Model statistics"
+    },
+    "ModelOverview": {
+      "metricsDropdown": "Metric(s)",
+      "accuracy": "Accuracy",
+      "f1Score": "F1 score",
```
imatiach-msft, line 1323: I already see some of these metrics here:
https://github.com/microsoft/responsible-ai-toolbox/blob/9a2380386828986306f367db3b99408abcdbeede/libs/localization/src/lib/en.json#L256
I wonder is there some central section we can put them under, to reduce duplication of metric names (and any possible inconsistencies)?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/StatsTableUtils.ts
```diff
@@ -0,0 +1,116 @@
...
+  selectableMetrics: IDropdownOption[],
+  labeledStatistics: ILabeledStatistic[][],
+  selectedMetrics: string[]
+) {
+  const items: PointOptionsObject[] = cohorts.map(
+    (errorCohort, cohortIndex) => {
+      return {
+        colorValue: 0,
+        value: errorCohort.cohortStats.totalCohort,
+
```
vinuthakaranth, line 19: nit: remove extra line if not needed

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/CohortStatsHeatmap.tsx
```diff
@@ -0,0 +1,147 @@
...
+            align: "left",
+            text: localization.ModelAssessment.ModelOverview
+              .dataCohortsHeatmapHeader
+          },
+          tooltip: {
+            formatter() {
+              // to avoid semantic error during build cast point to any
+              if (
+                this.point.y === undefined ||
+                (this.point as any).value === undefined ||
```
imatiach-msft, line 102: nit: might this be slightly better?

```
const pointValue = (this.point as any).value
if (
                this.point.y === undefined || pointValue === undefined || pointValue === null
  ) {
...
```

imatiach-msft: actually you already do this below on line 107, so now you can remove that line too
```
const value = (this.point as any).value;
```

vinuthakaranth (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/DatasetCohortStatsTable.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { CohortStatsHeatmap } from "./CohortStatsHeatmap";
+
+interface IDatasetCohortStatsTableProps {
+  selectableMetrics: IDropdownOption[];
+  selectedMetrics: string[];
+}
+
+class IDatasetCohortStatsTableState {}
+
+export class DatasetCohortStatsTable extends React.Component<
```
imatiach-msft, line 20: I'm terrified to mention that there are no tests for this new code, but then if I wrote this class I wouldn't have added them either, so you can just ignore this.  But it would be nice to add some e2e tests and perhaps at some point even we can just add simple unit tests in later PRs.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -0,0 +1,195 @@
...
+  constructor(props: IModelOverviewProps) {
+    super(props);
+    this.state = {
+      selectedMetrics: []
+    };
+  }
+
+  public componentDidMount(): void {
+    let defaultSelectedMetrics: string[] = [];
+    if (this.context.dataset.task_type === "classification") {
```
imatiach-msft, line 43: I think these constants should be moved to some constants file?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/StatsTableUtils.ts
```diff
@@ -0,0 +1,116 @@
...
+  selectableMetrics
+    .filter((element: IDropdownOption) =>
+      selectedMetrics.includes(element.key.toString())
+    )
+    .forEach((metricOption: IDropdownOption, metricIndex: number) => {
+      // first determine min and max values
+      let metricMin = Number.MAX_SAFE_INTEGER;
+      let metricMax = Number.MIN_SAFE_INTEGER;
+      cohorts.forEach((_errorCohort, cohortIndex) => {
+        const stat = labeledStatistics[cohortIndex].find(
```
imatiach-msft, line 35: nit: maybe call it metricStat instead of stat, so instead of stat.stat you will have metricStat.stat to reduce confusion when looking at the code

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/CohortStatsHeatmap.tsx
romanlutz: Great question! It's actually inverted, so x is y and y is x (just to make it a little more confusing). In any case, we put the metric name in lower case so that it renders with proper capitalization in a sentence. The y axis (or rather: x axis) on the other hand is the cohort name which may be capitalized on purpose. Those are decided by the user after all, so I felt like it's safer not to mess with the capitalization. I should really document this with a nice comment. Thanks for your question!

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/StatsTableUtils.ts
```diff
@@ -0,0 +1,116 @@
...
+  const openingParenthesis = " (";
+  const whitespace = " ";
+  const searchString = new Set(
+    label.slice(slicingIndex - maxLineLength / 2, slicingIndex)
+  );
+
+  if (searchString.has(closingParenthesis)) {
+    // option 1: wrap after closing parenthesis
+    slicingIndex =
+      label.indexOf(closingParenthesis, slicingIndex - maxLineLength / 2) + 2;
```
imatiach-msft, line 93: I think this code could be refactored a bit more, maybe like:

```
let axisLabelString = ""
if (searchString.has(closingParenthesis)) {
  axisLabelString = closingParenthesis
} else if (searchString.has(openingParenthesis)) {
  axisLabelString = openingParenthesis
} else if (wrapOnWhitespace && searchString.has(whitespace)) {
  axisLabelString = whitespace
}
if (axisLabelString != "") {
  slicingIndex = label.indexOf(axisLabelString, slicingIndex - maxLineLength / 2) + 1;
  if (axisLabelString == closingParenthesis) {
    slicingIndex += 1
  }
}
```
or maybe move the code:
```
def updateSlicingIndex(slicingIndex...): number => {
  label.indexOf(axisLabelString, slicingIndex - maxLineLength / 2) + 1;
  if (axisLabelString == closingParenthesis) {
    slicingIndex += 1
  }
}
```
into a function and call that in each of the if/else if's above

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/StatsTableUtils.ts
```diff
@@ -0,0 +1,116 @@
...
+  } else if (wrapOnWhitespace && searchString.has(whitespace)) {
+    // option 3: wrap after maxLineLength characters
+    slicingIndex =
+      label.indexOf(whitespace, slicingIndex - maxLineLength / 2) + 1;
+  }
+  label =
+    label.slice(0, slicingIndex) +
+    lineWrapHtmlTag +
+    label.slice(slicingIndex, label.length);
+
```
imatiach-msft, line 107: nit: maybe use a variable here:
```
const maxLabelLength = maxLineLength * maxLines + lineWrapHtmlTag.length
  if (label.length > maxLabelLength) {
    label =
      label.slice(
        0,
        maxLabelLength - ellipsis.length
      ) + ellipsis;
  }
```

imatiach-msft (approved):

libs/localization/src/lib/en.json
romanlutz: I was considering that initially. However, it seems like we don't have a full set somewhere and the strings don't exactly match either. Adding support for the metrics I need everywhere else is work I haven't budgeted for, so I am creating a special set tailored to what I need to create. As we move away from UI-calculated metrics towards SDK calculated metrics this will have to be overhauled anyway. Wdyt?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/DatasetCohortStatsTable.tsx
romanlutz: Certainly! The only reason I even put out the PR is because it's behind a feature flag. I have an item for writing a whole series of e2e tests, but I wanted to keep the PRs small. The number of comments on this PR tells me that making PRs small is a great idea if you want feedback. Love the energy and will keep making small changes. Don't worry, I've made sure to budget for e2e and I will not remove the feature flag unless we have full coverage.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/StatsTableUtils.ts
romanlutz: I wonder where that came from... Thanks!

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/StatsTableUtils.ts
romanlutz: You may have noticed that the index is not just `indexOf(...) + 1` in every case. It depends on the slicing position, too. I personally feel like it's clearer to determine the index first but I agree that there's some redundancy. We had a design meeting today in which we decided to simplify these cohort descriptions a bit, so I'll be able to remove the parenthesis cases altogether. That should make the code simple enough to avoid needing a separate function.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/StatsTableUtils.ts
romanlutz: I like that!

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/StatsTableUtils.ts
romanlutz: Updated this now!

libs/localization/src/lib/en.json
romanlutz: I'll open an item to consolidate these. I think it needs to be addressed, too, but it may be better suited for the big data discussion.

PR #1334, opened 2022-04-11
update erroranalysis to 0.2.1

imatiach-msft:
## Description

- update erroranalysis to 0.2.0 after the recent release to pypi

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] erroranalysis
- [ ] raiutils
- [x] raiwidgets
- [ ] rai_core_flask
- [x] responsibleai

## Tests

- [x] No new tests required.

## Documentation:

### DIFF
--- a/raiwidgets/requirements.txt
+++ b/raiwidgets/requirements.txt
@@ -6,5 +6,5 @@ itsdangerous==2.0.1
 jinja2==2.11.3
 scikit-learn>=0.22.1
 lightgbm>=2.0.11
-erroranalysis>=0.1.31
+erroranalysis>=0.2.1
 fairlearn>=0.7.0
--- a/responsibleai/requirements.txt
+++ b/responsibleai/requirements.txt
@@ -1,7 +1,7 @@
 dice-ml>=0.7.2,<0.8
 econml~=0.12.0
 jsonschema
-erroranalysis>=0.1.31
+erroranalysis>=0.2.1
 interpret-community>=0.25.0
 lightgbm>=2.0.11
 numpy>=1.17.2


tongyu-microsoft (approved):

raiwidgets/requirements.txt
```diff
@@ -1,10 +1,4 @@
-numpy>=1.17.2
```
gaugup, line 1: I see numpy being imported in cohort.py in raiwidgets. Why are we removing this direct dependency from requirements of this package?

gaugup (changes requested):

raiwidgets/requirements.txt
imatiach-msft: raiwidgets depends on responsibleai which depends on numpy already.  Specifying it multiple times is unnecessary duplication.

romanlutz: Hmm if it was a slightly different setup (e.g., the two packages not in the same repo) I would object. In this particular case I guess it's fine since we control both dependency files.

romanlutz (approved):

raiwidgets/requirements.txt
gaugup: I still feel that direct imports should be hard dependencies (we could you unpin these for reducing complexity).

But major concern that I have here is from dev point of view there is very little sense I can make out of being able to import from numpy when it is not a direct dependency. I am not a big fan of implicit dependency as such.

xuke444 (approved):

raiwidgets/requirements.txt
imatiach-msft: done, I changed the code back to keep the duplicate dependencies

gaugup (approved):

PR #1335, opened 2022-04-11
block empty cohort creation in RAI Dashboard

imatiach-msft:
## Description

We should block user from creating empty cohorts by checking if there is no datapoint in the cohort that user is going to create.
If user creates new cohort but has not added any filters yet, we should disable the "Save" and "Save and switch" buttons
If user adds filters that have no datapoints, have a pop up that blocks user when they click "Save" or "Save and switch"

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [x] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] erroranalysis
- [ ] raiutils
- [ ] raiwidgets
- [ ] rai_core_flask
- [ ] responsibleai

## Tests

- [ ] No new tests required.

image

## Documentation:

### DIFF
--- a/apps/dashboard-e2e/src/describer/interpret/cohort/describeCreateCohort.ts
+++ b/apps/dashboard-e2e/src/describer/interpret/cohort/describeCreateCohort.ts
@@ -10,6 +10,10 @@ export function describeCreateCohort(dataShape: IInterpretData): void {
       .get('#cohortEditPanel span:contains("No filters")')
       .should("exist");
   });
+  it("should have save and save and switch buttons disabled by default", () => {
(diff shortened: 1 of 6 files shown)


gaugup:
How do we handle this scenario for the cohorts injected from SDK? Is there some way of checking in python SDK if a user cohort has zero data points?

tongyu-microsoft (approved):

imatiach-msft:
@gaugup "How do we handle this scenario for the cohorts injected from SDK" PM decided that we can punt this question until late May for now, it seems we need a design for this functionality.

PR #1339, opened 2022-04-12
Add warning message in cohort editor for invalid input value; Update 'Shift cohort' to 'Switch cohort'

tongyu-microsoft:
This PR adds warning messages in cohort editor for invalid input values; This PR also updates wording from 'Shift cohort' to 'Switch cohort'.

## Description

1. Add warning messages in cohort editor for invalid input values: There was a bug created regarding to cohort input field, where manually putting in a value with keyboard does not work. The root cause for this bug is, we limit input value to be in between min value and max value if it is a numeric number. We don't allow users to input invalid values. In the case when user inputs an invalid value, we reset the value back to previous input value. This PR adds warning message to users so that users can understand the reason they are not able to input the value is because the input value is invalid.
image
image

2. We used to show "Shift cohort" in some places, while "Switch cohort" in others. This PR updates all the wordings from 'Shift cohort' to 'Switch cohort' to be consistent.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [x] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] erroranalysis
- [ ] raiutils
- [ ] raiwidgets
- [ ] rai_core_flask
- [ ] responsibleai

## Tests

- [ ] No new tests required.

image

## Documentation:

### DIFF
--- a/libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortEditorFilter.tsx
+++ b/libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortEditorFilter.tsx
@@ -21,11 +21,15 @@ import { FilterMethods, IFilter } from "../../Interfaces/IFilter";
 import { FabricStyles } from "../../util/FabricStyles";
 import { IJointMeta, JointDataset } from "../../util/JointDataset";
 
+import { cohortEditorStyles } from "./CohortEditor.styles";
+
 export interface ICohortEditorFilterProps {
   openedFilter: IFilter;
   jointDataset: JointDataset;
   filterIndex?: number;
   filters: IFilter[];
+  showInvalidMinMaxValueError: boolean;
+  showInvalidValueError: boolean;
   setSelectedProperty(
     ev: React.FormEvent<IComboBox>,
     item?: IComboBoxOption
@@ -99,7 +103,7 @@ export class CohortEditorFilter extends React.Component<ICohortEditorFilterProps
     // filterIndex is set when the filter is editing openedFilter and reset to filters.length otherwise
     const isEditingFilter =
       this.props.filterIndex !== this.props.filters.length;
-
+    const styles = cohortEditorStyles();
     let minVal, maxVal;
     if (selectedMeta.treatAsCategorical || !selectedMeta.featureRange) {
       // Numerical values treated as categorical are stored with the values in the column,
@@ -228,6 +232,17 @@ export class CohortEditorFilter extends React.Component<ICohortEditorFilterProps
                       this.props.setNumericValue(0, selectedMeta, 1, value);
                     }}
                   />
+                  {this.props.showInvalidMinMaxValueError &&
+                    selectedMeta.featureRange && (
+                      <p className={styles.invalidValueError}>
+                        {localization.formatString(
+                          localization.Interpret.CohortEditor
+                            .minimumGreaterThanMaximum,
+                          selectedMeta.featureRange.min,
+                          selectedMeta.featureRange.max
+                        )}
+                      </p>
+                    )}
                 </>
               ) : (
                 <SpinButton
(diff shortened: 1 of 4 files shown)


libs/localization/src/lib/en.json
```diff
@@ -82,7 +82,7 @@
       "Unknown": "unknown"
     },
     "ShiftCohort": {
```
imatiach-msft, line 84: should this be updated as well?
```
"ShiftCohort": {
```

libs/localization/src/lib/en.json
```diff
@@ -211,7 +211,7 @@
       "fullscreen": "Fullscreen",
       "heatMap": "Heat map",
       "newCohort": "New cohort",
-      "shiftCohort": "Shift cohort",
+      "shiftCohort": "Switch cohort",
```
imatiach-msft, line 214: do we want to update this key too?
```
"shiftCohort"
```

imatiach-msft (approved):

libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortEditorFilter.tsx
```diff
@@ -228,6 +232,17 @@ export class CohortEditorFilter extends React.Component<ICohortEditorFilterProps
                       this.props.setNumericValue(0, selectedMeta, 1, value);
                     }}
                   />
+                  {this.props.showInvalidMinMaxValueError &&
+                    selectedMeta.featureRange && (
+                      <p className={styles.invalidValueError}>
+                        {localization.formatString(
+                          localization.Interpret.CohortEditor
```
imatiach-msft, line 239: I think this should be Core instead of Interpret now:

```
localization.Interpret.CohortEditor
```
should be:

```
localization.Core.CohortEditor
```

it looks like this wasn't updated properly when it was moved to core.

romanlutz:
@tongyu-microsoft I have no objections, this looks great to me! 
I was just curious if you'd considered using the built-in warnings for dropdowns: https://developer.microsoft.com/en-us/fluentui#/controls/web/dropdown
image

The nice thing about this is that the dropdown itself also goes red which is a nice visual suggestion to the user.

I guess this goes back to whether we are forcing users to change it (error) or whether it's more of a suggestion (warning). I believe it's an error, not just a warning, right?

libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortEditorFilter.tsx
imatiach-msft: I guess, if you prefer, you can create an issue for this... but someone needs to fix it at some point

tongyu-microsoft:
> @tongyu-microsoft I have no objections, this looks great to me! I was just curious if you'd considered using the built-in warnings for dropdowns: https://developer.microsoft.com/en-us/fluentui#/controls/web/dropdown image
> 
> The nice thing about this is that the dropdown itself also goes red which is a nice visual suggestion to the user.
> 
> I guess this goes back to whether we are forcing users to change it (error) or whether it's more of a suggestion (warning). I believe it's an error, not just a warning, right?

Thanks for the comment! Yeah the built-in errors for <Dropdown /> or <TextField /> etc are nice! But here we are using component called <SpinButton /> and I didn't find built-in warning/error properties in SpinButton

libs/localization/src/lib/en.json
tongyu-microsoft: Yeah I was thinking about updating the key as well. but looks like we have components named by ShiftCohort, and there are too many changes if we want to rename all to Switch cohort. Since it doesn't impact user's experience or code readability if we continue using ShiftCohort as key in codespace, I only updated the localization strings here.
image

libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortEditorFilter.tsx
tongyu-microsoft: Looks like there are a few changes needed. Created an issue for tracking purpose: https://github.com/microsoft/responsible-ai-toolbox/issues/1340

PR #1341, opened 2022-04-12
Add disaggregated analysis table to Model Overview

romanlutz:
## Description

The title says it all 🙂 Other than the disaggregated analysis I'm also adding textured color for NaN values and changing the line wrapping code slightly. We now wrap on the last whitespace in a line (or at 40 characters), and all the labels go into a grid (as an extension to the table).

See the video below for a quick demo:

https://user-images.githubusercontent.com/10245648/163058848-1e8c5472-6653-4ad7-846b-2320d72f4a7e.mp4

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [x] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] erroranalysis
- [ ] raiutils
- [ ] raiwidgets
- [ ] rai_core_flask
- [ ] responsibleai

## Tests

Tests will be added before feature flag is removed, in a different PR.

- [ ] No new tests required.

see video above for reference

## Documentation:


tongyu-microsoft (approved):
UI styling question on feature selection dropdown from the demo video: do we want to add scroll bar for dropdown list, as the feature list might be really long? Should we limit Height of the dropdown list? Similar like how we currently do for counterfactual datapoint selection dropdown (attached screenshots below):
image

image

minthigpen:
> @romanlutz UI styling question on feature selection dropdown from the demo video: do we want to add scroll bar for dropdown list, as the feature list might be really long? Should we limit Height of the dropdown list? Similar like how we currently do for counterfactual datapoint selection dropdown (attached screenshots below): image
> 
> image

Yes we want to do a scroll bar and limit the height of the dropdown list. If we can make the drop down experience consistent across the RAI dashboard that would be the best.

romanlutz:
Added this using `FabricStyles` which has the entire configuration:
image

I've applied it to both dropdown (now combo boxes), but the metric selection doesn't have enough elements to require the scrollbar. 🙂

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/StatsTableUtils.ts
```diff
@@ -65,9 +65,24 @@ export function generateCohortsStatsTable(
             y: cohortIndex
           });
         } else {
-          // not a numeric value (NaN), so just put null and use color gray
+          // not a numeric value (NaN), so just put null and use textured color
           items.push({
-            color: "#808080",
+            color: {
```
imatiach-msft, line 70: oh cool!!!

imatiach-msft (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -120,42 +125,47 @@ export class ModelOverview extends React.Component<
...
+      >
         <Text variant="medium" style={{ maxWidth: descriptionMaxWidth }}>
           {localization.Interpret.ModelPerformance.helperText}
         </Text>
         {!this.props.showNewModelOverviewExperience && <OverallMetricChart />}
         {this.props.showNewModelOverviewExperience && (
           <>
             <Stack horizontal tokens={{ childrenGap: "20px" }}>
-              <Dropdown
+              <ComboBox
```
imatiach-msft, line 139: What happens when you choose a ton of features, say 100 features - is the first column still readable in that case?  Does the UI break in any way?

imatiach-msft: Or do we impose some limit on the number of features the user can choose in the dropdown?

romanlutz: This is the metrics dropdown. You can choose as many as you like, but there are only like 7 for binary classification that we currently support. Perhaps more in the future.

For the feature dropdown you can only choose two, at which point the other options get disabled. You can see the video for an example (see PR description).

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/StatsTableUtils.ts
romanlutz: 🙂

PR #1348, opened 2022-04-14
Add charts for new model overview experience

romanlutz:
## Description

This PR adds
- a box plot with one box per cohort
- a bar chart with one bar per cohort
- a cohort configuration flyout
- (more) standardized usage of highcharts box plot code between our various sections

A short demo:

https://user-images.githubusercontent.com/10245648/163516974-c7d05dfb-5f86-428d-a6b2-6d457f838b0e.mp4

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [x] responsibleai/core-ui
- [x] responsibleai/dataset-explorer
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] erroranalysis
- [ ] raiutils
- [ ] raiwidgets
- [ ] rai_core_flask
- [ ] responsibleai

## Tests

Tests will be added before removing the feature flag.

- [ ] No new tests required.

## Documentation:


minthigpen (approved):
Awesome progress :)

libs/core-ui/src/lib/util/calculateBoxData.ts
```diff
@@ -1,44 +1,59 @@
...
+): IHighchartBoxData {
+  data.sort((number1: number, number2: number) => {
+    return number1 - number2;
+  });
+  const firstQuartile = getPercentile(data, 25);
+  const median = getPercentile(data, 50);
+  const thirdQuartile = getPercentile(data, 75);
+  const interquartileRange = thirdQuartile - firstQuartile;
+  // calculate fences as min and max allowed values to be inside the box
+  const lowerFence = firstQuartile - interquartileRange * 1.5;
```
imatiach-msft, line 31: minor nit (move the *1.5 up):
```
const interquartileRange = (thirdQuartile - firstQuartile) * 1.5;
const lowerFence = firstQuartile - interquartileRange;
const upperFence = thirdQuartile + interquartileRange;
```

libs/core-ui/src/lib/util/calculateBoxData.ts
```diff
@@ -1,44 +1,59 @@
...
 
-function getPercentile(data: number[], percentile: number): number {
-  data.sort((a, b) => a - b);
-  const index = (percentile / 100) * data.length;
+function getPercentile(sortedData: number[], percentile: number) {
+  const index = (percentile / 100) * sortedData.length;
+  let result;
   if (Math.floor(index) === index) {
-    return (data[index - 1] + data[index]) / 2;
+    result = (sortedData[index - 1] + sortedData[index]) / 2;
```
imatiach-msft, line 54: if the chart only contained a single point, might this break?  wondering if we need to make sure sortedData length > 1 in this case?  Not really sure, maybe it's not relevant considering the code outside this method.
```
sortedData[index - 1] + sortedData[index]
```

libs/dataset-explorer/src/lib/getDatasetBoxOption.ts
```diff
@@ -25,7 +26,7 @@ export function getDatasetBoxOption(plotlyProps: IPlotlyProperty): any {
         fillColor: FabricStyles.fabricColorPalette[0],
         lineWidth: 35
       },
-      name: "outlier",
+      name: localization.Core.BoxPlot.outlierLabel,
```
imatiach-msft, line 29: It seems this code actually isn't in core but in dataset explorer project, going from the file location "[libs/dataset-explorer/src/lib/getDatasetBoxOption.ts](https://github.com/microsoft/responsible-ai-toolbox/pull/1348/files#diff-35195be7508785732abb8a2cc3115f17a4c9fbb56204bd9b811fbee7e3cbf195)".  I think the localization path should be modified to:

localization.DatasetExplorer.BoxPlot.outlierLabel

imatiach-msft: unless this is used in core as well... but I don't see it in core

libs/core-ui/src/lib/util/calculateBoxData.ts
imatiach-msft: I wonder if there are cases where someone might create a cohort with a single point and this might break

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyout.tsx
```diff
@@ -0,0 +1,213 @@
...
+      >
+        <Stack tokens={{ childrenGap: "10px" }}>
+          <Dropdown
+            label={
+              localization.ModelAssessment.ModelOverview
+                .dataCohortsChartSelectionHeader
+            }
+            multiSelect
+            options={[selectAllOption, ...datasetCohortOptions]}
+            styles={{ dropdown: { width: 250 } }}
```
imatiach-msft, line 119: I wonder if these styles changes need to be in the styles file

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyout.tsx
```diff
@@ -0,0 +1,213 @@
...
+    const selectAllOption = {
+      key: selectAllOptionKey,
+      text: localization.ModelAssessment.ModelOverview.selectAllCohortsOption
+    };
+    const datasetCohortOptions = this.props.datasetCohorts.map(
+      (cohort, index) => {
+        return { key: index.toString(), text: cohort.cohort.name };
+      }
+    );
+    const featureBasedCohortOptions = this.props.featureBasedCohorts.map(
```
imatiach-msft, line 74: I think this map over cohort/index to return key/text could be moved into a very small helper function and re-used above on line 69 as well

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverviewMetricChart.tsx
```diff
@@ -0,0 +1,183 @@
...
+            this.setState({ metricSelectionFlyoutIsVisible: false });
+          }}
+        >
+          <Dropdown
+            label={
+              localization.ModelAssessment.ModelOverview
+                .metricChartDropdownSelectionHeader
+            }
+            options={this.props.selectableMetrics}
+            styles={{ dropdown: { width: 250 } }}
```
imatiach-msft, line 166: move out styles to styles file

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionChart.tsx
```diff
@@ -0,0 +1,244 @@
...
+                  {
+                    localization.ModelAssessment.ModelOverview
+                      .boxPlotPlaceholder
+                  }
+                </Text>
+              </div>
+            )}
+            {!noCohortSelected && (
+              <Stack>
+                <BasicHighChart
```
imatiach-msft, line 143: This file looks quite large already, I wonder if it might help to refactor this "BasicHighChart" chart code out to a separate component.  Maybe it's not necessary now though, but in case more code gets added here I would recommend that.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionChart.tsx
```diff
@@ -0,0 +1,244 @@
...
+                      bar: {
+                        dataLabels: {
+                          enabled: true
+                        }
+                      }
+                    },
+                    series: [
+                      {
+                        data: boxPlotData,
+                        fillColor: "#b2d6f2",
```
imatiach-msft, line 162: hmm, can we replace the color with the fluentui constants, maybe cyanblue10 is close: https://developer.microsoft.com/en-us/fluentui#/styles/web/colors/personas

or themeTertiary:
https://developer.microsoft.com/en-us/fluentui#/styles/web/colors/theme-slots

imatiach-msft: image

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyout.tsx
imatiach-msft: I wonder if we need some sort of common CohortUtils file in core which would have such simple shared utilities for cohorts

imatiach-msft (approved):

vinuthakaranth:
Other axis panels have option to save/cancel the change. Here it is direct change. Will this cause consistency change?
other axis panel:
image

libs/core-ui/src/lib/util/calculateBoxData.ts
romanlutz: Addressed locally. Will be pushed shortly :-)

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionChart.tsx
xuke444: We should stay with sematic colors for all non-chart related stuff
https://msdata.visualstudio.com/Vienna/_wiki/wikis/Workspace%20Portal%20Wiki/4919/Theming

For charts I remember there was a predefined color palette somewhere in our code.

romanlutz: I'll check those, but this is the color that was given to me by Jingya. @minthigpen wdyt?

romanlutz:
> Other axis panels have option to save/cancel the change. Here it is direct change. Will this cause consistency change? other axis panel: image

Good point @vinuthakaranth ! I'm addressing this right now. There are a couple things that need fixing as it turns out. One is the confirm/cancel buttons instead of taking immediate effect, the other is radio buttons instead of dropdowns for some of the options.

libs/dataset-explorer/src/lib/getDatasetBoxOption.ts
romanlutz: It seemed like one of the few occasions where localization text is actually reusable, but sure, we can make it custom.

libs/core-ui/src/lib/util/calculateBoxData.ts
romanlutz: I'm writing unit tests for this now :-)

romanlutz: FWIW if there's only 1 then you go into the other case (`else`).

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyout.tsx
romanlutz: It seems to be fairly custom. If something like this keeps reoccurring we should do it, though.

romanlutz: I incorporated the helper, of course.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionChart.tsx
romanlutz: I will have another PR to add a different version of this chart (using lines rather than boxes). At that point I'll split it up because we'll certainly cross the 300 lines. For this PR I'd prefer to keep it as is because I don't quite know yet which parts will be common.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionChart.tsx
romanlutz: It is indeed the same color as cyanBlue10 from SharedColors of fluentui, but I'm getting errors using both fluentui and office-ui-fabric-react in the same file. I'll leave it as is until we switch to fluentui for everything. There are many such occurrences in the code that we should clean up at that point.

romanlutz:
Made all the updates as requested (with the exceptions noted in the corresponding threads) and here's a new video in case you're interested:

https://user-images.githubusercontent.com/10245648/163875446-a353d966-1242-479b-a4c0-6d2b08e2a4f8.mp4

romanlutz:
I'm going to merge this since I don't want this work to get delayed. But if there are comments I'll send a follow-up PR! @vinuthakaranth @zhb000

PR #1367, opened 2022-04-21
Add a summary for disaggregated analysis showing ratio and difference

romanlutz:
## Description

This PR adds another table with difference and ratio for all metrics. This includes a tooltip. Unlike the previous two tables this is not a heatmap with colored backgrounds. Instead, all fields are transparent.

In addition to the summary table, this PR also
- puts all y-axis labels in a grid (as an extension of the table) and aligns all those label columns so that there are no misaligned columns for the same metrics anymore.
- adds N/A as a shown value for undefined metrics (when, for example, FPR can't be computed due to lack of negatives in the data)
- fixes a bug in the disaggregated analysis code where the wrong cohorts were used to calculate the metrics

Tests will be added before the feature flag is removed (separate PR)

https://user-images.githubusercontent.com/10245648/164485416-83f10e83-dd0f-4892-a565-7fed66536ad9.mp4

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret

Python packages changed:

- [ ] erroranalysis
- [ ] raiutils
- [ ] raiwidgets
- [ ] rai_core_flask
- [ ] responsibleai

## Tests

- [ ] No new tests required.

## Documentation:


tongyu-microsoft (approved):

tongyu-microsoft:
From UI/user point of view, this tooltip might be too long to read. Could we limit the width?
image

romanlutz:
> From UI/user point of view, this tooltip might be too long to read. Could we limit the width? image

Excellent point! Let me see if they have some configuration option for that. Otherwise I'll try to hack it.

.eslintrc/.eslintrc.comments.eslintrc
```diff
@@ -2,6 +2,9 @@
   "plugins": ["eslint-plugin-eslint-comments"],
   "extends": ["plugin:eslint-comments/recommended"],
   "rules": {
-    "eslint-comments/no-use": ["error", { "allow": [] }]
+    "eslint-comments/no-use": [
+      "error",
+      { "allow": ["eslint-disable-next-line"] }
```
xuke444, line 7: we intend to disable this for people not abuse this inline disable.

Do you mind create a variable for null in high chart folder and disable rule in eslintrc under that package folder?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/CohortStatsHeatmap.tsx
```diff
@@ -115,7 +96,7 @@ export class CohortStatsHeatmap extends React.Component<
                 // make metric name lower case in sentence
                 this.series.xAxis.categories[this.point.x].toLowerCase(),
                 this.series.yAxis.categories[this.point.y],
-                pointValue
+                pointValue === null ? "N/A" : pointValue
```
xuke444, line 99: Is this visiable to user? Probably need localize

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/CohortStatsHeatmap.tsx
```diff
@@ -132,11 +113,11 @@ export class CohortStatsHeatmap extends React.Component<
               columns: [
                 {
                   labels: {
-                    // format labels to cap the line length
                     formatter() {
-                      return wrapYAxisLabels(this.value.toString(), true);
+                      const text = wrapYAxisLabels(this.value.toString());
+                      return `<div style='width:300px'>${text}</div>`;
```
xuke444, line 118: Does this formatter only take html string? not react element?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/FairnessMetricTable.tsx
```diff
@@ -0,0 +1,171 @@
...
+                .fairnessMetricDifference,
+              localization.ModelAssessment.ModelOverview.fairnessMetricRatio
+            ],
+            grid: {
+              borderWidth: 2,
+              columns: [
+                {
+                  labels: {
+                    formatter() {
+                      return `<div style='width:300px'><p>${this.value}</p></div>`;
```
imatiach-msft, line 156: nit: should style be moved into styles file here?
```
style='width:300px'
```

imatiach-msft (approved):

romanlutz:
@tongyu-microsoft new tooltip with wrapped text! I decided to make the cohort names bold to help with readability.

Tooltip for count column (special case because the values are integers rather than floating point numbers):
image

Tooltip for metric columns:
image

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/FairnessMetricTable.tsx
romanlutz: In general, I agree 100%.

In this particular case, you may have noticed that it's actually a string, not your usual React JSX. I briefly tried doing this with a style file, but it didn't work. I suspect that it has to do with this just being a string.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/CohortStatsHeatmap.tsx
romanlutz: Correct. You have two choices:
- a function (which is what I used) that returns a string
- just a format string (so not even a function) in a slightly differently named parameter

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/CohortStatsHeatmap.tsx
romanlutz: Done

.eslintrc/.eslintrc.comments.eslintrc
romanlutz: Good idea! Will do.

vinuthakaranth (approved):

PR #1375, opened 2022-04-27
Add a top-level pivot to model overview

romanlutz:
## Description

Through user studies we got the feedback that it's easier to work with the new model overview experience if we show EITHER the dataset cohort heatmap OR the disaggregated analysis. Currently, we're showing both at the same time (behind feature flag, of course). This PR changes this by adding a pivot for this purpose. Additionally, the chart below the heatmap only shows EITHER the dataset cohorts OR the feature-based cohorts, but no longer both (as it was before this PR). It is (by design) still possible to show either type below the heatmap if the user so chooses via the cohort configuration flyout.

This change simplifies the chart code considerably by removing code to merge all the cohorts together into a single data structure. Instead, the chart code is now cohort type-agnostic 🙂 

The `ChartConfigurationFlyout` code got significantly more complicated, though, since we have to be careful to check in `componentDidUpdate` whether the props changed. This is caused by the fact that we don't make changes take effect immediately, but rather need to save state until the user clicks the "Confirm" button. You'll find lots of `newly...` variables in the code for the yet-to-be-confirmed choices.

https://user-images.githubusercontent.com/10245648/165417437-b52b09cc-da1b-418c-9846-3cc7b7688c60.mp4

## Checklist

Extensive tests to be added next week, and definitely before removing feature flag.

### DIFF
--- a/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyout.tsx
+++ b/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyout.tsx
@@ -57,13 +57,13 @@ export class ChartConfigurationFlyout extends React.Component<
   constructor(props: IChartConfigurationFlyoutProps) {
     super(props);
     this.state = {
+      datasetCohortViewIsNewlySelected: this.props.datasetCohortViewIsSelected,
       newlySelectedDatasetCohorts:
         this.props.selectedDatasetCohorts ??
         this.props.datasetCohorts.map((_, index) => index),
       newlySelectedFeatureBasedCohorts:
         this.props.selectedFeatureBasedCohorts ??
-        this.props.featureBasedCohorts.map((_, index) => index),
-      datasetCohortViewIsNewlySelected: this.props.datasetCohortViewIsSelected
+        this.props.featureBasedCohorts.map((_, index) => index)
     };
   }
 
@@ -109,16 +109,16 @@ export class ChartConfigurationFlyout extends React.Component<
     if (
       this.state.datasetCohortViewIsNewlySelected !==
         datasetCohortViewIsSelected ||
-      (this.state.newlySelectedFeatureBasedCohorts.length !==
+      this.state.newlySelectedFeatureBasedCohorts.length !==
         newlySelectedFeatureBasedCohorts.length ||
-        this.state.newlySelectedFeatureBasedCohorts.some(
-          (num, index) => num !== newlySelectedFeatureBasedCohorts[index]
-        ))
+      this.state.newlySelectedFeatureBasedCohorts.some(
+        (num, index) => num !== newlySelectedFeatureBasedCohorts[index]
+      )
     )
       this.setState({
         ...this.state,
         datasetCohortViewIsNewlySelected: datasetCohortViewIsSelected,
-        newlySelectedFeatureBasedCohorts: newlySelectedFeatureBasedCohorts
+        newlySelectedFeatureBasedCohorts
       });
   }
 
@@ -169,10 +169,10 @@ export class ChartConfigurationFlyout extends React.Component<
           .dataCohortsChartSelectionHeader
       },
       {
+        disabled: this.props.featureBasedCohorts.length === 0,
         key: this.featureBasecCohortsChoiceGroupOption,
         text: localization.ModelAssessment.ModelOverview
-          .featureBasedCohortsChartSelectionHeader,
-        disabled: this.props.featureBasedCohorts.length === 0
+          .featureBasedCohortsChartSelectionHeader
       }
     ];
 
--- a/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyout.tsx
+++ b/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyout.tsx
@@ -52,7 +52,7 @@ export class ChartConfigurationFlyout extends React.Component<
   public context: React.ContextType<typeof ModelAssessmentContext> =
     defaultModelAssessmentContext;
   private datasetCohortsChoiceGroupOption = "datasetCohorts";
-  private featureBasecCohortsChoiceGroupOption = "featureBasedCohorts";
+  private featureBasedCohortsChoiceGroupOption = "featureBasedCohorts";
 
   constructor(props: IChartConfigurationFlyoutProps) {
     super(props);
@@ -170,7 +170,7 @@ export class ChartConfigurationFlyout extends React.Component<
       },
       {
         disabled: this.props.featureBasedCohorts.length === 0,
-        key: this.featureBasecCohortsChoiceGroupOption,
+        key: this.featureBasedCohortsChoiceGroupOption,
         text: localization.ModelAssessment.ModelOverview
           .featureBasedCohortsChartSelectionHeader
       }
@@ -189,7 +189,7 @@ export class ChartConfigurationFlyout extends React.Component<
             selectedKey={
               this.state.datasetCohortViewIsNewlySelected
                 ? this.datasetCohortsChoiceGroupOption
-                : this.featureBasecCohortsChoiceGroupOption
+                : this.featureBasedCohortsChoiceGroupOption
             }
           />
           <Dropdown
@@ -266,7 +266,7 @@ export class ChartConfigurationFlyout extends React.Component<
       if (option.key === this.datasetCohortsChoiceGroupOption) {
         this.setState({ datasetCohortViewIsNewlySelected: true });
       }
-      if (option.key === this.featureBasecCohortsChoiceGroupOption) {
+      if (option.key === this.featureBasedCohortsChoiceGroupOption) {
         this.setState({ datasetCohortViewIsNewlySelected: false });
       }
     }
--- a/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyout.tsx
+++ b/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyout.tsx
@@ -9,11 +9,13 @@ import {
 import { localization } from "@responsible-ai/localization";
 import {
   Dropdown,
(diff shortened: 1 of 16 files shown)


libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyout.tsx
```diff
@@ -113,13 +162,36 @@ export class ChartConfigurationFlyout extends React.Component<
...
         onDismiss={this.props.onDismissFlyout}
       >
         <Stack tokens={{ childrenGap: "10px" }}>
+          <ChoiceGroup
+            options={choiceGroupOptions}
+            onChange={this.onChoiceGroupChange}
+            selectedKey={
+              this.state.datasetCohortViewIsNewlySelected
+                ? this.datasetCohortsChoiceGroupOption
+                : this.featureBasecCohortsChoiceGroupOption
```
imatiach-msft, line 192: nit: featureBasecCohortsChoiceGroupOption -> did you mean Based instead of Basec?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyout.tsx
```diff
@@ -183,16 +258,25 @@ export class ChartConfigurationFlyout extends React.Component<
...
-  //     return index;
-  //   });
-  // }
+  private onChoiceGroupChange = (
+    _ev?: React.FormEvent<HTMLElement | HTMLInputElement> | undefined,
+    option?: IChoiceGroupOption | undefined
+  ): void => {
+    if (option) {
+      if (option.key === this.datasetCohortsChoiceGroupOption) {
+        this.setState({ datasetCohortViewIsNewlySelected: true });
```
imatiach-msft, line 267: datasetCohortViewIsNewlySelected is so long and a little confusing.  Why not just use an enum here?  Or perhaps call it:
datasetCohortViewSelected

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -167,6 +175,26 @@ export class ModelOverview extends React.Component<
...
+    let labeledStatistics: ILabeledStatistic[][];
+    if (this.state.datasetCohortChartIsVisible) {
+      chartCohorts = this.context.errorCohorts;
+      someCohortSelected =
+        this.state.selectedDatasetCohorts !== undefined &&
+        this.state.selectedDatasetCohorts.length > 0;
+      selectedChartCohorts = this.state.selectedDatasetCohorts ?? [];
+      labeledStatistics = datasetCohortLabeledStatistics;
+    } else {
+      chartCohorts = featureBasedCohorts;
```
imatiach-msft, line 190: nit: the if/else look similar, maybe refactor to a small method with different parameters?

imatiach-msft (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyout.tsx
romanlutz: Because there are two:
- you can pass it into this component (via props)
- you can change the selection in this component (hence "newly" selected) but it won't take effect until you hit "confirm" so we have to store that on the state. That complicates things considerably but it's the only flow that's consistent with the UX we have elsewhere with flyouts.

Enum isn't needed since there are only 2 views and there will only ever be 2.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
romanlutz: Yes, they're similar because I'm setting the same parameters that are needed for rendering. There's no logic that happens in the if, just setting of the parameter, so I don't quite know what a method would help with there. I could create a new component (not sure what I could call it since it is basically the ModelOverview) that takes in the view-agnostic params we're setting in this if/else, but then this component would just consist of this if/else and call the other component which doesn't seem that helpful. I will definitely look into breaking it up further once it's finished, but while there are still lots of pieces moving around I've found it to be easier to handle if it's not broken up into too many components.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionChart.tsx
```diff
@@ -72,19 +70,9 @@ export class ProbabilityDistributionChart extends React.Component<
...
-    );
-    const selectedFeatureBasedCohorts = this.props.featureBasedCohorts.filter(
-      (_cohort, index) => {
-        return this.props.selectedFeatureBasedCohorts.includes(index);
-      }
-    );
-    const selectedCohorts = selectedDatasetCohorts.concat(
-      selectedFeatureBasedCohorts
-    );
+    const selectedCohorts = this.props.cohorts.filter((_cohort, index) => {
```
xuke444, line 73: This is o(n^2) ^_^
Seems fine since # of cohort is limited.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -59,11 +63,15 @@ export class ModelOverview extends React.Component<
   public context: React.ContextType<typeof ModelAssessmentContext> =
     defaultModelAssessmentContext;
   private featureComboBoxRef = React.createRef<IComboBox>();
+  private datasetCohortViewPivotKey = "datasetCohortView";
```
xuke444, line 66: const

xuke444, line 66: this is for default value, correct?
Better to make it "static const", or put const outside of class to save few letters when referencing.
If changeable, it is better to use state.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -167,6 +175,26 @@ export class ModelOverview extends React.Component<
...
+      selectedChartCohorts = this.state.selectedDatasetCohorts ?? [];
+      labeledStatistics = datasetCohortLabeledStatistics;
+    } else {
+      chartCohorts = featureBasedCohorts;
+      someCohortSelected =
+        this.state.selectedFeatureBasedCohorts !== undefined &&
+        this.state.selectedFeatureBasedCohorts.length > 0;
+      selectedChartCohorts = this.state.selectedFeatureBasedCohorts ?? [];
+      labeledStatistics = featureBasedCohortLabeledStatistics;
+    }
```
xuke444, line 196: Seems a lot logic in the render.
This may slow the component rendering, if any of them is time consuming.
For long run process, it is better to save to state and use componentDidUpdate to check and update.

xuke444 (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
romanlutz: static const sounds good!

romanlutz:
Can't seem to comment directly on some of your comments, @xuke444, but I'll try using screenshots:
image

Indeed! We don't anticipate anyone having more than a dozen cohorts. Even if it was 100 I'm not worried. The alternative was to  duplicate all the cohorts in the state, but since cohorts can be changed I felt like it's easier to just keep track of the index.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
romanlutz: Yes! There are a few tiny things that don't matter much performance-wise, but the big pieces are 
- generating the cohorts from the selected features (selected via dropdown)
- generating the metrics for those cohorts and the existing user-created dataset cohorts.

I don't anticipate this component re-rendering at all, though. You do an initial render when showing it the first time. When changing between features you necessarily have to recalculate the metrics, of course. There's no way to precalculate them because there are simply too many possible combinations. After that, I don't think it should be re-rendering with the same cohorts at all. Let me know if I'm missing something!

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionChart.tsx
romanlutz: see my response below https://github.com/microsoft/responsible-ai-toolbox/pull/1375#issuecomment-1111512174

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
romanlutz: Sorry, I meant outside the class. It doesn't change so const works fine.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
romanlutz: closing this since I'll do more refactoring later on (out of scope for this PR since the overall feature isn't finished)

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyout.tsx
romanlutz: I hope this was sufficient as an explanation. I don't think there's a particular action item to take so I'll close it. If not please reopen or message me and I'll address in a future PR. 🙂

PR #1381, opened 2022-05-02
Add feature and metric configuration flyouts to model overview

romanlutz:
## Description

Until now, features and metrics were only selectable via a dropdown (well, technically a `ComboBox`) in the new model overview experience. Since users may want to know more information about the metrics they're choosing, and about the features they're selecting for the disaggregated analysis (including which groups are part of the analysis), the new configuration flyouts added in this PR bridge this gap. Additionally, so far continuous features were split into 3 evenly sized (by interval length) bins rather than allowing users to adjust it. This PR allows configuring the number of bins for continuous features with a default of 3 bins, a minimum of 2, and a maximum of 10. Less than 2 doesn't make sense since there's no split at that point, and more than 10 would get out of hand for the resulting table and chart.

Metric Configuration Flyout:

https://user-images.githubusercontent.com/10245648/166174435-a86dd1d6-368f-42d6-bb20-cc835dacc697.mp4

Feature Configuration Flyout:

https://user-images.githubusercontent.com/10245648/166174404-16d7bbd5-5e39-4533-adf4-05e53450f13f.mp4

Other changes:
- Reorganized the model overview section of `en.json` a bit to make it more intuitive.
- Refactored logic in `DisaggregatedAnalysisUtils.ts` to be reusable for other purposes. The result of this is the new function `generateFeatureBasedFilters`.

Caveats: 
- I'm not quite happy with the complexity of the flyout components. This is largely due to the uncommon way `Selection` works within `DetailsList`. There's no reasonable way to just update the selection rather than resetting it every time. We do this in other places in our codebase, too, and at least here we know that there are relatively reasonable bounds on the size of the tables.
- The new model overview has become large enough to become its own component IMO. I'll do that in a separate PR unless people have objections (?) @microsoft/responsible-ai-widgets 

## Checklist

Note that the feature flag is still on and none of this is visible to users at this point. Before it becomes visible we have to add tests!


libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/FeatureConfigurationFlyout.tsx
```diff
@@ -0,0 +1,334 @@
...
+        name: localization.ModelAssessment.ModelOverview.featureConfiguration
+          .groupsColumnHeader,
+        onRender: this.renderGroups
+      }
+    ];
+
+    return (
+      <Panel
+        isOpen={this.props.isOpen}
+        closeButtonAriaLabel="Close"
```
xuke444, line 131: arial labels need localize as well.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -337,6 +373,26 @@ export class ModelOverview extends React.Component<
                 this.state.datasetCohortChartIsVisible
               }
             />
+            <FeatureConfigurationFlyout
+              isOpen={this.state.featureConfigurationIsVisible}
+              onDismissFlyout={() => {
```
xuke444, line 378: better not to use in-line funciton

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -443,24 +524,38 @@ export class ModelOverview extends React.Component<
         // remove unselected feature
         newlySelectedFeatures.splice(unselectedFeatureIndex, 1);
       }
+      const numberOfContinuousFeatureBins: {
```
xuke444, line 527: this looks like a lodash.mapValue should do

xuke444 (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/FeatureConfigurationFlyout.tsx
romanlutz: Oh! We're not doing that anywhere (hundreds of places in the codebase). How about I open an item for this because we'll have to do that everywhere in the code. Better to take care of it in one big change that doesn't include other things.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
romanlutz: I'm not sure I follow. Do you mean the data structure? lodash's mapValues is a function, isn't it?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/FeatureConfigurationFlyout.tsx
romanlutz: Done

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
romanlutz: Will address this separately due to needing this PR merged urgently.

PR #1383, opened 2022-05-02
Create e2e package for tests shared with AML 

vinuthakaranth:
Create sharable e2e package for tests shared with AML

## Description

Creating a sharable e2e folder under libs to use these set of tests in AML.
All the describe files are moved under libs/e2e folder.
Visit to notebook will only be executed when name parameter is send while calling the describe else all the tests will be run directly (case of AML as visit is done prior to calling tests in describe).
datashape should be sent as the arguments to describe.

## Checklist


libs/e2e/package.json
```diff
@@ -0,0 +1,5 @@
+{
+    "name": "@responsible-ai/e2e",
+    "version": "0.0.1",
```
xuke444, line 3: let's make this 0.0.0 to match other packages

xuke444 (approved):

libs/e2e/package.json
vinuthakaranth: Sure, updated to 0.0.0.

tongyu-microsoft (approved):

PR #1385, opened 2022-05-04
Provide feature flight support to dashboard, test environment, and Python widget

romanlutz:
## Description

This PR adds support for feature flights in general, and for the new model overview experience in particular as a first feature flight. This includes
- processing feature flights as strings (so all of them together come as a list of strings) starting as props to the model assessment dashboard
- parsing code to process flights passed as a single string with a separator `&` inspired by web URLs, but we can definitely change that! Please leave suggestions below if any.
- passing `feature_flights` as a kwarg from the widget to the TS code. This is not advertised in the API since flighted features may not be stable yet.
- support for selecting feature flights (and multiple at the same time!) in the test env. This required a little extra handling for the special case where there are no flights (`"none"`) since we generate a URL for the test env based on the changeable variables like version, language, etc. and now feature flights.

image

This PR does not yet allow for testing the flighted dashboard in pipelines. This will be achieved in a future PR by processing the notebook python code in a way that the flights are added as kwargs to widget constructor calls.

## Checklist


apps/dashboard/src/app/AppHeader.tsx
```diff
@@ -120,4 +154,43 @@ export class AppHeader extends React.Component<IAppHeaderProps> {
...
+            newFlights.length === 0
+              ? "none"
+              : newFlights.join(featureFlightSeparator)
+          );
+        }
+      } else if (!this.props.featureFlights.includes(item.data)) {
+        // need to add flight
+        this.props.onSettingChanged(
+          field,
+          this.props.featureFlights === "none"
```
imatiach-msft, line 185: maybe move this repeating string to a constant?

apps/dashboard/src/app/AppHeader.tsx
```diff
@@ -120,4 +154,43 @@ export class AppHeader extends React.Component<IAppHeaderProps> {
...
+    if (item?.data) {
+      // item.checked means it was checked before the click
+      if (item.checked) {
+        if (this.props.featureFlights.includes(item.data)) {
+          // need to remove flight
+          const previousFlights = parseFeatureFlights(
+            this.props.featureFlights
+          );
+          const newFlights = previousFlights.filter(
+            (flight) => flight !== item.data
```
imatiach-msft, line 172: can you do _  != item.data here (?)

imatiach-msft (approved):

raiwidgets/raiwidgets/responsibleai_dashboard.py
```diff
@@ -30,7 +30,7 @@ class ResponsibleAIDashboard(Dashboard):
     """
     def __init__(self, analysis: RAIInsights,
```
imatiach-msft, line 31: can you add some tests for this parameter in the python API?

raiwidgets/raiwidgets/dashboard.py
```diff
@@ -74,7 +75,8 @@ def __init__(self, *,
             'id': self.id,
             'baseUrl': self._service.env.base_url,
             'withCredentials': self._service.with_credentials,
-            'locale': locale
+            'locale': locale,
+            'featureFlights': kwargs.get('feature_flights')
```
imatiach-msft, line 79: do you need to validate the type of the parameter perhaps before passing it to the UX?

imatiach-msft: especially since this is set by the user which can be error-prone

libs/core-ui/src/lib/Context/ModelAssessmentContext.tsx
```diff
@@ -25,13 +25,16 @@ export interface IModelAssessmentContext {
...
+  featureFlights?: string[];
   errorCohorts: ErrorCohort[];
   readonly baseErrorCohort: ErrorCohort;
   readonly selectedErrorCohort: ErrorCohort;
+
+  // jointDataset and modelMetadata should eventually be removed.
+  // Instead, dataset and modelExplanationData should suffice.
   jointDataset: JointDataset;
   modelMetadata: IExplanationModelMetadata;
+
```
vinuthakaranth, line 37: nit: was this intentionally added?

vinuthakaranth (approved):
LGTM! This is such a good functionality to have and view flights is local dashboard!

raiwidgets/raiwidgets/dashboard.py
imatiach-msft: wait, is it a string, or a list of strings?  I thought it was a list of strings?  And we can also validate the strings are expected perhaps?

apps/dashboard/src/app/AppHeader.tsx
imatiach-msft: oh I thought instead of (flight) => flight you can just use underscore.  Maybe I'm getting confused with scala here :sweat_smile:

raiwidgets/raiwidgets/dashboard.py
imatiach-msft: how does this work with multiple feature flights?  can you pass multiple flights in one parameter?

imatiach-msft: oh I see, you do a split on the separator, which is an ampersand:
```
return featureFlights.split(featureFlightSeparator)
```
might it not be easier to specify all flights as a list of strings in the python api?  Or is this getting too particular about the user interface which we don't expect users to use anyway?

imatiach-msft: I guess you can add a test to check that it works with a string, and add a validation method to check that it indeed is a string (and not, say, a list) and that it doesn't have weird spaces/characters etc, other than ampersands?

apps/dashboard/src/app/App.tsx
```diff
@@ -4,6 +4,7 @@
 import { ITheme } from "@fluentui/react";
 import { generateRoute } from "@responsible-ai/core-ui";
 import { Language } from "@responsible-ai/localization";
```
imatiach-msft, line 6: do you think it might be useful to add a small blurb about this in the contributing.md file:
https://github.com/microsoft/responsible-ai-toolbox/blob/main/CONTRIBUTING.md
to tell other developers how they could leverage this flight flag?

apps/dashboard/src/app/AppHeader.tsx
```diff
@@ -30,6 +35,10 @@ export class AppHeader extends React.Component<IAppHeaderProps> {
   private readonly onDatasetSelect = this.onSelect.bind(this, "dataset");
   private readonly onThemeSelect = this.onSelect.bind(this, "theme");
   private readonly onLanguageSelect = this.onSelect.bind(this, "language");
+  private readonly onFeatureFlightSelect = this.onFlightSelect.bind(
```
xuke444, line 38: this was lagacy code, bind is not good practice here, we are moving to arrow function, do you mind update all these to be arrow functions?

apps/dashboard/src/app/App.tsx
```diff
@@ -121,6 +122,11 @@ export class App extends React.Component<IAppSetting, IAppState> {
                   this.state.version
                 ]
               }
+              featureFlights={
+                this.state.featureFlights === "none"
```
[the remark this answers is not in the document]
romanlutz: It can't be because we're not allowing it to be. I can look into whether I can change that, but I'd have to change it for all the other app settings as well.

apps/dashboard/src/app/AppHeader.tsx
romanlutz: I remember you telling me that :-) I was surprised to see it here, so I didn't want to do something different, but let me change this!

apps/dashboard/src/app/App.tsx
romanlutz: I think that's a good idea. This knowledge shouldn't just get lost.

apps/dashboard/src/app/AppHeader.tsx
romanlutz: Underscore sort of indicates it's an arg that's not being used. Or lodash I guess?

I don't see an issue with this line of code as is.

libs/core-ui/src/lib/Context/ModelAssessmentContext.tsx
romanlutz: Yes, actually! I want to keep the two above separate from the rest since they should eventually be removed :-)

raiwidgets/raiwidgets/dashboard.py
romanlutz: Correct, it's not expected to be used by anyone other than us. I initially wanted it to be a list of strings to be easy to use. However, it's easiest to use IAppSetting in the test env and that is just strings. Since we pass flights as a string in Studio anyway I figured this isn't the worst solution. If someone gets it wrong that's really on them (or whoever gave them that flight string). I'll write the contributor guide section as you suggested, that should be sufficient. I'll validate it's a string as well.

apps/dashboard/src/app/App.tsx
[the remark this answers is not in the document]
romanlutz: Here's the problem: we generate the URL for the test env based on the input parameters. If we allow empty args like this then it fails in 
```
export function generateRoute(params: readonly string[]): string {
  return params.map((p) => `/:${p}?`).join("");
}
```
because we're expecting at least one character.

That's why I introduced "none" as a special case (in the test env only, this doesn't exist in the actual app). Even if we were to figure out how to allow for empty (or undefined) flight params we'd still need to handle this case when showing the currently selected flights in the multi-select dropdown. Right now it can say "none" and it's intuitive. Otherwise we'd have to handle this in a different way. Anyway, I guess this is my argument for keeping it the way it is 🙂

apps/dashboard/src/app/AppHeader.tsx
romanlutz: Resolving this for now, but feel free to ping me directly if you have further thoughts.

raiwidgets/raiwidgets/responsibleai_dashboard.py
romanlutz: done

apps/dashboard/src/app/App.tsx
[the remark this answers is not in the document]
xuke444: So even empty does not work, does it?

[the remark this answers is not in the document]
romanlutz: No because `generateRoute` doesn't allow that.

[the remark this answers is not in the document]
romanlutz: It's this particular route generation that won't work because you need characters:
image

Hence "none". Only in the test env, though, and it's just for the special case where we have no flights. Otherwise it uses the flight names.

PR #1386, opened 2022-05-04
fix race condition in tree view update code which can cause root stats to be ignored on selected cohort in RAI dashboard static view

imatiach-msft:
## Description

Fix race condition in tree view update code which can cause root stats to be ignored on selected cohort in RAI dashboard static view.
There is a race condition where the tree view may not be updated correctly because the setState is not chained- basically, this.state.root would be undefined sometimes because the state hasn't been updated yet.  This can cause the RAI dashboard to show the wrong metric in regression scenario when using the static view.

## Checklist

### DIFF
--- a/apps/dashboard/src/model-assessment/App.tsx
+++ b/apps/dashboard/src/model-assessment/App.tsx
@@ -21,7 +21,9 @@ import {
   generateJsonTreeAdultCensusIncome,
   generateJsonTreeWine,
   getJsonMatrix,
-  getJsonTreeAdultCensusIncome
+  getJsonTreeAdultCensusIncome,
+  getJsonTreeBoston,
+  getJsonTreeWine
 } from "../error-analysis/utils";
 
 interface IAppProps extends IModelAssessmentData {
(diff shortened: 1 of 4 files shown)


apps/dashboard/src/model-assessment/App.tsx
```diff
@@ -89,9 +91,16 @@ export class App extends React.Component<IAppProps> {
           );
       }
     } else {
-      const staticTree = getJsonTreeAdultCensusIncome(
+      let staticTree = getJsonTreeAdultCensusIncome(
```
romanlutz, line 94: If I understand correctly, you're doing "adult census" as default, and in the other two cases you load the other datasets. Why not do if three times? It just reads better.

romanlutz (approved):

apps/dashboard/src/model-assessment/App.tsx
imatiach-msft: Yes, that is what it is doing.  Hmm, I like this style better, it's more concise.

PR #1415, opened 2022-05-10
fix error on machines with pyspark installed where passed dataframe is not spark pandas

imatiach-msft:
## Description

PM ran into an error when using latest erroranalysis package on a machine with pyspark installed (Compute Instance), but where they were not passing in a spark pandas dataframe.
Strangely, the error was related to the import:

```
AttributeError: module 'pyspark' has no attribute 'pandas'
```

Importing ```import pypark``` worked, but using pyspark.pandas.frame.DataFrame did not - but it started working after explicitly importing ```import pyspark.pandas as ps```.  Note on older versions of pyspark, the namespace pyspark.pandas may not even exist though, so I added try/except just to be very safe and ensure this never causes issues.

## Checklist

### DIFF
--- a/erroranalysis/erroranalysis/_internal/utils.py
+++ b/erroranalysis/erroranalysis/_internal/utils.py
@@ -6,7 +6,7 @@ import random
 import numpy as np
 
 try:
-    import pyspark
+    import pyspark.pandas as ps
     spark_available = True
 except ImportError:
     spark_available = False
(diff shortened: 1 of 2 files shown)


gaugup (commented):
Do we have tests for pyspark in out gates in OSS?

imatiach-msft:
@gaugup yes, we do have a test that I added:
https://github.com/microsoft/responsible-ai-toolbox/blob/main/erroranalysis/tests/test_pyspark_surrogate_error_tree.py
I've only run it locally, it's not yet setup in a gate.  I was planning on eventually adding a new github action to test this using:

https://github.com/marketplace/actions/setup-apache-spark

It's quite complicated to setup spark on a machine locally so I think having a separate github action for spark tests would be the best way forward.

imatiach-msft:
this seems to be kind of a weird issue which happens when spark is installed, but the dataframe passed in is a pandas dataframe (as usual), or that's what I've understood from when I was able to reproduce it on a compute instance

gaugup (approved):

PR #1416, opened 2022-05-10
Bug fixes on 'Set value' not copying over feature values correctly in what if counterfactual panel

tongyu-microsoft:
This PR fixes issues on what if counterfactual panel:
(1) 'Set value' is not copying over the categorical feature values correctly
(2) 'Set value' is not working when selecting categorical feature with option index 0
(3) error message bar in what if counterfactual panel is not displayed correctly

## Description
1. 'Set value' is not copying over the categorical feature values correctly: when user clicks on 'Set value' of an example point, we are only copying the numeric values, but for those categorical feature values, they are not updated and passed to fetch data call `/predict`. 
The fix is to check if feature data type is string, if it is string, we pass the feature option index to setCustomRowProperty method, so the editingData can be set correctly.
**Before**:
When click on 'Set value' of example data point, the request we are making are using the wrong (previously selected) values.
**After**:
When click on 'Set value' of example data point, the request we are making are using the feature values of current selected example data point.

2. 'Set value' is not working when selecting categorical feature with option index 0: the issue was, when clicking on 'Set value' of a data point, if one of the categorical feature index is 0, setCustomRowProperty method was skipped. The fix is, check if newValue is 0, and only early return when `(!newValue && newValue !== 0)`

3. error message bar in what if counterfactual panel is not displayed correctly: the issue was, error message bar was in the wrong stack and not visible if user scrolls down in what if counterfactual panel. 
Before:
image

After:
image

## Checklist


libs/counterfactuals/src/lib/CounterfactualChart.tsx
```diff
@@ -814,7 +814,7 @@ export class CounterfactualChart extends React.PureComponent<
     isString: boolean,
     newValue?: string | number
   ): void => {
-    if (!this.temporaryPoint || !newValue) {
+    if (!this.temporaryPoint || (!newValue && newValue !== 0)) {
```
romanlutz, line 817: Are you checking for `newValue === undefined`?

romanlutz (approved):

libs/counterfactuals/src/lib/CounterfactualChart.tsx
tongyu-microsoft: Yes, and also `newValue ===“”` when the value is empty string

PR #1417, opened 2022-05-11
add postbuild branch trigger

romanlutz:
## Description

Adding a set of pipeline triggers for the `postbuild` branch.

## Checklist

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -19,7 +19,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
+    branches: [main, postbuild]
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/Ci-raiwigets-python-typescript.yml
+++ b/.github/workflows/Ci-raiwigets-python-typescript.yml
@@ -2,9 +2,9 @@ name: CI RAIWidgets Python Typescript
 
 on:
   push:
-    branches: [main]
+    branches: [main, postbuild]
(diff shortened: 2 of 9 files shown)


tongyu-microsoft (approved):

.github/workflows/CD.yml
```diff
@@ -19,7 +19,7 @@ on:
   push:
```
gaugup, line 19: Should a version of builds run after merge to postBuild branch?

.github/workflows/Ci-raiwigets-python-typescript.yml
```diff
@@ -2,9 +2,9 @@ name: CI RAIWidgets Python Typescript
 
 on:
   push:
```
gaugup, line 4: Would this publish the npm packages for postBuild?

imatiach-msft (approved):

.github/workflows/Ci-raiwigets-python-typescript.yml
romanlutz: This is just the build and test pipeline. No publishing included. You may be thinking about the CD pipeline, but I've excluded merged PRs to `postbuild` from triggering npm publish steps.

gaugup (approved):

.github/workflows/CD.yml
romanlutz: That would mean we're pushing to npm which we don't need/want to do.

PR #1418, opened 2022-05-11
filter out missing values from what if dropdown to prevent explanation dashboard from crashing

imatiach-msft:
## Description

A customer encountered an issue where the explanation dashboard crashed when looking at local importance values.  On closer inspection, we discovered this was due to missing values, specifically when trying to convert them to strings.  This PR filters the missing values from the dropdown options, so they will not be available.  The alternative might be to make them available as something like "null" or just "" to choose from, but it might be more complicated to ensure these scenarios work fully as intended end to end.

## Checklist


libs/core-ui/src/lib/util/getFeatureOptions.ts
```diff
@@ -14,14 +14,17 @@ export function getFeatureOptions(
       const key = JointDataset.DataLabelRoot + index.toString();
       const meta = jointDataset.metaDict[key];
       const options = meta.isCategorical
-        ? meta.sortedCategoricalValues?.map(
-            (optionText: string | number, index: number) => {
+        ? meta.sortedCategoricalValues
+            ?.filter(
+              (optionText: string | number, _: number) =>
+                optionText !== undefined && optionText !== null
```
xuke444, line 20: I think lint would not allow null to pass

libs/core-ui/src/lib/util/getFeatureOptions.ts
```diff
@@ -14,14 +14,17 @@ export function getFeatureOptions(
...
+            )
+            .map((optionText: string | number, index: number) => {
               if (typeof optionText !== "string") {
                 optionText = optionText.toString();
```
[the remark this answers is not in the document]
imatiach-msft: If it's a missing value, like defined in the example data, which is what the customer had in their dataset.  I assume it could be null or undefined, perhaps it's always only null in reality, but handling undefined just in case seems like a good idea.

libs/core-ui/src/lib/util/getFeatureOptions.ts
imatiach-msft: it seems to be passing, at least I saw null checks elsewhere too

xuke444: We definitely don't want null in our repo. it is very dangerous to mix null and undefined.
do you mind change to !!optionText

xuke444 (approved):

xuke444 (approved):

PR #1422, opened 2022-05-12
Add the user class name to causal UI strings

gaugup:
## Description

Add the user class name to causal UI strings. Previously it just said class 1. Now we infer the class from the dataset object in typescript.

Screenshots

Before:-
image

After:-
image

## Checklist


libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAggregateView/CausalAggregateView.tsx
```diff
@@ -108,15 +108,30 @@ export class CausalAggregateView extends React.PureComponent<ICausalAggregateVie
 
   private getContinuousDescription(): string {
     if (this.context.dataset.task_type === "classification") {
-      return localization.CausalAnalysis.AggregateView.continuousDescription;
+      let positiveClass = "1";
```
romanlutz, line 111: What if it's multiclass? Is "positive" even a thing then?

romanlutz (approved):
I'd argue that calling a class "positive" adds some interpretation to something we don't have knowledge of (that is: how did the user structure their classes?). 1 could very well be negative or a negative outcome. But that's a bigger problem that I'm sure you don't want to tackle in this PR with a tight deadline... Just food for thought ;-)

libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAggregateView/CausalAggregateView.tsx
gaugup: Can change the variable name if you think it is somewhat misleading. We don't have multiclass for causal.

gaugup: changed the variable name.

PR #1424, opened 2022-05-12
upgrade numpy to fix random segfault test failures

imatiach-msft:
## Description

I'm seeing random segfaults happening when running pytest.  This PR upgrades numpy from conda-forge channel to fix random pytest segfault test failures.

EDIT: The pytest test failures seemed to be resolved, but now some tests are getting stuck on macos.  It seems installing lightgbm from conda-forge fixes this somehow on macos.

## Checklist

### DIFF
--- a/.github/workflows/Ci-raiwigets-python-typescript.yml
+++ b/.github/workflows/Ci-raiwigets-python-typescript.yml
@@ -45,8 +45,10 @@ jobs:
         run: yarn buildall
 
       - if: ${{ matrix.operatingSystem == 'macos-latest' }}
-        name: Use Homebrew to install libomp on MacOS for lightgbm
-        run: brew install https://publictestdatasets.blob.core.windows.net/data/libomp-11.1.0.catalina.bottle.tar.gz
+        name: Install latest numpy from conda-forge for MacOS
+        shell: bash -l {0}
+        run: |
+          conda install --yes --quiet -c conda-forge numpy
 
       - if: ${{ matrix.operatingSystem != 'macos-latest' }}
         name: Install pytorch on non-MacOS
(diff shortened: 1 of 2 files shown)


erroranalysis/requirements-dev.txt
```diff
@@ -1,4 +1,4 @@
-pytest==7.0.1
+pytest<=7.1.2
```
gaugup, line 1: Can we strong pin this to 7.1.2? <= can pull an older version.

imatiach-msft: I can't pin to 7.1.2 directly because older versions of python are not supported.  7.1.2 only supports python >= 3.7:
https://pypi.org/project/pytest/7.1.2/
image

imatiach-msft: we are still testing python 3.6, perhaps we need to remove that at some point soon

romanlutz (approved):

.github/workflows/Ci-raiwigets-python-typescript.yml
```diff
@@ -44,9 +44,10 @@ jobs:
       - name: Build Typescript
         run: yarn buildall
 
-      - if: ${{ matrix.operatingSystem == 'macos-latest' }}
```
imatiach-msft, line 47: note: this libomp seems to be already installed by default and isn't needed actually based on logs

raiwidgets/requirements-dev.txt
```diff
@@ -9,9 +9,6 @@ requirements-parser==0.2.0
 
 wheel
 
-# Required for interpret-community 0.18.0
-lightgbm==2.3.0
```
imatiach-msft, line 13: note: this is already specified in requirements.txt for raiwidgets and isn't needed here

gaugup (approved):

xuke444 (approved):

PR #1435, opened 2022-05-17
Remove duplicate raiwidgets pytest suite

romanlutz:
## Description

We're running the raiwidgets pytest suite in both the CI-python pipeline and the CI-raiwigets-python-typescript pipeline. This PR removes the pytest command and codecov from the latter since the former is dedicated to this and we don't need to do it twice. This should speed up the pipelines a bit, too.

## Checklist


romanlutz:
closing since it's not quite as simple as I had thought. Will get back to parallelizing test code at a later point

PR #1437, opened 2022-05-17
fix dependency chart axis updating with incorrect values in explanation dashboard

imatiach-msft:
## Description

Customer noticed an issue in the explanation dashboard where changing the dependency plot feature in the aggregate feature importances tab causes the x-axis values to sometimes be "mixed", especially when changing between numeric and categorical features.
When debugging, the interesting thing I found is that we don't even pass the x-axis values to highcharts for numeric data, so this is clearly some internal issue with using update in the highcharts code instead of creating a new chart on feature change.

image

## Checklist


memoryz:
Thank you for the quick fix! Looks good to me.

tongyu-microsoft (approved):

PR #1267, opened 2022-03-11
Integrate Data Balance insights into the RAI Dashboard

ms-kashyap:
## Description

Our team at Microsoft believes that having a well balanced data representation is critical when developing models in a responsible way, specially in terms of fairness. Finding data balance insights in the data exploration stage reduces risks for unbalanced models that are pushed to production, such as [this one](https://www.reuters.com/article/us-amazon-com-jobs-automation-insight-idUSKCN1MK08G).

We added the ability to compute data balance insights on both [Spark](https://microsoft.github.io/SynapseML/docs/features/responsible_ai/Data%20Balance%20Analysis/) and [Python](https://pypi.org/project/raimitigations/).

However, these insights are hard to understand when printed out in a Python notebook. Therefore, we would like to incorporate these data balance insights as part of the Responsible AI Dashboard, specifically the Data Explorer part.

These visualizations do not include mitigation steps yet, but this is what we will work on next. To learn more about these measures, [please click here](https://microsoft.github.io/SynapseML/docs/features/responsible_ai/Data%20Balance%20Analysis/#measure-explanations).

---

Please note, the following things are still to be done, but I wanted to get this Pull Request out for review while I work on them:
- Localization
- Tests
- Documentation
- Extra notebook showing how the feature works with Spark data
- Finalization of the Python `DataBalanceManager`

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [x] responsibleai/core-ui
- [ ] responsibleai/counterfactuals
- [x] responsibleai/dataset-explorer
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret
- [ ] responsibleai/localization
- [ ] responsibleai/mlchartlib
- [x] responsibleai/model-assessment

Python packages changed:

- [ ] raiwidgets
- [x] responsibleai
- [ ] erroranalysis
- [ ] rai_core_flask

## Tests

(As mentioned above, I will work on tests while this PR is out for review)

- [ ] No new tests required.
- [ ] New tests for the added feature are part of this PR.
- [ ] I validated the changes manually.

## Screenshots (if appropriate):

Overall view:
Overall-View

Feature Balance Measures (gif):
Feature-Balance-Measures

Distribution Balance Measures (gif):
Distribution-Balance-Measures

Aggregate Balance Measures (gif):
Aggregate-Balance-Measures

## Documentation:

- [X] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1440, opened 2022-05-18
[Python] Introduce DataBalanceManager with unit tests

ms-kashyap:
## Description

About two months ago, I created a PR (#1267) that added Data Balance insights to the RAI Dashboard.
I got really good feedback that I should split up the PR into Python side and UX to make it easier to review.
This PR represents the Python side.

Additionally, I've added unit tests for:
- the `DataBalanceManager` (including serialization)
- everything I added under `responsibleai/_tools/data_balance` that `DataBalanceManager` calls
- the `raimitigations` package we are taking a dependency on

I will introduce:
- UX + unit tests as PR 2
- Documentation + Python notebooks as PR 3

The overall flow is:
- `DataBalanceManager` creates a `df` using `train` and `test` from RAIInsights
- When user does `rai_insights.data_balance.add()`, they have the option to override `train` and `test` with a `custom_data` which can either be a `pandas` or `spark` DataFrame
- Manager may override `train + test` with user-specified `custom_data`, and then manager infers the `backend` of the `df`
- Using this `df` and `backend`, manager calls `DataBalance` module to prepare the df and compute measures
- The `DataBalance` module uses a `Factory` to get a `{`Pandas, Spark} DataBalanceService` which computes the measures
- `Pandas` service uses `raimitigations` package and `Spark` service uses `synapseml` package to compute measures

## Checklist

- [X] New tests were added or changes were manually verified.


responsibleai/requirements.txt
```diff
@@ -7,6 +7,7 @@ lightgbm>=2.0.11
 numpy>=1.17.2
 numba<0.54.0
 pandas>=0.25.1
+raimitigations~=0.0.3
```
romanlutz, line 10: I see this on PyPI https://pypi.org/project/raimitigations/
Will it be public soon?

responsibleai/responsibleai/managers/data_balance_manager.py
```diff
@@ -0,0 +1,321 @@
...
+        self._pos_label = None
+
+        # Populated in compute()
+        self._data_balance_measures = None
+
+    def add(
+        self,
+        cols_of_interest: List[str],
+        pos_label: Optional[str] = None,
+        custom_data: Optional[Any] = None,
```
romanlutz, line 71: I'm not sure I follow. We have a dataset via train and test. In which scenario would I want to pass more data?
I may just not have thought of the case you're thinking of which is totally fine, but the API reference should be able to explain it to me which it currently doesn't seem to be.

romanlutz (commented):
I think I got the gist in about 20 minutes of reading this, but to be honest 2400+ lines of python is still a lot. Any chance we could have the first PR be the data balance scaffold, and then two more for the spark part and the synapseml part?

There's one more thing: we've locked the `main` branch until after //Build. If you're hoping to merge this before sometime early June you need to target the `postbuild` branch which will be merged into `main` at that point.

responsibleai/requirements.txt
ms-kashyap: Yes, it's currently hosted in a private Github repo called "responsible-ai-mitigations" and both the repo and package will be public soon!

responsibleai/responsibleai/managers/data_balance_manager.py
ms-kashyap: I included a `custom_data` parameter for scenarios where:
- You want to run data balance on your full dataset. I noticed from many sample notebooks that the `train` and `test` data are usually sampled.
- If the dataset is a `spark.sql.DataFrame`. We mainly did this so we can use [the Spark Data Balance](https://microsoft.github.io/SynapseML/docs/features/responsible_ai/Data%20Balance%20Analysis/) to compute measures.

ms-kashyap:
> I think I got the gist in about 20 minutes of reading this, but to be honest 2400+ lines of python is still a lot. Any chance we could have the first PR be the data balance scaffold, and then two more for the spark part and the synapseml part?
> 
> There's one more thing: we've locked the `main` branch until after //Build. If you're hoping to merge this before sometime early June you need to target the `postbuild` branch which will be merged into `main` at that point.

I really feel that if I were to split it up into smaller PRs, you and other reviewers would lose a lot of context into the feature as a whole. 
The PR is 900 lines of the feature itself and 1500 lines of unit tests. I definitely get that it may seem unruly, so I don't mind splitting it up, but again I fear you will lose context.

Also thank you for pointing that the //Build thing. That's not an issue; I believe that this and its associated UX part should not be rushed so I'll keep this targeted to `main`.

ms-kashyap:
> > I think I got the gist in about 20 minutes of reading this, but to be honest 2400+ lines of python is still a lot. Any chance we could have the first PR be the data balance scaffold, and then two more for the spark part and the synapseml part?
> > There's one more thing: we've locked the `main` branch until after //Build. If you're hoping to merge this before sometime early June you need to target the `postbuild` branch which will be merged into `main` at that point.
> 
> I really feel that if I were to split it up into smaller PRs, you and other reviewers would lose a lot of context into the feature as a whole. The PR is 900 lines of the feature itself and 1500 lines of unit tests. I definitely get that it may seem unruly, so I don't mind splitting it up, but again I fear you will lose context.
> 
> Also thank you for pointing that the //Build thing. That's not an issue; I believe that this and its associated UX part should not be rushed so I'll keep this targeted to `main`.

Hey, I spoke to Ilya offline and I'll set up a meeting to go over the PR for 30 minutes after //Build conference!

responsibleai/responsibleai/managers/data_balance_manager.py
romanlutz: I believe the sampling is a choice due to lack of scalability in the dashboard code. @gaugup is looking into resolving this the next few months AFAIK.

It just seems unintuitive from a user perspective to have another data param, especially with a name that doesn't indicate what it's for. Why not pass the spark SQL DF as train and test? @imatiach-msft may have thoughts, too.

responsibleai/requirements.txt
romanlutz: It would be nice to see that publicly on GitHub first, because otherwise we're integrating something private into this open source tool which isn't quite in the spirit of open source 🙂

PR #1447, opened 2022-05-24
Fix TypeError: Cannot read properties of undefined (reading 'treatAsCategorical')"

vinuthakaranth:
## Description

Fix TypeError: Cannot read properties of undefined (reading 'treatAsCategorical')".
metaDict[xKey] can be undefined leading to this error.

## Checklist


tongyu-microsoft (approved):

libs/interpret/src/lib/MLIDashboard/Controls/GlobalExplanationTab/GlobalExplanationTab.tsx
```diff
@@ -438,7 +438,7 @@ export class GlobalExplanationTab extends React.PureComponent<
     // set to dependence plot initially, can be changed if other feature importances available
     const xKey = JointDataset.DataLabelRoot + featureIndex.toString();
     const xIsDithered =
-      this.context.jointDataset.metaDict[xKey].treatAsCategorical;
+      this.context.jointDataset.metaDict[xKey]?.treatAsCategorical;
```
gaugup, line 441: Is there a unit test for this change? In what situations can treatAsCategorical be not available as a property?

imatiach-msft (approved):

libs/interpret/src/lib/MLIDashboard/Controls/GlobalExplanationTab/GlobalExplanationTab.tsx
vinuthakaranth: In this case when xKey isn't available, treatAsCategorical will be undefined. @imatiach-msft [imatiach-msft] or @riedgar-ms  any idea when is xKey not available?

imatiach-msft: hmm, no idea.  It's unfortunate we don't have a repro of this issue...

romanlutz (approved):

libs/interpret/src/lib/MLIDashboard/Controls/GlobalExplanationTab/GlobalExplanationTab.tsx
riedgar-ms: Is there any logging we can usefully do for when the 'null' case is hit? Or are the usual values of xKey considered customer data?

gaugup: are there other places in the code that we need to fix knowing that `treatAsCategorical` can be optional?

gaugup (approved):

libs/interpret/src/lib/MLIDashboard/Controls/GlobalExplanationTab/GlobalExplanationTab.tsx
gaugup: Thanks for fixing this everywhere.

vinuthakaranth: Thanks for catching and commenting about this!! :)

PR #1448, opened 2022-05-24
Add more unittests RAI dashboard input class 

gaugup:
## Description

Seems like there are missing tests for responsibleai_dashboard_input.py

## Checklist


raiwidgets/tests/test_responsibleai_dashboard_input.py
```diff
@@ -1,24 +1,222 @@
...
-
+from raiwidgets.interfaces import WidgetRequestResponseConstants
 from raiwidgets.responsibleai_dashboard_input import \
     ResponsibleAIDashboardInput
 
 
-class TestResponsibleAIDashboardInput:
-    def test_model_analysis_adult(
+class TestResponsibleAIDashboardInputClassification:
+    def test_model_analysis_adult_on_predict_success(
```
imatiach-msft, line 10: nit: change all test names to use rai_insights or responsibleai since model_analysis isn't a term we want to use anymore

imatiach-msft (approved):

raiwidgets/tests/test_responsibleai_dashboard_input.py
gaugup: Done. Change to rai_dashboard_input instead of model_analysis.

PR #1450, opened 2022-05-25
Update the way to get the length of elements obtained in e2e tests

vinuthakaranth:

## Description

One of the causal test was failing for the way we were getting the length of the elements. This is the updated way to get the length.

image

Add collapse dropdown before clicking on cancel button in panel
## Checklist


gaugup:
How was this passing earlier?

vinuthakaranth:
Surprisingly, this isn't failing in oss at all. This issue is present when using e2e lib in AML. So updated to easy code as sing be was not recommended. But nothing wrong with using 'be' though.

tongyu-microsoft (approved):

Issue #1453, opened 2022-05-27
Data Explorer chart does not render when count is on the Y axis


jamescarpinter:
Repo instructions:
1. Install the latest `raiwidgets` package (0.18.2).
2. Run either Housing sample notebook (`$/notebooks/responsibleaidashboard/responsibleaidashboard-housing-decision-making.ipynb` or `$/notebooks/responsibleaidashboard/responsibleaidashboard-housing-classification-model-debugging.ipynb`).
3. Run the notebook and open the link to the dashboard once available.
4. In the Data Explorer panel, change the Y-axis to "Count".
5. Observe the graph is no longer visible. Switching to other Y axis values still show the graph.

This worked as expected in v0.17.

imatiach-msft:
closing as PR https://github.com/microsoft/responsible-ai-toolbox/pull/1454 has been merged

PR #1454, opened 2022-05-31
fix dataset explorer plot with count not displaying

imatiach-msft:
## Description

Fixed issue https://github.com/microsoft/responsible-ai-toolbox/issues/1453 where count plot is not displaying.

The main problem was that count uses "none" as the metadata name, which does not have any sorted categorical values (see https://github.com/microsoft/responsible-ai-toolbox/blob/bc7a9f91445db86e41784c9355bc7b9d230c7cb2/libs/core-ui/src/lib/util/JointDataset.ts#L739 which sets treatAsCategorical=True, but does not specify categorical values).  However, when creating the highchart options, we skip the step to pass the values of the "none" group, since we iterate over all categorical values.  There are several possible fixes here (another I had was to insert an array with some dummy "count" value for the categoricals values in JointDataset which also worked), but I felt this was the least intrusive and safest one.

Screenshot of fixed plot:
image

## Checklist

### DIFF
--- a/libs/dataset-explorer/src/lib/getDatasetBar.ts
+++ b/libs/dataset-explorer/src/lib/getDatasetBar.ts
@@ -24,10 +24,11 @@ export function getDatasetBar(
     chartProps?.xAxis.property &&
     jointData.metaDict[chartProps.xAxis.property].sortedCategoricalValues
       ?.length;
+  const noneGroup = "none";
 
   if (customData && xData) {
     for (const [i, customDatum] of customData.entries()) {
-      const yValue = (customDatum as any).Y;
+      const yValue = (customDatum as any).Y ?? noneGroup;
       if (!groupedData[yValue]) {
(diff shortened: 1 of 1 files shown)


tongyu-microsoft (approved):

romanlutz:
In the screenshot I don't see `none` showing up. What am I missing? [Note: I don't mean to hold this up at all. Feel free to proceed in any case.]

tongyu-microsoft:
> In the screenshot I don't see `none` showing up. What am I missing? [Note: I don't mean to hold this up at all. Feel free to proceed in any case.]

@romanlutz, I checked that locally, the `none` shows up in the pop up when you hover over each column in the chart. For other y-axis choices like categorical features, we are displaying `apple: 5, banana: 8`. And for Ilya's change for `count`, we display `none`

romanlutz (approved):

PR #1455, opened 2022-06-01
[Bugfix] HeatmapHighChart: Prioritize configOverride over defaultOptions

ms-kashyap:
### Context
When trying to create a heatmap using `HeatmapHighChart`, I had set `legend: { enabled : true }` but I was not seeing the legend in the heatmap.

It turns out that `lodash.merge` prioritizes values from right to left. Therefore, we should specify `configOverride` last since we want its values to be prioritzed over `defaultOptions`.

### Showcasing bug

Original code:
image
image

### Showcasing fix

Fix (swap `configOverride` and `defaultOptions`): 
image
image

## Checklist


ms-kashyap:
@romanlutz helped me realize that CI/CD from my fork won't work, so I re-created PR #1457 from the main repo instead of my fork. Closing this one.

PR #1457, opened 2022-06-01
[Bugfix] HeatmapHighChart: Prioritize configOverride over defaultOptions

ms-kashyap:
Note: Recreating PR #1455 from the main repo instead of my fork, so that CI/CD can work properly.

### Context
When trying to create a heatmap using `HeatmapHighChart`, I had set `legend: { enabled : true }` but I was not seeing the legend in the heatmap.

It turns out that `lodash.merge` prioritizes values from right to left. Therefore, we should specify `configOverride` last since we want its values to be prioritzed over `defaultOptions`.

### Showcasing bug

Original code:
image
image

### Showcasing fix

Fix (swap `configOverride` and `defaultOptions`): 
image
image

## Checklist


xuke444 (approved):

romanlutz:
This looks great. The thing is we have the `main` branch blocked for urgent fixes until probably Monday. We might relax that requirement as early as tomorrow, but no guarantees on that. If you're not in a hurry we can just wait until then. Otherwise consider targeting postbuild

ms-kashyap:
No rush at all, feel free to complete this at your convenience! 😄

romanlutz:
I checked the blob and don't see any significant differences to before. That doesn't mean there are none, but at least it passes the cursory glance check.

PR #1458, opened 2022-06-01
Update predicted class value to uneditable text in counterfactual panel 

tongyu-microsoft:
We shouldn't allow users to edit the predicted value column. So this PR is to update predicted value from editable input text field to uneditable text. 

## Description
**Before:
Predicted value is editable:**
image
image

**After:
Predicted value is not editable:**
uneditablePredictedClass
image

## Checklist


libs/counterfactuals/src/lib/CounterfactualList.tsx
```diff
@@ -369,6 +371,22 @@ export class CounterfactualList extends React.Component<
...
+        <Stack horizontal={false} tokens={{ childrenGap: "s1" }}>
+          <Stack.Item className={styles.dropdownLabel}>
+            <Text>{column.name}</Text>
+          </Stack.Item>
+          <Stack.Item>
+            <Text>{predictedClass}</Text>
+          </Stack.Item>
+        </Stack>
+      );
+    }
```
gaugup, line 389: Can you add a test for this change in describeWhatIfCreate.ts?

tongyu-microsoft: Thanks for the comment, I will add E2E test in describeWhatIfCreate.ts in the other PR: https://github.com/microsoft/responsible-ai-toolbox/pull/1459
Close this for now.

PR #1459, opened 2022-06-01
Fix unable to input decimal point in what if counterfactual feature text field

tongyu-microsoft:
Users are not able to input decimal point in what if counterfactual to update feature value. The root cause is, we check on column value on each update. One example is, when user wants to type in "18.5" and on typing "18.", `toNumber()` ignores the decimal point, result in user can only input "185" instead of the desired "18.5". The fix is to pass string to this.state.displayedData when the input field value ends with decimal point. 

## Description
Before:
Unable to input decimal point:
decimalCFAfter

After:
decimalCFAfter

## Checklist


libs/counterfactuals/src/lib/CounterfactualList.tsx
```diff
@@ -336,7 +336,9 @@ export class CounterfactualList extends React.Component<
       this.props.data?.feature_names_including_target.indexOf(id);
     this.props.setCustomRowProperty(`Data${keyIndex}`, false, newValue);
     this.setState((prevState) => {
-      prevState.data[id] = toNumber(newValue);
+      prevState.data[id] = newValue?.endsWith(".")
+        ? newValue
```
gaugup, line 340: So if the value has "." then it will be treated as a string? Can we test the prediction what it returns in this case?

tongyu-microsoft: yeah that's a good point. We will throw exception while predicting, as data must be int, float or bool:
image

tongyu-microsoft: I am thinking of 2 approaches:
(1) From UI side, we check and convert to number if there is any string.
(2) Without updating UI, we create this.state.displayedData, where diaplayedData can be string | number, and we limit this.state.data to be number[] only.

tongyu-microsoft: Added this.state.displayedData. So this.state.data is the same as before, the prediction call won't hit error in this case. 
And we allow users to input decimal point by adding this.state.displayedData

gaugup:
Can you please capture this change in the e2e tests?

libs/counterfactuals/src/lib/CounterfactualList.tsx
```diff
@@ -76,6 +78,7 @@ export class CounterfactualList extends React.Component<
     super(props);
     this.state = {
       data: {},
+      displayedData: {},
```
romanlutz, line 81: Storing the data twice could eventually lead to inconsistencies if we're not extremely careful. I'm wondering if it's possible to manipulate the data before displaying rather than storing it twice. Did you consider that option? If so, I'm curious why that might not work.

tongyu-microsoft:
Add E2E test for targetClass in describeWhatIfCreate.ts

libs/counterfactuals/src/lib/CounterfactualList.tsx
tongyu-microsoft: No, this is only for UI display. The request we send to /predict call is still decimal value.

libs/counterfactuals/src/lib/CounterfactualList.tsx
tongyu-microsoft: Removed the change so that we only store the data once. The reason we can remove this is because, we don't need to store a numeric version of data here. We can just pass the string version with decimal point, and it is converted to number before sending the request (in setCustomRowProperty method inside CounterfactualChart.tsx where we have `const asNumber = +newValue`)

tongyu-microsoft:
> Add E2E test for targetClass in describeWhatIfCreate.ts

Tests are added

PR #1462, opened 2022-06-02
Add e2e tests for new model overview experience (part 1)

romanlutz:
## Description

While adding e2e tests for the new model overview experience I encountered several challenges that needed to be addressed first. Therefore, this PR contains somewhat more than just the tests:

- `e2e-widget.js`: 
  - ability to execute individual notebooks only instead of the entire set of notebooks - this is done via a parameter for `yarn e2e-widget` called `-n`. Also, I added a parameter `--skipgen` to skip the notebook generation since that takes a while every time, too. When debugging this locally it can save lots of time to skip the notebook generation.
  - ability to generate notebooks with flights injected! We just pass `feature_flights` to the dashboard constructor and voila we can run e2e tests on the flighted UI. Note that the flighted notebook has a different name (which includes the flight name(s)). Right now there's only 1 flight, but if there were multiple we'd need to specify which combinations we want to have e2e tests for and add spec files accordingly (more on spec files below).
- notebook/test case naming standardization: There was some inconsistency before between test case naming and notebook naming. This complicated the aforementioned individual notebook execution considerably, so I've gone ahead and standardized the names. This accounts for a lot of the changed files, although it should be trivial to skim through.
- renamed `modelStatistics.spec.ts` to `modelOverview.spec.ts` - simply because we don't call it model statistics anymore, not even in the "old" model overview. The "new" model overview spec files just call `describeModelOverview`.
- added a very basic set of tests for the new model overview experience in the e2e package (`describeNewModelOverview` under `v2`) and moved the existing ones into a `v1` folder. Tests don't cover all tasks yet (only regression as of now) and are only applied to 1 notebook for now. The rest will be in a part 2 PR.
- lots of spec files for the flighted notebooks. These aren't necessarily very different. The only file that actually differs is the one about model overview (using `describeNewModelOverview`). You can easily imagine future scenarios where there is quite a bit of difference so this duplication may end up being less duplication than it looks right now. Fortunately, I found a nice way to avoid duplicating the "expected" values (see `modelAssessmentDatasets.ts`). Right now, only one notebook, `diabetes-regression-model-debugging`, actually uses the new tests that I wrote in this PR. I will have another PR to spread it to all notebooks. It just felt like it's getting too large to add a lot more logic here.
- fix `random_state` in notebooks where needed. Without deterministic outputs it's kind of hard to test for expected values... I'm kind of amazed we didn't run into this before.

For the record, I fully expect to delete everything v1/old as soon as we are confident about v2/new. This will not linger around like other "old" parts that we've seen in the past.

## Checklist


scripts/e2e-widget.js
```diff
@@ -15,6 +17,7 @@ const fileNames = [
   "responsibleaidashboard-housing-decision-making",
   "responsibleaidashboard-multiclass-dnn-model-debugging"
 ];
+const flightCombinations = [[], ["newModelOverviewExperience"]];
```
vinuthakaranth, line 20: why do we have this empty array?

notebooks/responsibleaidashboard/responsibleaidashboard-census-classification-model-debugging.ipynb
```diff
@@ -119,7 +119,7 @@
     "    # Append classifier to preprocessing pipeline.\n",
     "    # Now we have a full prediction pipeline.\n",
     "    pipeline = Pipeline(steps=[('preprocessor', feat_pipe),\n",
-    "                               ('model', LGBMClassifier())])\n",
+    "                               ('model', LGBMClassifier(random_state=0))])\n",
```
vinuthakaranth, line 122: jfmk: why is random_state added?

libs/e2e/src/util/comboBox.ts
```diff
@@ -20,3 +20,19 @@ export function selectComboBox(selector: string, item: string | number): void {
 export function getComboBoxValue(selector: string): string | undefined {
   return cy.$$(`${selector} input.ms-ComboBox-Input`).val()?.toString();
 }
+
+export function multiSelectComboBox(selector: string, item: string | number, firstSelect?: boolean): void {
```
vinuthakaranth, line 24: Thanks for adding this!

vinuthakaranth:
Will we move these tests to AML once v2 of model overview is fully released?

libs/e2e/src/lib/describer/modelAssessment/IModelAssessmentData.ts
```diff
@@ -135,10 +148,16 @@ export interface IWhatIfCounterfactualsData {
...
   "HousingClassificationModelDebugging" = "responsibleaidashboard-housing-classification-model-debugging.py",
+  "HousingClassificationModelDebuggingNewModelOverviewExperience" = "responsibleaidashboard-housing-classification-model-debugging-newModelOverviewExperience.py",
   "DiabetesDecisionMaking" = "responsibleaidashboard-diabetes-decision-making.py",
+  "DiabetesDecisionMakingNewModelOverviewExperience" = "responsibleaidashboard-diabetes-decision-making-newModelOverviewExperience.py",
   "HousingDecisionMaking" = "responsibleaidashboard-housing-decision-making.py",
-  "MulticlassDnnModelDebugging" = "responsibleaidashboard-multiclass-dnn-model-debugging.py"
+  "HousingDecisionMakingNewModelOverviewExperience" = "responsibleaidashboard-housing-decision-making-newModelOverviewExperience.py",
+  "MulticlassDnnModelDebugging" = "responsibleaidashboard-multiclass-dnn-model-debugging.py",
+  "MulticlassDnnModelDebuggingNewModelOverviewExperience" = "responsibleaidashboard-multiclass-dnn-model-debugging-newModelOverviewExperience.py"
```
vinuthakaranth, line 162: Do we use all of these newly added python notebook enums?

scripts/e2e-widget.js
romanlutz: The default is no flights. I'll add a comment to document this!

notebooks/responsibleaidashboard/responsibleaidashboard-census-classification-model-debugging.ipynb
romanlutz: It's impossible to assert anything if we don't fix the inherently random behavior of these ML techniques. Every execution will result in a different outcome. This fixes it since we provide the random state rather than it choosing a random state itself. People can easily remove or change it if they like, of course, but this makes the notebook reproducible.

libs/e2e/src/util/comboBox.ts
romanlutz: Sure! Strangely, the `click` is only necessary on the first call and not on further ones. Couldn't quite figure out why, but I guess it doesn't matter as long as it works as intended.

romanlutz:
> Will we move these tests to AML once v2 of model overview is fully released?

I would assume so! I haven't quite had a chance to look into how we're testing for AML in detail yet, but if it's as simple as calling `describe<insert component name>` then I'm sure we'll want to do it!

romanlutz:
> The steps to run e2e still remains same? should we add updated script commands to https://github.com/microsoft/responsible-ai-toolbox/blob/main/CONTRIBUTING.md

Current specs:
```
    "e2e": "nx e2e dashboard-e2e",
    "e2e-watch": "yarn e2e dashboard-e2e --watch",
    "e2e-widget": "node ./scripts/e2e-widget.js",
```
`e2e` here means the non-notebook tests, `e2e-watch` is the same with individual test case execution, and `e2e-widget` is the notebook e2e tests. I've only adjusted the `e2e-widget` script, but will update the instructions now.

notebooks/responsibleaidashboard/responsibleaidashboard-census-classification-model-debugging.ipynb
romanlutz: Just to be sure I'm adding @imatiach-msft in case this makes the experience bad for users.

imatiach-msft: @romanlutz  yes this makes sense to me, sure!

vinuthakaranth (approved):

notebooks/responsibleaidashboard/responsibleaidashboard-census-classification-model-debugging.ipynb
romanlutz: Thanks for confirming!

apps/widget-e2e/src/integration/modelAssessment/responsibleaitoolboxMulticlassDnnModelDebuggingNewModelOverviewExperience/causalAnalysis.spec.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import {
+  describeCausalAnalysis,
+  modelAssessmentDatasets
+} from "@responsible-ai/e2e";
+
+const datasetShape = modelAssessmentDatasets.MulticlassDnnModelDebugging;
+describeCausalAnalysis(datasetShape, "MulticlassDnnModelDebuggingNewModelOverviewExperience");
```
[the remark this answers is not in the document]
romanlutz: For sure! I'll look into that as soon as I got it merged here :-)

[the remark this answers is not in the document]
romanlutz: ... and the naming will be updated when we're confident about removing the old experience (v1).

romanlutz:
Closing this PR to open a new one targeted at `main`!

PR #1464, opened 2022-06-02
Release SciKit-Learn upper bound

riedgar-ms:
## Description

The issue encountered with causal analysis appears to be fixed with SciKit-Learn v1.1.1. Probably [this issue](https://scikit-learn.org/dev/whats_new/v1.1.html#sklearn-tree).

## Checklist


responsibleai/tests/causal/test_causal_general.py
```diff
@@ -10,6 +10,7 @@
 
 def test_causal_classification_scikitlearn_issue():
     # This test gets stuck on SciKit-Learn v1.1.0
+    # Issue appears fixed with v1.1.1
```
romanlutz, line 13: So... it seems the most reasonable fix would be to say >1.1 but that excludes lots of older versions that do work as well. Am I right in assuming we don't want to be that strict because lots of users are using these older versions?

imatiach-msft (approved):

responsibleai/tests/causal/test_causal_general.py
riedgar-ms: Double checks... I hadn't realised that `!=` was allowed:
https://peps.python.org/pep-0508/

riedgar-ms:
Per comment in Teams thread, we may want to pause on this, since I believe that SciKit-Learn doesn't tend to keep pickle-compatibility between versions.

responsibleai/tests/causal/test_causal_general.py
riedgar-ms: However, per my other comment, I'm rather worried that this will break backwards compatibility. I understand that SciKit-Learn doesn't tend to be pickle-compatible between versions?

romanlutz: This would mean we have to be quite strict on versions, wouldn't it?

riedgar-ms: Or, as we were just discussing in the other meeting, start thinking some more about MLFlow.

gaugup: Should this test just run for scikit-learn version greater than 1.0?

romanlutz: >= 1.0 AND ~= 1.1.0 ?
Not sure how one does the AND logic here...

romanlutz: Reviving this thread: This seems important in that we don't allow people to use the latest 3 versions of sklearn with the current logic in `main`. Any objections with merging as is until we want to require `>1.1.0` at which point this special case goes away anyway?

@imatiach-msft @gaugup
Not sure if @riedgar-ms still has thoughts on this, but if so please share them, too.

imatiach-msft:
closing this PR as it has been superseded by merged PR https://github.com/microsoft/responsible-ai-toolbox/pull/2173

PR #1158, opened 2022-01-20
Separate out raiwidgets and UI tests into its own workflow

gaugup:
## Description

It seems like the responsible-ai-toolbox/CI-python.yml at main · microsoft/responsible-ai-toolbox (github.com) file has a lot of custom code for raiwidgets. Also, because the widget tests uses yarn and run widget tests it is at times flaky. At times when the raiwidgets tests fail, we need to requeue all the other jobs (responsibleai, erroranalysis and rai_core_flask). This makes the PR merge time longer.
 
This PR separates the raiwidgets tests and UI tests into a separate workflow. This way we don't need to maintain custom code in the same yml for raiwidgets and we can isolate the flaky pipeline tests and prevent needless reruns of the stable tests.

## Areas changed

npm packages changed:

- [ ] responsibleai/causality
- [ ] responsibleai/core-ui
- [ ] responsibleai/counterfactuals
- [ ] responsibleai/dataset-explorer
- [ ] responsibleai/fairness
- [ ] responsibleai/interpret
- [ ] responsibleai/localization
- [ ] responsibleai/mlchartlib
- [ ] responsibleai/model-assessment

Python packages changed:

- [ ] raiwidgets
- [ ] responsibleai
- [ ] erroranalysis
- [ ] rai_core_flask

## Tests

- [x] No new tests required.
- [ ] New tests for the added feature are part of this PR.
- [ ] I validated the changes manually.

## Documentation:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1478, opened 2022-06-09
Split raiwidgets/e2e-notebooks test pipeline in two

romanlutz:
## Description

The existing pipeline `.github\workflows\Ci-raiwigets-python-typescript.yml` serves two distinct purposes:
- running raiwidgets tests (pytest)
- running e2e notebook tests
They are in one pipeline since they both require `yarn install`, `yarn buildall`, and the subsequent installation of `raiwidgets`. However, this pipeline has become the most time-consuming of all of our pipelines and it will only get worse in the future as we add more notebooks and more test cases. For that reason, this PR splits them into two and prunes unnecessary steps (e.g., installing torch for the pytest tests). The e2e-notebook pipeline isn't run on macos-latest which also removed a lot of custom parts. So far, this was just handled via `if` statements.

Beyond that, the e2e-notebook pipeline is also broken down into each individual notebook using the matrix strategy. This required extending `e2e-widget.js` to take additional input params `-n` for the notebook name and `--skipgen` to skip notebook generation which considerably speeds up the local test iterations. In this new setup, the notebook names need to match the test folders so I renamed some of the test folders to match the notebook names.

Furthermore, this now does `yarn install` and `yarn buildall` in a single workflow, stores the output as an artifact and installs `raiwidgets` using this artifact in the individual workflows. This means we're running the `yarn buildall` part only once instead of `n` times. It often takes up to 30 minutes for just that part, so the savings should be considerable. 

This change was originally part of #1462 but the PR got too big. In that PR, I'm adding e2e notebook tests for flighted versions of our dashboard. This alone doubles the test suite and made it run out of memory on the node that ran it. Splitting it up was the reasonable adjustment, but we felt like it makes sense to go a step further as we'll be adding more notebooks in the future (presumably).

## Checklist


.github/workflows/CI-raiwidgets-pytest.yml
```diff
@@ -7,7 +7,7 @@ on:
     branches: [main]
 
 jobs:
-  ci-raiwidgets-python-typescript:
+  ci-raiwidgets-pytest:
     env:
```
gaugup, line 11: If this is just running pytest, then could you move raiwidgets under CI-python.yml?

romanlutz: I can't actually, since it's required to do `yarn install` and `yarn buildall` first to build `raiwidgets`. That's why it was a separate pipeline to begin with.

imatiach-msft: they used to all be in one pipeline but raiwidgets was too flaky and it also took much more time than the others, hence it was moved into its own pipeline

imatiach-msft: see PR for context:
https://github.com/microsoft/responsible-ai-toolbox/pull/1158
the description on the PR is really good

romanlutz: Thanks for linking! Exactly. That's why they should remain separate from CI-python.

gaugup: Why do we need to do `yarn install` and `yarn buildall` to run pytest for raiwidgets? These are python SDK tests so should be able to  run without `yarn`

xuke444 (approved):

.github/workflows/CI-raiwidgets-pytest.yml
romanlutz: Great question! I was wondering that myself a few weeks ago until I realized that we want to test this particular version of raiwidgets on the PR branch. So if you look into the raiwidgets folder there's no JS code there. To get it there we need to run `yarn buildall` and then install the python package. If you don't do the yarn part then it'll fail as soon as you call the dashboard. If you're interested I can walk you through the handover between the two at some point 🙂

gaugup: Makes sense. I had a hunch but wasn't entirely sure. Thanks for clarifying this.

.github/workflows/CI-e2e-notebooks.yml
```diff
@@ -0,0 +1,124 @@
...
+
+  ci-e2e-notebook:
+    needs: ui-build
+
+    env:
+      node-version: 16.8
+
+    strategy:
+      matrix:
+        operatingSystem: [ubuntu-latest, windows-latest]
```
gaugup, line 47: Seems like these tests aren't being run on MAC OS> Could we add a comment that these need to be added at some point. I know @vinuthakaranth has a work item for re-enabling these tests for MAC OS>

romanlutz: Certainly! Just added a TODO comment right above that line.

gaugup (approved):

scripts/e2e-widget.js
```diff
@@ -129,19 +171,23 @@ function e2e(watch) {
...
   }
-  console.log("E2e finished\r\n");
+  console.log("e2e finished\r\n");
 }
 
 async function main() {
   commander
     .option("-w, --watch", "Watch mode")
+    .option("--skipgen", "Skip notebook generation")
+    .option("-n, --notebook [notebook]", "Run specific notebook")
```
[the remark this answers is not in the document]
romanlutz: Copying it over from my other PR! Thanks!

vinuthakaranth (approved):

PR #1485, opened 2022-06-09
Create DataBalanceManager + unit tests; clean up data balance conftest

ms-kashyap:
## Description

Creating a `DataBalanceManager` that can be added to RAIInsights via `rai_insights.data_balance.add()`.
This manager's `compute()` creates a `data_balance_measures` dict which is part of the the `Dataset` interface.
The manager uses the `databalanceanalysis` module that was recently added to compute metrics.
Additionally, there is a helper which helps transform dataframe --> dict for each of the 3 sets of measures.

Originally, I had created a PR (#1449) assuming that the `databalanceanalysis` module wasn't going to be a part of the `responsibleai` package. Now that it is, it eliminates a lot of redundant code from that PR (such as removing `PandasDataBalanceBackend` that calls an external library).

Also, addressed comments that @imatiach-msft and @gaugup had in that PR.

## Checklist


responsibleai/responsibleai/databalanceanalysis/data_balance_helper.py
```diff
@@ -0,0 +1,200 @@
...
+    Helper class for computing and transforming data balance measures
+    to be used in a dashboard.
+    """
+
+    @staticmethod
+    def prepare_df(
+        df: pd.DataFrame, target_column: str, pos_label: Optional[str]
+    ) -> pd.DataFrame:
+        """
+        Prepare dataframe for data balance analysis.
```
romanlutz, line 27:
```suggestion
        Prepare DataFrame for data balance analysis.
```

responsibleai/responsibleai/databalanceanalysis/data_balance_helper.py
```diff
@@ -0,0 +1,200 @@
...
+
+        :param df: The dataframe to prepare.
+        :type df: pd.DataFrame
+        :param target_column: The target column in the dataframe.
```
romanlutz, line 31:
```suggestion
        :param target_column: The target column in the DataFrame.
```

responsibleai/responsibleai/databalanceanalysis/data_balance_helper.py
```diff
@@ -0,0 +1,200 @@
...
+        :type target_column: str
+        :param pos_label: The positive label of the target column.
+        :type pos_label: Optional[str]
+        :return: The prepared dataframe.
```
romanlutz, line 35:
```suggestion
        :return: The prepared DataFrame.
```

responsibleai/responsibleai/databalanceanalysis/data_balance_helper.py
```diff
@@ -0,0 +1,200 @@
...
+        :rtype: pd.DataFrame
+        """
+        try:
+            # Deepcopy the df otherwise the original df will mutate
+            df = df.copy(deep=True)
+
+            # If the positive label is specified, transform the target column
+            # to {0, 1} because Data Balance Analysis only supports binary
+            # classification for now.
```
romanlutz, line 44: Are we surfacing that limitation somewhere? For example, is there an error warning the user if we have, say, a regression task?

responsibleai/responsibleai/databalanceanalysis/data_balance_helper.py
```diff
@@ -0,0 +1,200 @@
...
+            DISTRIBUTION_BALANCE_MEASURES_KEY: dist_measures_dict,
+            AGGREGATE_BALANCE_MEASURES_KEY: agg_measures_dict,
+        }
+
+    @staticmethod
+    def transform_feature_balance_measures(
+        df: pd.DataFrame,
+    ) -> Dict[str, Dict[str, Any]]:
+        """
+        Transform the feature balance measures df into a dictionary
```
romanlutz, line 107:
```suggestion
        Transform the feature balance measures DataFrame into a dictionary
```

responsibleai/responsibleai/managers/data_balance_manager.py
```diff
@@ -0,0 +1,318 @@
...
+                (
+                    "All columns in `cols_of_interest` must be present in"
+                    " the dataset."
+                )
+            )
+
+        if self._target_column not in self._df.columns:
+            raise ValueError(
+                (
+                    f"The target_column `{self._target_column}` must be"
```
romanlutz, line 115: It seems like sometimes there's a backquote and sometimes a normal quote (see next error message). Can we do this uniformly? I think normal quotes are fine (?)

responsibleai/responsibleai/managers/data_balance_manager.py
```diff
@@ -0,0 +1,318 @@
...
+        )
+
+    def _set_data_balance_measures(
+        self,
+        feature_balance_measures: pd.DataFrame,
+        distribution_balance_measures: pd.DataFrame,
+        aggregate_balance_measures: pd.DataFrame,
+    ):
+        """
+        Take the data balance measure dataframes, transform them into
```
romanlutz, line 168:
```suggestion
        Take the data balance measure DataFrames, transform them into
```

responsibleai/tests/databalanceanalysis/conftest.py
```diff
@@ -1,13 +1,15 @@
...
 
-import copy
-
 import pandas as pd
 import pytest
 
 from ..common_utils import create_adult_income_dataset
 
+SYNTHETIC_DATA_LABEL = "Label"
+GENDER = "Gender"
```
romanlutz, line 10: Technically, "adult" only has sex, not gender. Depending on where the data comes from that may not be represented correctly (sadly).

responsibleai/responsibleai/managers/data_balance_manager.py
ms-kashyap: Ya my mistake! Switched to normal quotes

responsibleai/responsibleai/databalanceanalysis/data_balance_helper.py
ms-kashyap: So right now, the `DataBalanceManager._validate` method raises an error if the target column contains != 2 unique values.
I thought it was good to do it this way rather than checking `task_type`. What do you think?

The snippet is:
```
if self._df[self._target_column].unique().size != 2:
            raise ValueError(
                (
                    f"The target_column '{self._target_column}' must contain"
                    " only a positive label and a negative label to compute"
                    " data balance measures."
                )
            )
```

responsibleai/tests/databalanceanalysis/conftest.py
ms-kashyap: Cleaning this up by updating the constant names and adding a small comment!

```
# Constants representing columns in the synthetic_data fixture
SYNTHETIC_DATA_LABEL = "Label"
SYNTHETIC_DATA_GENDER = "Gender"
SYNTHETIC_DATA_ETHNICITY = "Ethnicity"
```

responsibleai/responsibleai/databalanceanalysis/data_balance_helper.py
romanlutz: I think `task_type` is safer, although I would have agreed with you a couple of years ago 🙂  Since then I have run into too many situations where the test or training set somehow randomly ended up with only 1s or only 0s, so suddenly you have only 1 label and things fail.

responsibleai/tests/databalanceanalysis/conftest.py
romanlutz: It's synthetic? Adult is a real dataset, isn't it?

responsibleai/responsibleai/databalanceanalysis/data_balance_helper.py
ms-kashyap: Is it okay if I do both checks -- `task_type == classification` and `target_column.unique().size == 2`?
This way, it would specifically check for binary classification 
However, the user can have a dataset with all 0s or 1s in the label column even if it is a binary classification scenario... hmm, thoughts?

responsibleai/tests/databalanceanalysis/conftest.py
ms-kashyap: These constants are specifying col names for the synthetic data dataset in conftest.py!

Should I move these constants closer to the dataset instead of having at the very top?

responsibleai/responsibleai/databalanceanalysis/data_balance_helper.py
romanlutz: You can check for task_type and `size <= 2` 🙂

responsibleai/tests/databalanceanalysis/conftest.py
romanlutz: Oh I'm just surprised the adult dataset is synthetic. Are you quite sure?
https://archive.ics.uci.edu/ml/datasets/Adult

ms-kashyap: Sorry for the confusion 😁 the constants are being used to create this synthetic dataset in conftest:
https://github.com/microsoft/responsible-ai-toolbox/blob/1c6532c696e363e5c8499be1b3967861dc6f8b39/responsibleai/tests/databalanceanalysis/conftest.py#L85

responsibleai/responsibleai/databalanceanalysis/data_balance_helper.py
ms-kashyap: Done! 😊

romanlutz (approved):

responsibleai/responsibleai/databalanceanalysis/data_balance_helper.py
```diff
@@ -0,0 +1,200 @@
...
+
+    @staticmethod
+    def transform_feature_balance_measures(
+        df: pd.DataFrame,
+    ) -> Dict[str, Dict[str, Any]]:
+        """
+        Transform the feature balance measures DataFrame into a dictionary
+        acceptable by the RAI dashboard.
+
+        :param df: The feature balance measures df.
```
imatiach-msft, line 110: df -> DataFrame

responsibleai/responsibleai/databalanceanalysis/data_balance_helper.py
```diff
@@ -0,0 +1,200 @@
...
+AGGREGATE_BALANCE_MEASURES_KEY = "AggregateBalanceMeasures"
+
+
+class DataBalanceHelper:
+    """
+    Helper class for computing and transforming data balance measures
+    to be used in a dashboard.
+    """
+
+    @staticmethod
```
imatiach-msft, line 22: if these are all static methods, I would think it would be better to just have them as methods in the top-level file, as opposed to inside a helper class

responsibleai/responsibleai/managers/data_balance_manager.py
```diff
@@ -0,0 +1,336 @@
...
+        """
+        self._data_balance_measures = (
+            DataBalanceHelper.transform_measures_to_dict(
+                feature_balance_measures=feature_balance_measures,
+                distribution_balance_measures=distribution_balance_measures,
+                aggregate_balance_measures=aggregate_balance_measures,
+            )
+        )
+
+    def get(self):
```
imatiach-msft, line 200: why do we have both get and get_data?  I think maybe get can be removed?

ms-kashyap: The base class for managers (`BaseManager`) defines `get()` and `RAIBaseInsights` calls `.get()` for every manager, so I think it is required for me to implement `get()`!

As for `get_data()`, since data balance manager doesn't have its own `XYZData` class to return to `RAIInsights`, I'm just calling `get()` inside it so it can be used like this:

image

imatiach-msft: "since data balance manager doesn't have its own XYZData class to return to RAIInsights"
Yeah I guess that's my confusion, why it doesn't have such a class.  Maybe it's ok, just seems very different from the other managers.

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -370,6 +376,14 @@ def counterfactual(self) -> CounterfactualManager:
         """
         return self._counterfactual_manager
 
+    @property
+    def data_balance(self) -> DataBalanceManager:
```
xuke444, line 380: I just checked with pm
It is probably better to always have data balance instead of require user to call another add method.

romanlutz: How computationally intensive are the calculations?

responsibleai/responsibleai/managers/data_balance_manager.py
ms-kashyap: My thought process was that if the measures are a part of an `IDataset` and are a simple `dict`, then they don't need their own class. However, your point that every manager has its own `XYZData` class totally makes sense. Do you think I should introduce an `XYZData` class here?

responsibleai/responsibleai/rai_insights/rai_insights.py
ms-kashyap: Tagging @imatiach-msft and @PIC123 also.

Hi guys, I ran a small benchmark, and calculating measures is not very compute intensive, especially for datasets with <1M rows.
I ran these tests on my local machine (AMD Ryzen 5 3600 Processor, 16 GB RAM, M.2 SSD, has a GPU but not applicable):
```
Running data balance measures on 7 columns
Time taken for dataset with 48,842 rows: 1.5469965934753418 seconds
Time taken for dataset with 97,684 rows: 1.3919968605041504 seconds
Time taken for dataset with 195,368 rows: 1.6330010890960693 seconds
Time taken for dataset with 390,736 rows: 1.9849941730499268 seconds
Time taken for dataset with 781,472 rows: 2.7039999961853027 seconds
Time taken for dataset with 1,562,944 rows: 4.72900128364563 seconds
Time taken for dataset with 3,125,888 rows: 7.672998905181885 seconds
Time taken for dataset with 6,251,776 rows: 14.046992540359497 seconds
Time taken for dataset with 12,503,552 rows: 26.853999853134155 seconds
Time taken for dataset with 25,007,104 rows: 75.52636933326721 seconds
```

---

Pasting my script for future references:

```
import time
from responsibleai.managers.data_balance_manager import DataBalanceManager


# from https://publictestdatasets.blob.core.windows.net/data/adult-train.csv and adult-test.csv
train_df, test_df = train_data, test_data 
target_feature = 'income'
categorical_features = ['education', 'marital-status',
                        'occupation', 'relationship', 'race', 'gender', 'native-country']

print(f"Running data balance measures on {len(categorical_features)} columns")

for i in range(10):
    t0 = time.time()

    manager = DataBalanceManager(
        train=train_df,
        test=test_df,
        target_column=target_feature,
        task_type="classification",
    )
    manager.add(cols_of_interest=categorical_features)
    manager.compute()
    
    t1 = time.time()
    num_rows = manager._df.shape[0]
    print(f"Time taken for dataset with {num_rows:,} rows: {t1 - t0} seconds")

    # Double the size of each dataframe
    train_df = pd.concat([train_df, train_df])
    test_df = pd.concat([test_df, test_df])
```

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -148,6 +150,23 @@ def _get_classes(task_type, train, target_column, classes):
         else:
             return None
 
+    def _try_add_data_balance(self):
```
ms-kashyap, line 153: This logic automatically adds data balance measures (`compute()` still has to be called) if the right conditions are met when user creates an `RAIInsights()` object.

@imatiach-msft @xuke444 @PIC123 Let me know what you think of this change and how we can re-shape it if it's too complicated 😊

xuke444 (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
ms-kashyap: @imatiach-msft @romanlutz @memoryz @martthalch 

Calling out two big changes that I discussed with @xuke444 and @PIC123:

### Change 1

Instead of the user adding data balance the same way they would for managers:
```
rai_insights = RAIInsights(...)
rai_insights.data_balance.add(cols_of_interest=["col1", "col2"])
rai_insights.compute()
```

We moved towards adding data balance when user creates `RAIInsights`. Compute still needs to be called.
```
rai_insights = RAIInsights(...)
rai_insights.compute()
```

If the following criteria is met:
1. `task_type == "classification"`
2. `len(categorical_features) > 0`
3. target column `classes` is not null

Then data balance measures will be automatically added for the given `categorical_features` list.

Also, user no longer has access to `rai_insights.data_balance` property
But they can still set their own cols_of_interest via
```
rai_insights._data_balance_manager.add(...)  <-- overwrites the automatic add from RAIInsights constructor
```

---

### Change 2

Initially, the user was required to pass in the positive label for their dataset. For example, if the target column contains {<=50k, >50k}, feature balance measures need to know which of these is considered the positive label to compute measures.

Now, feature balance measures are being computed for all values in the target column. That way, user can choose in the UI which label is considered the positive label.

image

This also makes it so that data balance can work for all classification scenarios (both binary and multiclass) instead of just binary like before. 😁

imatiach-msft (approved):
LGTM

PR #1487, opened 2022-06-10
Sort individual feature importance by residual for regression

tongyu-microsoft:
This PR orders the samples in individual feature importance by abs(true_y - predicted_y) for regression. It will show the predictions which are better first than the prediction which are way off in the model.

## Description
Before: sort by index
image

After: sort by abs(true_y - predicted_y)
image

## Checklist

### DIFF
--- a/libs/core-ui/src/lib/util/JointDataset.ts
+++ b/libs/core-ui/src/lib/util/JointDataset.ts
@@ -81,6 +81,7 @@ export class JointDataset {
   public static readonly PredictedYLabel = "PredictedY";
   public static readonly ProbabilityYRoot = "ProbabilityClass";
   public static readonly TrueYLabel = "TrueY";
+  public static readonly Residual = "Residual";
   public static readonly DitherLabel = "Dither";
   public static readonly DitherLabel2 = "Dither2";
   public static readonly ClassificationError = "ClassificationError";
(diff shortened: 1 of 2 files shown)


gaugup (approved):

imatiach-msft (approved):

romanlutz:
@tongyu-microsoft I'm curious what prompted this change. Was this a feature request from somebody?
As a user, I'd be a bit surprised about the ordering since we had it by index so far and that's immediately obvious. The residual ordering is not immediately obvious unless we point it out in writing. Perhaps that should be added? If you haven't yet you may want to consult with a designer 🙂

gaugup:
> @tongyu-microsoft I'm curious what prompted this change. Was this a feature request from somebody? As a user, I'd be a bit surprised about the ordering since we had it by index so far and that's immediately obvious. The residual ordering is not immediately obvious unless we point it out in writing. Perhaps that should be added? If you haven't yet you may want to consult with a designer 🙂

@romanlutz, I think sorting based on some regression metric might be useful for understanding as to why model might be predicting with higher accuracy for some samples and doesn't predict the same way for other samples. For classification scenario as well, we differentiate between which samples were correctly predicted by model vs which were not correctly predicted. Agree, that we should document this in the dashboard and should use some standard regression metric like r2_score to order these.

imatiach-msft:
I think it would also be nice to see the values by which these are sorted as a separate column.  If they are sorted by abs(true_y - predicted_y), it might be nice to have a columns with name something like "Abs difference" right after the index, so it's clear that this is what the data is sorted by.  It doesn't necessarily have to be part of this PR though.

PR #1496, opened 2022-06-13
Model overview: add cohort selection hint

romanlutz:
## Description

So far we didn't give users any kind of hint as to what the cohort selection flyout does or how they are supposed to interact with it:
image

This PR changes that! Below you see the strings at the top (above the choice group and dropdown). The two pictures show the same strings, but I thought it makes sense to see it in both configurations.

Default without features selected:
image

With feature selected:
image

## Checklist


xuke444 (approved):

libs/localization/src/lib/en.json
```diff
@@ -1438,6 +1438,10 @@
         "metricNameColumnHeader": "Metric name",
         "metricDescriptionColumnHeader": "Metric Description"
       },
+      "cohortSelection": {
+        "flyoutHeader": "Choose cohorts",
+        "flyoutDescription": "You can choose to view dataset cohorts or feature-based cohorts. To generate feature-based cohorts you need to first select one or more features in the feature cohorts view."
```
imatiach-msft, line 1443: "you need to first select"
This text feels incomplete because it doesn't state what needs to be done after selecting the features.  It should be a completed idea by then stating something like "Then, select this <...> button which will enable <...> view."

romanlutz: That's fair. How about this:
```
        "flyoutDescription": "You can choose to view dataset cohorts or feature cohorts. If feature cohorts are unavailable you need to first select one or more features in the feature cohorts view. Subsequently, feature cohorts are generated and you can select them here."
```

imatiach-msft: I think that looks much better, it's clearer and doesn't leave the reader thinking about what's next

gaugup (approved):

PR #1498, opened 2022-06-14
Remove download svg option due to highcharts limitations

tongyu-microsoft:
This PR removes 

## Checklist

### DIFF
--- a/libs/core-ui/src/lib/Highchart/getDefaultHighchartOptions.ts
+++ b/libs/core-ui/src/lib/Highchart/getDefaultHighchartOptions.ts
@@ -34,6 +34,20 @@ export function getDefaultHighchartOptions(theme: ITheme): Highcharts.Options {
       zoomType: "xy"
     },
     credits: undefined,
+    exporting: {
+      buttons: {
+        contextButton: {
+          menuItems: [
+            "viewFullscreen",
+            "printChart",
+            "separator",
+            "downloadPNG",
+            "downloadJPEG",
+            "downloadPDF"
+          ]
+        }
+      }
+    },
     legend: {
       enabled: false
     },


libs/core-ui/src/lib/Highchart/getDefaultHighchartOptions.ts
```diff
@@ -34,6 +34,20 @@ export function getDefaultHighchartOptions(theme: ITheme): Highcharts.Options {
...
+    exporting: {
+      buttons: {
+        contextButton: {
+          menuItems: [
+            "viewFullscreen",
+            "printChart",
+            "separator",
+            "downloadPNG",
+            "downloadJPEG",
+            "downloadPDF"
```
gaugup, line 46: Can we add a test for testing the download of different file types? It this seems out of scope could we create an issue to track this?

imatiach-msft (approved):

romanlutz (approved):
This is a great find! Is this a known bug in highcharts? If not, should we file it?

No objection to your change, of course!

libs/core-ui/src/lib/Highchart/getDefaultHighchartOptions.ts
tongyu-microsoft: Thanks for the comment. I created a task and assigned to myself: https://msdata.visualstudio.com/Vienna/_workitems/edit/1830300
Will work on it later

tongyu-microsoft:
A workaround: https://jsfiddle.net/oysteinmoseng/yezu6nx3/
The latest version of Highcharts has the fix for this svg export issue: https://github.com/highcharts/highcharts/issues/16589

PR #1500, opened 2022-06-14
Localized file check-in by OneLocBuild Task: Build definition ID 21368: Build ID 65085801

riedgar-ms:
This is the pull request automatically created by the OneLocBuild task in the build process to check-in localized files generated based upon translation source files (.lcl files) handed-back from the downstream localization pipeline. If there are issues in translations, visit https://aka.ms/ceLocBug and log bugs for fixes. The OneLocBuild wiki is https://aka.ms/onelocbuild and the localization process in general is documented at https://aka.ms/AllAboutLoc.


romanlutz (approved):

romanlutz:
Some of these are actually erasing the existing translation in favor of the original English string in localized files. Is it fair to assume that they will be localized again at a later point?

riedgar-ms:
> Some of these are actually erasing the existing translation in favor of the original English string in localized files. Is it fair to assume that they will be localized again at a later point?

Let me ask

imatiach-msft (approved):

imatiach-msft:
"Some of these are actually erasing the existing translation in favor of the original English" yes that is strange, huh

PR #1505, opened 2022-06-16
Unify naming between notebooks and test cases

romanlutz:
## Description

This PR aligns naming between notebooks and test cases. It's a prereq for the e2e-with-flighting tests I've written in #1467. This is another change refactored out of #1467 to make that PR more digestable. What does that mean in detail?
- `ClassificationModelDebugging` -> `CensusClassificationModelDebugging`
- `*modelStatistics` -> `*modelOverview`
- `casualAnalysis` -> `causalAnalysis` (love the typo, though! 😆 )

## Checklist


imatiach-msft:
"casualAnalysis -> causalAnalysis (love the typo, though! 😆 )"
hehe very funny typo indeed 😆, I guess casual analysis is when you just sometimes want to do some spontaneous analysis but aren't really serious about it?? But we are very serious about our analyses in our dashboards 🥽🥼, perhaps we should call it a "serious analysis" instead.

libs/e2e/src/index.ts
```diff
@@ -6,6 +6,7 @@ export * from "./lib/describer/modelAssessment/causalAnalysis/describeCausalAnal
 export * from "./lib/describer/modelAssessment/dataExplorer/describeDatasetExplorer";
 export * from "./lib/describer/modelAssessment/errorAnalysis/describeErrorAnalysis";
 export * from "./lib/describer/modelAssessment/featureImportances/individualFeatureImportance/describeIndividualFeatureImportance";
-export * from "./lib/describer/modelAssessment/modelStatistics/describeModelStatistics";
+export * from "./lib/describer/modelAssessment/modelOverview/describeModelOverview";
```
vinuthakaranth, line 9: just a fyi, this has to be updated in AML once we release and this change is being used in AML.

vinuthakaranth (approved):

imatiach-msft:
even in the econml docs they seem to make the same spelling mistakes:
https://econml.azurewebsites.net/spec/estimation/deepiv.html
"What is required is a vector of instruments Z, assumed to **casually** affect the distribution of the treatment T"

gaugup (approved):

libs/e2e/src/index.ts
romanlutz: Good point! Need to keep that on our radar for when we do that. Or do you think I should proactively go and upgrade the version Studio uses and make this change?

vinuthakaranth: You can update this once we release new version with this code change, that will be fine.

PR #1506, opened 2022-06-16
Upgrade highcharts from v9.3.0 to v10.1.0, remove highcharts bundle by setting it to external module

tongyu-microsoft:
This PR upgrades highcharts from v9.3.0 to v10.1.0, removes highcharts bundle by setting it to external module.

## Description
The reason we want to upgrade highcharts version is, there is a bug on v9.3.0 where exporting svg file is broken. This bug is fixed after v10.0.0. So here we upgrade highcharts to 10.1.0. 

## Checklist

### DIFF
--- a/libs/core-ui/src/lib/Highchart/getDefaultHighchartOptions.ts
+++ b/libs/core-ui/src/lib/Highchart/getDefaultHighchartOptions.ts
@@ -34,20 +34,6 @@ export function getDefaultHighchartOptions(theme: ITheme): Highcharts.Options {
       zoomType: "xy"
     },
     credits: undefined,
-    exporting: {
-      buttons: {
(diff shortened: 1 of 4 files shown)


libs/core-ui/src/lib/Highchart/getDefaultHighchartOptions.ts
```diff
@@ -34,20 +34,6 @@ export function getDefaultHighchartOptions(theme: ITheme): Highcharts.Options {
       zoomType: "xy"
     },
     credits: undefined,
-    exporting: {
```
tongyu-microsoft, line 37: We use default exporting which includes all options below as well as downloadSVG

imatiach-msft (approved):

PR #1508, opened 2022-06-20
Model overview: use spline instead of line chart

romanlutz:
## Description

This PR replaces the line chart with a spline chart to make it more pleasant to look at. The only real change is that the chart option is changed from `"line"` to `"spline"`, the rest are renaming changes.

Old experience:

image

New experience:

image

## Checklist


imatiach-msft:
@romanlutz  what do you think about giving them the option of doing both?  The spline can be misleading in many cases.

tongyu-microsoft (approved):

xuke444 (approved):

PR #1511, opened 2022-06-21
Data Balance UX Part 1: add data_balance_measures to IDataset

ms-kashyap:
## Description

This introduces the data schema of data balance measures.

## Checklist


libs/core-ui/src/lib/Interfaces/IDataset.ts
```diff
@@ -11,6 +13,7 @@ export interface IDataset {
   categorical_features: string[];
   class_names?: string[];
   target_column?: string;
+  data_balance_measures?: IDataBalanceMeasures;
```
romanlutz, line 16: More of a conceptual question: Are these only going to be shown for the dataset as a whole? What if we have a global cohort that shows only a subset of the data? Does it show the data balance measures for that subset, or still globally?
From the adultCensus.ts file I'm gathering that these are all things that are precomputed, which would indicate that we don't show this per-cohort, but rather for the whole dataset. I do feel like this would totally be possible and perhaps be a more natural fit, but then I think that's more of a PM/design question and should go to the people who make such decisions as they probably have thought about it more than me 🙂

libs/core-ui/src/lib/Interfaces/IDataBalanceMeasures.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export interface IFeatureBalanceMeasures {
+  [key: string]: IFeatureBalanceMeasure[];
+}
+
+export interface IFeatureBalanceMeasure {
+  ClassA: string;
+  ClassB: string;
+  StatisticalParity: number;
```
romanlutz, line 38: That one is just for me, but I would love to see the definition 🙂

libs/core-ui/src/lib/Interfaces/IDataset.ts
ms-kashyap: You are right: the measures are currently precomputed using the whole dataset for whatever columns user specifies in `columns_of_interest`!

I think it could also be possible to compute per-cohort, provided that the dataset isn't so big that it slows things down. You saw from the Python module that the computation logic is fairly simple -- lots of `groupby`s to compute `P(feature), P(feature and positive label), P(positive label)` etc. (atleast for feature balance measures)!

I think these per-dataset insights are are the first step to something concrete, and I can totally see it working per-cohort like the rest of the dashboard.

libs/core-ui/src/lib/Interfaces/IDataBalanceMeasures.ts
ms-kashyap: The definition will be in part 2, but [here is a sneak peek](https://github.com/ms-kashyap/responsible-ai-widgets/pull/5/files#diff-a95892d97ed2f98a5605dbf0300e2a67fd753a8653b71a49336084db08f74fff) to the definition from the whole feature PR in my own fork 😊

Please feel free to put comments there if you'd like!

libs/core-ui/src/lib/Interfaces/IDataset.ts
ms-kashyap: Good point! Tagging @memoryz, @imatiach-msft, and my PM @martthalch for your guys' thoughts 😊

romanlutz (approved):
I don't have further comments other than what I pointed out already. I strongly suggest waiting for @imatiach-msft  and @xuke444 to review. @gaugup may also be interested but I he may not be able to review in the next few days.

libs/core-ui/src/lib/Interfaces/IDataset.ts
PIC123: I agree with all that was said above. I think we'll focus on just supporting all data since there is a lot more to consider when adding cohort support from what I understand. And yah, making sure that its clear that cohorts aren't supported, maybe with some UI updates to provide context or something like that

imatiach-msft: +1 I agree with the thoughts above.  I would suggest to get this working for all data first before trying to do it for all cohorts - and we also may indeed want to limit this in some scenarios, like big data, to only compute over all data and not at cohort level in the future.

libs/core-ui/src/lib/Interfaces/IDataBalanceMeasures.ts
```diff
@@ -0,0 +1,46 @@
...
+export interface IDistributionBalanceMeasures {
+  [key: string]: IDistributionBalanceMeasure;
+}
+
+export interface IDistributionBalanceMeasure {
+  ChiSquarePValue: number;
+  ChiSquareStat: number;
+  InfiniteNormDist: number;
+  JensenShannonDist: number;
+  KLDivergence: number;
```
imatiach-msft, line 25: do we really need to compute so many data balance measures for the dataset?  I'm just wondering if this might become too slow in some cases, and if it may be useful to allow users to only compute a subset of these - in which case perhaps instead of this interface something more dynamic like a map should be used from balance measure name to value.  Not meant to block this PR though, just something to think about.  Do any of these measures become much slower for larger data in terms of # of rows or # of features?

imatiach-msft (approved):

libs/core-ui/src/lib/Interfaces/IDataBalanceMeasures.ts
ms-kashyap: Map seems like a really good idea. That way, we can add/remove measures in the future with less code changes. Both UX and Python side would be touched, but atleast the interface would not be touched.

Also, I did a small benchmark in the Python side PR so we can better understand the compute time of various dataset sizes, [adding the link here](https://github.com/microsoft/responsible-ai-toolbox/pull/1485#discussion_r914225527).

ms-kashyap: I had an easy time doing this for `IAggregateBalanceMeasures` and `IDistributionBalanceMeasure` since their interfaces only contain str -> number pairs.

For `IFeatureBalanceMeasure`, since each measure has both str -> number and str -> str pairs:
```
{
  ClassA: "Bachelors",
  ClassB: "HS-grad",
  JaccardIndex: 0.1009922102843339,
  KendallRankCorrelation: 1.9020631282523714,
  LogLikelihoodRatio: 0.27920025739587184,
  PointwiseMutualInfo: 0.9519638161627417,
  SorensonDiceCoeff: 0.07642301570298538,
  StatisticalParity: 0.25222421370244086,
  TTest: 0.4236184547392873,
  TTestPValue: 0.342278369038513
},
```
I'm having a hard time defining the interface to where both `ClassA` `ClassB` and the measures can be specified. Is there any way I can do this without changing the schema of what the Python side creates?

image

ms-kashyap: I was able to fix it by allowing the indexable type to also returning a `string`

```
export interface IFeatureBalanceMeasure {
  [key: string]: number | string;
  ClassA: string;
  ClassB: string;
}

```

ms-kashyap: Updated the PR to use indexable type instead of hardcoded measure names .

I'm a huge fan of this change, thank you for the suggestion Ilya! 😊

I'm resolving this thread and we can talk about compute time / slowness [in the other PR](https://github.com/microsoft/responsible-ai-toolbox/pull/1485#discussion_r914225527).

xuke444 (approved):

imatiach-msft (approved):
LGTM

PR #1512, opened 2022-06-22
Data Balance UX Part 2: Add visualizations as a flight

ms-kashyap:
## Description

This PR requires PR #1511 to be completed first since it relies on interfaces from that PR.

This PR introduces the data balance visualizations as a flight (Thank you for the idea @romanlutz)!
image

Without flighting, the Data explorer looks like the usual:
image

If user selects the data balance flight, the Data explorer becomes a parent tab with two pivot items:
image

---

In the Data balance tab, there are 3 sets of measures:

image
For feature balance measures, the heatmap updates based on the feature and measure selected on the top left.
For every feature, there is a description of what it means.

image
For distribution balance measures, each measure has its own range therefore each measure has its own subplot. The user can hide / view subplots via the legend. There is a small "i" icon next to the measure name which gives a description of the measure.

Having multiple visible/hidden subplots in Highcharts is complicated, so I added a lot of comments in this logic.

image
For aggregate balance measures, these measures are hard to visualize in a column or bar chart (or any other options Highcharts has), so I am keeping it simple with a DetailsList. 

---

When we are able to add mitigations (i.e. provide actionable steps to user based on data imbalance), these 3 charts will need to update to provide that information (maybe as a list of actionable items below each chart).

I am correctly working on e2e tests for the UX, but I am holding off on writing too many tests until the team is OK with the UX itself.

Localization for all strings has been added, and this can easily be changed depending on the proper terminology we want to use 😊

## Checklist


romanlutz:
image

The "Data explorer" header followed by "Data explorer" pivot tab is a bit redundant. I wish I had a good suggestion for an alternative...

libs/dataset-explorer/src/lib/DataExplorerParentTab.tsx
```diff
@@ -0,0 +1,90 @@
...
+interface IDataExplorerParentTabState {
+  activeViewOption: DataExplorerParentTabOptions;
+}
+
+enum DataExplorerParentTabOptions {
+  DatasetExplorer = "DatasetExplorer",
+  DataBalance = "DataBalance"
+}
+
+export class DataExplorerParentTab extends React.Component<
```
romanlutz, line 25: Hmm, I'm not a fan of the naming here. If the overall component remains "Data explorer" then `DataExplorerTab` should be the parent (without further mentioning it in the name). I guess this goes back to the same naming conondrum I mentioned in the other comment 🙁

libs/dataset-explorer/src/lib/DataExplorerParentTab.tsx
```diff
@@ -0,0 +1,90 @@
...
+          </Stack>
+        </Stack.Item>
+        <Stack.Item>
+          {this.state.activeViewOption ===
+            DataExplorerParentTabOptions.DatasetExplorer && (
+            <DatasetExplorerTab />
+          )}
+          {this.state.activeViewOption ===
+            DataExplorerParentTabOptions.DataBalance && <DataBalanceTab />}
+        </Stack.Item>
```
romanlutz, line 74: Any particular reason for rendering these contents separately? You can do this:

```
<PivotItem ...>
    <DatasetExplorerTab />
</PivotItem>
<PivotItem ...>
    <DatasetBalanceTab />
</PivotItem>
```

libs/dataset-explorer/src/lib/DistributionBalanceMeasuresChart.tsx
```diff
@@ -0,0 +1,72 @@
...
+export interface IDistributionBalanceMeasureProps {
+  distributionBalanceMeasures: IDistributionBalanceMeasures;
+}
+
+export class DistributionBalanceMeasuresChart extends React.PureComponent<IDistributionBalanceMeasureProps> {
+  public render(): React.ReactNode {
+    const distributionBalanceMeasures = this.props.distributionBalanceMeasures;
+    if (!distributionBalanceMeasures) {
+      return;
+    }
```
romanlutz, line 25: Should we point out to users that they'd need to upload these / pass them in to view this type of chart? Particularly, already existing uploaded dashboard data won't include this so it's a case users will actually hit.

libs/dataset-explorer/src/lib/DistributionBalanceMeasuresChart.tsx
```diff
@@ -0,0 +1,72 @@
...
+          <Stack horizontal>
+            <Stack.Item>
+              <Text variant="large" className={styles.boldText}>
+                {measuresLocalization.Name}
+              </Text>
+            </Stack.Item>
+
+            <Stack.Item className={styles.callout}>
+              <LabelWithCallout
+                label=""
```
romanlutz, line 44: Was something meant to be here?

libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
```diff
@@ -0,0 +1,232 @@
...
+                  // TODO: Replace link with https://responsibleaitoolbox.ai/ link once docs are published there
+                  href="https://microsoft.github.io/SynapseML/docs/features/responsible_ai/Data%20Balance%20Analysis/#feature-balance-measures"
+                  target="_blank"
+                >
+                  {localization.ModelAssessment.DataBalance.LearnMore}
+                </Link>
+              </LabelWithCallout>
+            </Stack.Item>
+          </Stack>
+        </Stack.Item>
```
romanlutz, line 92: It feels like I just read this in the previous file. Can we refactor this into its own React component and have the specific parts be props?

libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
```diff
@@ -0,0 +1,232 @@
...
+                    {measuresLocalization.MeasurePicker}
+                  </Text>
+                </Stack.Item>
+                <Stack.Item>
+                  <Dropdown
+                    styles={{
+                      callout: {
+                        selectors: {
+                          ".ms-Button-flexContainer": {
+                            width: "100%"
```
romanlutz, line 142: I am not quite familiar with this. Can you explain what it does?

libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
```diff
@@ -0,0 +1,232 @@
...
+                    </Text>
+                    <Text variant="mediumPlus" className={styles.boldText}>
+                      {` ${selectedMeasure} `}
+                    </Text>
+                    <Text variant="mediumPlus">
+                      {measuresLocalization.Description2}
+                    </Text>
+                    <Text variant="mediumPlus" className={styles.boldText}>
+                      {` ${selectedFeature}`}
+                    </Text>
```
romanlutz, line 181: I see the problem you faced here. Are we sure that the word order will translate the same way to all other supported languages? If not, then localization will make this very weird. I can't say I have a good solution. @xuke444 maybe?

libs/dataset-explorer/src/lib/getDistributionBalanceMeasuresChart.ts
```diff
@@ -0,0 +1,261 @@
...
+      // axis.left defines how far from the left to place the axis
+      // axis.width defines the width of the axis relative to the entire chart
+      xAxes.push({
+        categories: features,
+        left: hiddenCharts.has(measureName) ? "0%" : `${axisLeftStart}%`,
+        offset: 0,
+        showEmpty: false,
+        title: {
+          // If user hovers over the title, they are presented with the measure description.
+          text: `<div title="${measureInfo.Description}">${measureName} ${infoIcon}</div>`,
```
romanlutz, line 144: This seems a little bit hacky, but I've faced similar issues with highcharts a few times. There's no way to do this with, say, a formatting function arg, is there?

libs/dataset-explorer/src/lib/getDistributionBalanceMeasuresChart.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  // If the clicked-on chart was already hidden, that means user clicked to show it, so remove it from hidden charts.
+  if (hiddenCharts.has(e.target.name)) {
+    hiddenCharts.delete(e.target.name);
+  } else {
+    // Otherwise, the user clicked to hide the chart so add it to hidden charts.
+    hiddenCharts.add(e.target.name);
+  }
+
+  // Recompute # of visible charts and the width of each subplot after updating list of hiddden charts.
```
romanlutz, line 230:
```suggestion
  // Recompute # of visible charts and the width of each subplot after updating list of hidden charts.
```

libs/dataset-explorer/src/lib/getDistributionBalanceMeasuresChart.ts
```diff
@@ -0,0 +1,261 @@
...
+  const numVisibleCharts =
+    DistributionBalanceMeasuresMap.size - hiddenCharts.size;
+  const width = 100 / numVisibleCharts;
+
+  let axisLeftStart = 0;
+  e.target.chart.xAxis.forEach((xAxis, i) => {
+    const yAxis = e.target.chart.yAxis[i];
+
+    // For each chart, if its title is in the hidden charts list, hide it.
+    if (hiddenCharts.has(xAxis.options.title?.text ?? "")) {
```
romanlutz, line 240:
```suggestion
    if (xAxis.options.title?.text && hiddenCharts.has(xAxis.options.title?.text)) {
```
Wdyt?

libs/dataset-explorer/src/lib/getFeatureBalanceMeasuresChart.ts
```diff
@@ -0,0 +1,209 @@
...
+      ) {
+        // If a measure value is found for ClassB vs. ClassA, flip its value
+        // since every feature balance measure gap is symmetric
+        const measureValue = row[measureName] * (foundFlipped ? -1 : 1);
+        items.push([rowIndex, colIndex, measureValue]);
+      }
+    });
+  });
+
+  // Chosen color scale is "viridis", the default color scale for matplotlib, because it is:
```
romanlutz, line 138: Does this mesh well with the color scheme we use in AML? For model overview I was asked to use blue tones.

libs/localization/src/lib/en.json
```diff
@@ -1312,6 +1313,141 @@
...
+          },
+          "TheilTIndex": {
+            "Description": "Theil's T Index is a measure of income inequality that is calculated using the Generalized Entropy Index with α = 1, and therefore it is more sensitive to differences at the higher end of the distribution. This index better captures record combinations that occur very often in the dataset.",
+            "Name": "Theil Index (T)"
+          }
+        },
+        "Name": "Aggregate Balance Measures",
+        "Table": {
+          "Description": "Description",
+          "FeatureName": "Feature Name",
```
romanlutz, line 1339: General point in this file: Should the second (and further) words be capitalized? I think we're not doing that elsewhere. So that would mean "Feature name", "Feature value", "Aggregate balance measures", etc.

libs/localization/src/lib/en.json
```diff
@@ -1312,6 +1313,141 @@
...
+        "Name": "Distribution Balance Measures"
+      },
+      "FeatureBalanceMeasures": {
+        "Callout": {
+          "Description": "Feature balance measures allow us to see whether each combination of feature values is receiving the positive outcome (true prediction) at balanced probability.",
+          "Title": "What are feature balance measures?"
+        },
+        "Chart": {
+          "Axes": {
+            "ClassA": "Class A",
```
romanlutz, line 1399: Should this not be the actual name of the feature?

libs/localization/src/lib/en.json
```diff
@@ -1312,6 +1313,141 @@
...
+            "ClassB": "Class B"
+          },
+          "Tooltip": {
```
romanlutz, line 1402: Can you share a screenshot showing the tooltips?

PIC123:
> image
> 
> The "Data explorer" header followed by "Data explorer" pivot tab is a bit redundant. I wish I had a good suggestion for an alternative...

Could we maybe switch the section to be called Data Analysis , with the Data explorer tab always there? Not sure if that's more confusing, but we could maybe add a place holder saying that other features are coming soon?

libs/dataset-explorer/src/lib/DataExplorerParentTab.tsx
ms-kashyap: I didn't even realize you can do this! Fixed 😊

- Removed `IDataExplorerParentTabState` completely
- Removed use of StackItems in favor of PivotItems
- Removed `onViewTypeChange` since we are no longer keeping state
- Added a `pivotLabelWrapper` style

libs/dataset-explorer/src/lib/DataBalanceTab.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+  callout: IStyle;
+  infoWithText: IStyle;
+  page: IStyle;
+}
+
+export const dataBalanceTabStyles: () => IProcessedStyleSet<IDataBalanceTabStyles> =
+  () => {
+    const theme = getTheme();
+    return mergeStyleSets<IDataBalanceTabStyles>({
+      boldText: {
```
ms-kashyap, line 23: Try office ui fabric label

xuke444 (approved):

libs/dataset-explorer/src/lib/DistributionBalanceMeasuresChart.tsx
ms-kashyap: No, its empty because I was trying to make an empty `LabelWithCallout` so that just the clickable (i) appears. Not sure if `Callout` would work better here, but this was working well for the use case.

libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
ms-kashyap: Great suggestion, done! :)

libs/localization/src/lib/en.json
ms-kashyap: For sure! 

ClassA = Doctorate and ClassB = HS-grad so the format becomes:
`<Class A> and <Class B> have a <Measure> of <Value>`
image

libs/localization/src/lib/en.json
ms-kashyap: These are the x- and y- axes labels:
image

libs/dataset-explorer/src/lib/getFeatureBalanceMeasuresChart.ts
ms-kashyap: @PIC123 Do you have any thoughts on the color scheme? Do you know if AzureML would be okay with my chosen color scheme?

Also, since we are adding this as a flight, how about we keep it to viridis, see if AzureML + users like it, and then make the choice?

ms-kashyap:
> > image
> > The "Data explorer" header followed by "Data explorer" pivot tab is a bit redundant. I wish I had a good suggestion for an alternative...
> 
> Could we maybe switch the section to be called Data Analysis , with the Data explorer tab always there? Not sure if that's more confusing, but we could maybe add a place holder saying that other features are coming soon?

Love the idea of the section being called "Data analysis" with "Data explorer" and "Data balance" being tabs of this section.

For the default experience, I tried adding an empty "Coming soon" tab but I think the user will be tempted to click on it

@PIC123 @romanlutz What do you guys think of this:

No flight chosen (default)
image

Data balance flight chosen
image

libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
ms-kashyap: @xuke444 Do you happen to have an idea for how to tackle this?

libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+            {/* Renders the description of what user chose from dropdowns */}
+            <Stack.Item>
+              <Stack tokens={{ childrenGap: "s1" }}>
+                <Stack.Item>
+                  {/* Empty placeholder so that description lines up with dropdowns */}
+                  <br />
+                </Stack.Item>
+                <Stack.Item>
+                  <Text>
```
imatiach-msft, line 195: nit: I would recommend using Stack, and putting other Text inside the Stack.  Putting Text inside Text seems a bit strange to me.

libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
```diff
@@ -0,0 +1,276 @@
...
+                    <Text variant="mediumPlus">
+                      {measuresLocalization.Description1}
+                    </Text>
+                    <Text variant="mediumPlus" className={styles.boldText}>
+                      {` ${selectedMeasure} `}
+                    </Text>
+                    <Text variant="mediumPlus">
+                      {measuresLocalization.Description2}
+                    </Text>
+                    <Text variant="mediumPlus" className={styles.boldText}>
```
imatiach-msft, line 207: for boldText you can use Label instead of Text

libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
```diff
@@ -0,0 +1,276 @@
...
+            </Stack.Item>
+          </Stack>
+        </Stack.Item>
+
+        {/* Renders the description of the chosen measure */}
+        <Stack.Item>
+          {/* Since measure name is bolded, the <Text>s are separate and form "<measureName> <measureDesc>" */}
+          <Text variant="medium" className={styles.infoWithText}>
+            <Text className={styles.boldText}>{selectedMeasure}</Text>
+            <Text> </Text>
```
imatiach-msft, line 228: nit: looks like this is empty? remove?

libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
```diff
@@ -0,0 +1,276 @@
...
+        </Stack.Item>
+      </Stack>
+    );
+  }
+
+  private setSelectedLabel = (
+    _: React.FormEvent<HTMLDivElement>,
+    item?: IDropdownOption
+  ): void => {
+    if (item?.key !== undefined) {
```
imatiach-msft, line 254: we try not to use undefined, so perhaps:
```
if (!item?.key) {
```

imatiach-msft: same below on lines 263, 272

libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
```diff
@@ -0,0 +1,276 @@
...
+                    options={measureOptions}
+                    selectedKey={this.state.selectedMeasureIndex}
+                    onChange={this.setSelectedMeasure}
+                  />
+                </Stack.Item>
+              </Stack>
+            </Stack.Item>
+
+            {/* Renders the description of what user chose from dropdowns */}
+            <Stack.Item>
```
imatiach-msft, line 188: nit: can you refactor this out into a separate description component?  we try to keep files under 300 lines or even smaller if possible.  I think this could be its own class.

libs/dataset-explorer/src/lib/getFeatureBalanceMeasuresChart.ts
```diff
@@ -0,0 +1,219 @@
...
+
+export function getFeatureBalanceMeasuresChart(
+  targetColFeatureBalanceMeasures: ITargetColumnFeatureBalanceMeasures,
+  selectedLabel: string,
+  selectedFeature: string,
+  selectedMeasure: string
+): IHighchartsConfig {
+  const measureInfo = FeatureBalanceMeasuresMap.get(selectedMeasure);
+  if (
+    targetColFeatureBalanceMeasures === undefined ||
```
imatiach-msft, line 97: nit:
```
!targetColFeatureBalanceMeasures ||
...
!measureInfo
```

libs/dataset-explorer/src/lib/getFeatureBalanceMeasuresChart.ts
imatiach-msft: I think this is fine for now

libs/e2e/src/lib/describer/modelAssessment/Constants.ts
```diff
@@ -131,5 +131,30 @@ export enum Locators {
   ModelOverviewMetricChart = "#ModelOverview #ModelOverviewMetricChart",
   ModelOverviewMetricChartBars = "#ModelOverview #ModelOverviewMetricChart .highcharts-bar-series > rect",
   MissingParameterPlaceholder = "#ModelOverview #MissingParameterPlaceholder",
-  CounterfactualHeader = "#ModelAssessmentDashboard span:contains('What-If counterfactuals')"
+  CounterfactualHeader = "#ModelAssessmentDashboard span:contains('What-If counterfactuals')",
```
imatiach-msft, line 134: great to see these tests!

libs/e2e/src/lib/describer/modelAssessment/dataBalance/describeDataBalance.ts
```diff
@@ -0,0 +1,51 @@
...
+        });
+        cy.task("log", hostDetails.host);
+        cy.visit(hostDetails.host);
+      }
+      cy.get("#ModelAssessmentDashboard").should("exist");
+      cy.get(Locators.DataAnalysisPivot).should("exist");
+      cy.get(Locators.DataBalancePivotItem).click();
+    });
+    if (
+      !datasetShape.dataBalanceData?.aggregateBalanceMeasuresComputed &&
```
imatiach-msft, line 36: nit: maybe refactor out datasetShape.dataBalanceData into its own variable

libs/e2e/src/lib/describer/modelAssessment/dataBalance/describeDistributionBalanceMeasures.ts
```diff
@@ -0,0 +1,52 @@
...
+        4
+      );
+    });
+
+    it("Should have 8 subplots, each having a title for the measure", () => {
+      cy.get(Locators.DistributionBalanceMeasuresXAxes).should(
+        "have.length",
+        8
+      );
+      cy.get(Locators.DistributionBalanceMeasuresXAxes).should(
```
imatiach-msft, line 43: nit: it looks like this is repeated twice, remove?  Or did you mean to check y axis here?

libs/localization/src/lib/en.json
```diff
@@ -1341,6 +1343,143 @@
...
     },
+    "DataBalance": {
+      "AggregateBalanceMeasures": {
+        "Callout": {
+          "Description": "Aggregate balance measures allow us to obtain a higher notion of inequality about our dataset. They are calculated on the set of our columns of interest and do not depend on the label column. These measures look at the distribution of records across all feature value combinations of our columns of interest.",
+          "Title": "What are aggregate balance measures?"
+        },
+        "Measures": {
+          "AtkinsonIndex": {
+            "Description": "The Atkinson Index is a measure of income inequality that presents the percentage of total income that a given society would have to forego in order to have more equal shares of income between its citizens. For the dataset, this represents the percentage of records that need to be dropped to have a perfectly balanced dataset with respect to the columns of interest. The index lies between 0 and 1, where smaller values of the index indicate lower levels of inequality.",
```
imatiach-msft, line 1354: nice descriptions!

libs/localization/src/lib/en.json
```diff
@@ -1341,6 +1343,143 @@
...
+          "Title1": "Comparing",
+          "Title2": "with the Uniform Distribution"
+        },
+        "Measures": {
+          "ChiSquarePValue": {
+            "Description": "The p-value of the chi-square test gives evidence against the null hypothesis that the difference in observed and expected frequencies is by random chance.",
+            "Name": "Chi-Square P-Value"
+          },
+          "ChiSquareStatistic": {
+            "Description": "The chi-square test tests the null hypothesis that the categorical data has the given frequencies when compared with the expected frequencies in each category.",
```
imatiach-msft, line 1392: nit: maybe use validates instead of tests here, or some other synonym:
The chi-square test tests -> The chi-square test validates

imatiach-msft (approved):
added some minor comments

libs/dataset-explorer/src/lib/DataExplorerParentTab.tsx
ms-kashyap: Changed this to `DataAnalysisTab` which contains the Data Explorer and Data Balance tabs! 😊

This is based off Phil's suggestion that we call the overall component "Data analysis" with children "Data explorer" and "Data balance" tabs

libs/dataset-explorer/src/lib/AggregateBalanceMeasuresTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+    ];
+
+    return (
+      <Stack tokens={{ childrenGap: "l1" }} id="aggregateBalanceMeasures">
+        <Stack.Item>
+          <HeaderWithInfo
+            title={measuresLocalization.Name}
+            calloutTitle={measuresLocalization.Callout.Title}
+            calloutDescription={measuresLocalization.Callout.Description}
+            // TODO: Replace link with https://responsibleaitoolbox.ai/ link once docs are published there
```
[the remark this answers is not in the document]
ms-kashyap: Thanks for the reminder, I created a task item in AzureDevOps to track this :)

libs/e2e/src/lib/describer/modelAssessment/Constants.ts
ms-kashyap: Thanks for calling them out! 😊
First time writing UI tests and it was an interesting experience...

libs/e2e/src/lib/describer/modelAssessment/dataBalance/describeDistributionBalanceMeasures.ts
ms-kashyap: I meant y axis here! Oops

libs/localization/src/lib/en.json
ms-kashyap: Thank you! They can definitely be improved upon, so definitely give me suggestions in the future as we flight this

libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
ms-kashyap: I kept this method for now. I am also concerned about the word order in other languages. I actually switched the `boldText` `<Text>` objects to `<Label>`, but the idea is the same. Its a bunch of `<Text>`s and `<Label>`s one after another.

libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
ms-kashyap: It actually turns out that it's not necessary at all, only the `dropdown: width` is! 
Thank you for pointing this out, this is what I get for copying dropdown code from another part in the repo without asking what it does 😁

ms-kashyap:
Added some logic to show a message when some sets of measures are not computed. i.e. Feature balance measures are computed but not the other two, due to some computation error. 

When one set of measures are not computed:
image

When two sets of measures are not computed:
image

When all three set of measures are not computed:
image

ms-kashyap:
Hey @PIC123 and @martthalch, I wanted to inform you of these changes. Do they look good?

> Added some logic to show a message when some sets of measures are not computed. i.e. Feature balance measures are computed but not the other two, due to some computation error.
> 
> When one set of measures are not computed: image
> 
> When two sets of measures are not computed: image
> 
> When all three set of measures are not computed: image

libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+            {/* Renders the Feature dropdown and its heading */}
+            <Stack.Item>
+              <Stack tokens={{ childrenGap: "s1" }}>
+                <Stack.Item>
+                  <Label>{measuresLocalization.FeaturePicker}</Label>
+                </Stack.Item>
+                <Stack.Item>
+                  <Dropdown
+                    styles={{ dropdown: { width: 150 } }}
```
imatiach-msft, line 130: I think these styles belong in a separate .styles file

libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
```diff
@@ -0,0 +1,226 @@
...
+                  />
+                </Stack.Item>
+              </Stack>
+            </Stack.Item>
+          </Stack>
+        </Stack.Item>
+
+        {/* Renders the description of the chosen measure */}
+        <Stack.Item>
+          <Text style={{ fontWeight: 600 }}>{selectedMeasure} </Text>
```
imatiach-msft, line 179: did you try changing this to Label?  that should have same effect as increasing fontWeight

ms-kashyap: Ya, the main issue is that this text is part of a paragraph.
The paragraph format is `<measure name bolded> <measure desc>`. And when we use Label + Text, the two pieces of text don't flow together.

So current implementation looks like this (look at bottom text):
image

But when you do Label + Text, you get this:
image

So then I tried putting the Label + Text within a `<Stack horizontal>`, but it didn't look right either:
image

Any suggestions on how I can make Label + Text work? I'd prefer this but wasn't sure how

imatiach-msft: hmm, interesting, in my TextHighlighting component I had it set as:

https://github.com/microsoft/responsible-ai-toolbox/blob/main/libs/interpret-text/src/lib/TextExplanationDashboard/Control/TextHighlighting/TextHightlighting.tsx#L28

```<Stack horizontal horizontalAlign="start" tokens={textStackTokens}>```
where:
```
const textStackTokens: IStackTokens = {
  childrenGap: "s2",
  padding: "s2"
};
```
For label I have the style set to:
```
          fontSize: theme.fonts.large.fontSize,
          padding: 0,
```
Not sure why it doesn't work in your scenario, perhaps it's due to the wrapping. Does using horizontalAlign="start" change anything?

ms-kashyap: In trying this suggested, I realized that when a `<Text>` wraps to a newline (due to there being lots of text), the label and text don't render properly.

MicrosoftTeams-image (4)

PR #1522, opened 2022-06-28
Add License and Requirements to Source Distribtion

janjagusch:
## Description

Hi!

We're currently making an effort to put the responsible-ai-toolbox on conda-forge. You can find the corresponding PRs here:

- https://github.com/conda-forge/staged-recipes/pull/19439
- https://github.com/conda-forge/staged-recipes/pull/19449
- https://github.com/conda-forge/staged-recipes/pull/19445
- https://github.com/conda-forge/staged-recipes/pull/19446
- https://github.com/conda-forge/staged-recipes/pull/19448

We've noticed that there are some files missing in the source distribution that you ship on PyPI.

- The license is missing
- The `requirements.txt` file is missing, which is read inside the `setup.py` ([example](https://github.com/microsoft/responsible-ai-toolbox/blob/0a303e6fa1fc2a08d6c29f6b7f46b536579584e2/erroranalysis/setup.py#L14-L16))

Adding the `requirements.txt` to the source distribution can be accomplished by adding a `MANIFEST.in` file to the root of the package repository. As there is no way to reference files outside the root of the repository when building a source distribution ([source](https://github.com/pypa/setuptools/issues/2339)), we had to copy the license file into all package directories. We also made sure that the `dist/` directory does not get committed to version control when testing this.

## Checklist


romanlutz (approved):
Thanks @janjagusch ! This looks fine to me. @xuke444 @imatiach-msft ?

romanlutz:
FWIW we'll have to override the failing CD build since the secret isn't available for people with their own forks (which is also something we should fix at some point). Since there are no real code changes I'm not worried about that, though.

imatiach-msft (approved):

imatiach-msft:
@janjagusch do you need us to do another pypi release, or are you all set to add those packages to conda-forge?

imatiach-msft:
FYI I force-merged this PR due to the CD build failing, I think this is something that we really need to resolve as soon as possible

janjagusch:
> @janjagusch do you need us to do another pypi release, or are you all set to add those packages to conda-forge?

Thanks for merging @imatiach-msft! If you could release new versions, that would be great. I can work around the missing files, but this way to build would be much easier. Thanks a lot!

PR #1524, opened 2022-06-28
Panel buttons should be uniform: change all buttons to "Apply"; all buttons should be at bottom

tongyu-microsoft:
This PR updates all panel buttons include "Select", "Confirm", "Apply" to "Apply"; all buttons stays at the bottom.

## Description
1. feature list:
Before:
image

After:
image

2. 

## Checklist

### DIFF
--- a/libs/localization/src/lib/en.json
+++ b/libs/localization/src/lib/en.json
@@ -175,6 +175,7 @@
       "subText": "Learn about the selected cohort. Edit its cohort name. Delete this cohort."
     },
     "FeatureList": {
+      "Apply": "Apply",
       "features": "Features",
       "importances": "Importances",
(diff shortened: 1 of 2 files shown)


libs/localization/src/lib/en.json
```diff
@@ -180,6 +180,7 @@
       "subText": "Learn about the selected cohort. Edit its cohort name. Delete this cohort."
     },
     "FeatureList": {
+      "Apply": "Apply",
```
vinuthakaranth, line 183: nit: "apply"?

vinuthakaranth (approved):

libs/localization/src/lib/en.json
tongyu-microsoft: updated! thanks!

romanlutz (approved):

PR #1526, opened 2022-06-29
Fix slider bug

vinuthakaranth:
## Description

Fix slider bug caused by fluentui upgrade.
onChanged has a bug which resulted in not getting updated value. OnChange works as expected.

## Checklist


tongyu-microsoft (approved):

romanlutz:
Is it possible to write a test case that uses the slider to make sure we don't have future regressions?

vinuthakaranth:
> Is it possible to write a test case that uses the slider to make sure we don't have future regressions?
Good idea, I will add in coming PR. Will merge this as all gated have passed.

romanlutz (approved):
OK, great! Thank you.

PR #1525, opened 2022-06-29
Use of "==" and "!=" confusing to non-programmers, update operators to strings

tongyu-microsoft:
Feedback from a customer meeting: When the dashboard was using '!=' they didn't appreciate that this meant "Not Equal" - similarly "==" didn't mean anything to them. We should probably use more widely understood operators.

Acceptance criteria:
Age == 41: Age equal to 41
Age != 41: Age not equal to 41
Age < 41: Age less than 41
Age > 41: Age greater than 41

## Description
For causal treatment policy:
Before:
image

After:
image

For cohort filter list:
Before:
image

After:
image

## Checklist

- [x] I have added screenshots above for all UI changes.
- [ ] I have added e2e tests for all UI changes.
- [ ] Documentation was updated if it was needed.

PR #1528, opened 2022-06-30
Update tree operator from symbols to strings

tongyu-microsoft:
This PR updates tree operators from symbols to strings.

## Description
This is a follow up PR for https://github.com/microsoft/responsible-ai-toolbox/pull/1525 `Use of "==" and "!=" confusing to non-programmers, update operators to strings`.

Feedback from a customer meeting: When the dashboard was using '!=' they didn't appreciate that this meant "Not Equal" - similarly "==" didn't mean anything to them. We should probably use more widely understood operators.

Acceptance criteria:
Age == 41: Age equal to 41
Age != 41: Age not equal to 41
Age < 41: Age less than 41
Age > 41: Age greater than 41

## Checklist

### DIFF
--- a/apps/dashboard/src/error-analysis/__mock_data__/dummyTreeAdultCensusIncome.ts
+++ b/apps/dashboard/src/error-analysis/__mock_data__/dummyTreeAdultCensusIncome.ts
@@ -26,7 +26,7 @@ export const dummyTreeAdultCensusIncomeData: IErrorAnalysisTreeNode[] = [
   {
     arg: [1],
     badFeaturesRowCount: 0,
-    condition: "marital-status !=  Married-civ-spouse",
+    condition: "marital-status not equal to Married-civ-spouse",
     error: 1,
     id: 1,
     isErrorMetric: true,
@@ -45,7 +45,7 @@ export const dummyTreeAdultCensusIncomeData: IErrorAnalysisTreeNode[] = [
   {
     arg: [1],
     badFeaturesRowCount: 0,
-    condition: "marital-status ==  Married-civ-spouse",
+    condition: "marital-status equal to  Married-civ-spouse",
     error: 17,
     id: 2,
     isErrorMetric: true,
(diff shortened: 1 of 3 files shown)


apps/dashboard/src/error-analysis/__mock_data__/dummyTreeAdultCensusIncome.ts
```diff
@@ -26,7 +26,7 @@ export const dummyTreeAdultCensusIncomeData: IErrorAnalysisTreeNode[] = [
   {
     arg: [1],
     badFeaturesRowCount: 0,
-    condition: "marital-status !=  Married-civ-spouse",
+    condition: "marital-status not equal to Married-civ-spouse",
```
vinuthakaranth, line 29: should this all be localized?

tongyu-microsoft: Good point, looks like we have quite a few places where we don't localize the strings, not only this one. Let me update this one for now and I have reached out to PM and Ke for other localizations

gaugup:
Are there other places in the code where this needs to be changed? What about cohort conditions? example "age > 20"

erroranalysis/erroranalysis/_internal/cohort_filter.py
```diff
@@ -178,17 +178,20 @@ def build_query(filters, categorical_features, categories):
             arg0 = str(filter[ARG][0])
             colname = filter[COLUMN]
             if method == CohortFilterMethods.METHOD_GREATER:
-                queries.append("`" + colname + "` > " + arg0)
+                queries.append("`" + colname + "` greater than " + arg0)
```
imatiach-msft, line 181: hmm if we are using string in backend now instead of operators then it seems like we need to localize them now.  I haven't seen a case like this before, where the string is constructed in backend but displayed in frontend.  I wonder if it might be easier to send the two columns and operator to frontend and let it localize them.

PR #1541, opened 2022-07-07
Add column and feature range meta data to meta file

vinuthakaranth:
Add column and feature range meta data to meta file to help UI get these values in data explorer step of score card generation.

## Checklist


responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -524,14 +531,37 @@ def _save_metadata(self, path):
...
-            _CLASSES: classes
+            _CLASSES: classes,
+            _COLUMNS: list(self.test.columns),
+            _FEATURERANGES: feature_ranges
+
         }
         with open(top_dir / _META_JSON, 'w') as file:
             json.dump(meta, file)
+    
+    def _get_feature_ranges(self):
```
gaugup, line 547: Can you add unit tests for this change?

tongyu-microsoft (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -29,12 +29,19 @@
 _TARGET_COLUMN = 'target_column'
 _TASK_TYPE = 'task_type'
 _CLASSES = 'classes'
+_COLUMNS = 'columns'
+_FEATURERANGES = 'feature_ranges'
```
imatiach-msft, line 33: nit: maybe add an underscore "_" here, similar to variable value, so _FEATURE_RANGES

responsibleai/responsibleai/rai_insights/rai_insights.py
imatiach-msft: +1.  Also, can you please add doc strings to the method?  We don't have them on all private methods, although we really should, and I think it's better to have more than fewer docs in the code.

imatiach-msft (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
vinuthakaranth: Thanks, added test and doc string.

responsibleai/responsibleai/rai_insights/rai_insights.py
vinuthakaranth: thanks, updated

imatiach-msft (approved):

responsibleai/tests/test_rai_insights_save_and_load_scenarios.py
```diff
@@ -265,14 +271,18 @@ def validate_rai_insights(
...
+    columns
 ):
-
     pd.testing.assert_frame_equal(rai_insights.train, train_data)
     pd.testing.assert_frame_equal(rai_insights.test, test_data)
     assert rai_insights.target_column == target_column
     assert rai_insights.task_type == task_type
     assert rai_insights.categorical_features == (categorical_features or [])
+    assert feature_range_keys.sort() == \
+        list(rai_insights._feature_ranges[0].keys()).sort()
```
imatiach-msft, line 284: you probably want to validate the other metadata as well, like range type, min/max etc (both that keys exist and some value) to make sure the format is as expected

vinuthakaranth: I didn't do exact value comparison as it was producing different value for erroranalysis/countefactual at times.

gaugup (approved):

PR #1552, opened 2022-07-12
SDK change for computing aggregate causal effects and policy tree per cohort

gaugup:
## Description

SDK change for computing aggregate causal effects and policy tree per cohort. It seems we report the cohort stats on the "All data" cohort. Changing the cohort doesn't change the aggregate stats. This change adds the SDK flask end points which when queried returns the global stats per cohort.

## Checklist


raiwidgets/raiwidgets/responsibleai_dashboard.py
```diff
@@ -66,6 +66,13 @@ def causal_whatif():
 
         self.add_url_rule(causal_whatif, '/causal_whatif', methods=["POST"])
 
+        def causal_global_effects():
```
imatiach-msft, line 69: should we add some dropdown in the UI to choose the cohort?  I know the other components either have a dropdown (like dataset explorer) or allow the user to see the cohorts side by side (model performance, feature importances):
image

responsibleai/tests/causal/test_causal_manager.py
```diff
@@ -27,6 +27,30 @@ def test_causal_no_categoricals(self, housing_data):
...
+        EFFECTS_ATTRIBUTES = [
+            'point',
+            'stderr',
+            'zstat',
+            'ci_lower',
+            'ci_upper',
+            'p_value',
+        ]
+        for effect in global_effect_X_test:
+            assert all(attr in effect for attr in EFFECTS_ATTRIBUTES)
```
imatiach-msft, line 52: I wonder, is there some way to assert that the statistics are indeed for that specific cohort and not for the overall all data group?  Maybe something as simple as looking at the effects and validating that the values are different from the all data cohort, or for a better test, perhaps even validating the values shifted based on the instances we reduced to in the dataset (maybe then you have to choose the rows with a filter instead of random sample in order to make it easier to get consistent results and test for this).  Not necessary for this PR, but I think it would be good to have some tests for cohorts like that.  I suppose our other components should have similar tests if we don't have them already, to validate that our computed statistics do indeed change as we expect them to for specific cohorts.

imatiach-msft (approved):

raiwidgets/tests/test_responsibleai_dashboard_input.py
```diff
@@ -220,3 +222,44 @@ def test_rai_dashboard_input_housing_causal_whatif_failure(
...
+    def test_rai_dashboard_input_housing_causal_causal_global_effects_failure(
+            self, create_rai_insights_object_regression):
+        ri = create_rai_insights_object_regression
+        id = "some_id_that_does_not_exist"
+        causal_whatif_test_data = ri.test.head(1).drop(
+            "target", axis=1).to_dict(orient='records')
+
+        dashboard_input = ResponsibleAIDashboardInput(ri)
+        post_data = (id, causal_whatif_test_data)
+        flask_server_prediction_output = dashboard_input.causal_global_effects(
```
imatiach-msft, line 253: nit: I think most of the code in this test from lines 245-254 and in the test above from lines 228-236 could be refactored into a common helper method (and id passed as a parameter) to reduce code duplication

gaugup: Refactored the tests into per feature file.

raiwidgets/raiwidgets/responsibleai_dashboard.py
gaugup: The PR just adds the SDK end points for global effects and global policy. The way I am thinking about these would work in UI would be through global selected cohort. What do you think?

responsibleai/tests/causal/test_causal_manager.py
gaugup: Now that we have cohort filtering is responsibleai, we should have some tests around cohort filtering. But as you said, this is outside of scope of this PR.

PR #1553, opened 2022-07-13
Add ID to counterfactual config and serialized counterfactual data

gaugup:
## Description

This will enable querying for counterfactuals from UI. UI can pass the id using which we will be able to look up the counterfactual config / explainer using which we can generate the counterfactuals.

## Checklist

### DIFF
--- a/responsibleai/tests/counterfactual/test_counterfactual_advanced_features.py
+++ b/responsibleai/tests/counterfactual/test_counterfactual_advanced_features.py
@@ -119,10 +119,12 @@ class TestCounterfactualAdvancedFeatures(object):
         cf_obj_2 = rai_insights_copy.counterfactual.get()[1]
         assert cf_obj_2 is not None
 
-        assert counterfactual_config_list_before_save[0].id == \
-            counterfactual_config_list_after_save[0].id
-        assert counterfactual_config_list_before_save[1].id == \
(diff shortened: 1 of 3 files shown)


responsibleai/tests/counterfactual/test_counterfactual_advanced_features.py
```diff
@@ -136,3 +150,37 @@ def test_counterfactual_manager_save_load(self, tmpdir):
...
+            task_type='classification')
+
+        rai_insights.counterfactual.add(
+            total_CFs=10, desired_class=2,
+            features_to_vary=[feature_names[0]],
+            permitted_range={feature_names[0]: [2.0, 5.0]})
+        rai_insights.counterfactual.compute()
+
+        serialized_data = rai_insights.counterfactual.get_data()
+        assert serialized_data is not None
```
imatiach-msft, line 176: nit: maybe extract the asserts into a utility function, as it seems like it could be re-used in other tests as well?

imatiach-msft (approved):

responsibleai/tests/counterfactual/test_counterfactual_advanced_features.py
gaugup: Right now these asserts are only in this function. Maybe, we can move them into a function when other tests use these asserts.

PR #1554, opened 2022-07-13
Fix bug in deserializing counterfactual explainer

gaugup:
## Description

The deserializing of the counterfactual had a bug which was causing the retraining of counterfactual explainer during load().

## Checklist


tongyu-microsoft (approved):

imatiach-msft:
is there some way to add a test to validate the explainer is not getting re-trained on load?

PR #1542, opened 2022-07-08
[UI refactor] Remove .bind() part2 - Remove .bind in counterfactual folder and part of .bind() in other folders

tongyu-microsoft:
This PR removes .bind() in libs\counterfactuals folder. And also removes part of .bind() in core-ui, counterfactual, dataset-explorer, error-analysis folders.

## Description
The UI refactor work is to remove all .bind() functions. And this PR is part 2. 
part 1 PR link: https://github.com/microsoft/responsible-ai-toolbox/pull/1523.
There will be following PRs to remove other .bind() functions. 
The reason that we want to remove .bind() is that, a different callback is created each time the component renders. So we want to remove .bind() and set lint rule in the end to ban adding .bind() in the future.

## Checklist

- [ ] I have added screenshots above for all UI changes.
- [ ] I have added e2e tests for all UI changes.
- [ ] Documentation was updated if it was needed.

PR #1555, opened 2022-07-13
[UI refactor] Remove .bind() part3 - Remove .bind for SpinButton in AxisConfigDialog, MultiICEPlot and Fairness

tongyu-microsoft:
This PR removes .bind() for SpinButton in AxisConfigDialog, MultiICEPlot and Fairness folder. Also removes .bind() for cohortList by creating a cohortNameColumn component.

## Description
The UI refactor work is to remove all .bind() functions. And this PR is part 3.
part 1 PR link: https://github.com/microsoft/responsible-ai-toolbox/pull/1523.
part 2 PR link: https://github.com/microsoft/responsible-ai-toolbox/pull/1542.
There will be following PRs to remove other .bind() functions.
The reason that we want to remove .bind() is that, a different callback is created each time the component renders. So we want to remove .bind() and set lint rule in the end to ban adding .bind() in the future.

## Checklist


libs/core-ui/src/lib/components/AxisConfigDialogSpinButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+export interface IAxisConfigSpinButtonProps {
+  label: string;
+  min?: number;
+  max?: number;
+  styles?: IStyleFunctionOrObject<ISpinButtonStyleProps, ISpinButtonStyles>;
+  value?: string;
+  setNumericValue: (delta: number, stringVal: string) => string | void;
+}
+
+export class AxisConfigDialogSpinButton extends React.PureComponent<IAxisConfigSpinButtonProps> {
```
imatiach-msft, line 23: nice refactor into a separate class!

imatiach-msft (approved):

libs/core-ui/src/lib/components/AxisConfigDialogSpinButton.tsx
tongyu-microsoft: Thanks!!

libs/fairness/src/lib/Controls/BinDialog.styles.ts
```diff
@@ -95,3 +95,17 @@ export const BinDialogStyles: () => IProcessedStyleSet<IBinDialogStyles> =
       }
     });
   };
+
+export const SpinButtonStyles = {
```
RubyZ10, line 99: nit: Is this the same as the one added in MultiCEPlpt.styles.ts? wondering if it makes sense to take it out and put in a common place

RubyZ10 (approved):

RubyZ10 (approved):

libs/fairness/src/lib/Controls/BinDialog.styles.ts
tongyu-microsoft: Good point! Let me create a common SpinButtonStyles so it can be shared!

PR #1556, opened 2022-07-14
[Model Overview] Fix bug: Unable to hover/click on download pdf/svg image for dataset cohort metric chart

RubyZ10:

This PR fix bug for the model overview: Unable to hover/click on download pdf/svg image for dataset cohort metric chart
https://msdata.visualstudio.com/Vienna/_workitems/edit/1874045

## Description

When there is only one cohort, users are not able to click on the last two download options.
 
Root case: when there is only one cohort, the height is set to 160, but the minimum height for showing the whole menu is around 190, so the menu is cut off and users are not able to click on the last two options

Fix: set the minimum height as 190

Before change
MOdownloadMenuBeforeFix

After change

MOdownloadMenu

## Checklist


tongyu-microsoft (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/CohortStatsHeatmap.tsx
```diff
@@ -97,6 +97,9 @@ export class CohortStatsHeatmap extends React.Component<
             min: 0,
             minColor
           },
+          custom: {
```
imatiach-msft, line 100: two minor things:
1.) I think most of these properties belong in some styles file, and can be re-imported here
2.) "magic numbers" - is the 190 coming from somewhere else?  can we re-use the same variable to reference that value?

imatiach-msft (approved):

PR #1562, opened 2022-07-15
Improve error message when categorical string column is treated as numeric and error analysis is run

imatiach-msft:
## Description

When a column is not specified as categorical, although it has string values, and then error analysis is run on the column, a cryptic crash can appear:

```
TypeError: '<=' not supported between instances of 'str' and 'float'
```

This PR improves the error message to be more useful to the user when using error analysis:

```
Column MY_COL_NAME of type string is incorrectly treated as numeric with threshold value 7.77. Please make sure it is marked as categorical instead.
```

## Checklist

### DIFF
--- a/erroranalysis/erroranalysis/_internal/surrogate_error_tree.py
+++ b/erroranalysis/erroranalysis/_internal/surrogate_error_tree.py
@@ -691,7 +691,8 @@ def node_to_dict(df, tree, nodeid, categories, json,
                 arg = float(parent_threshold)
                 condition = "{} <= {:.2f}".format(p_node_name,
                                                   parent_threshold)
-                df = df[df[p_node_name_val] <= parent_threshold]
+                filter_by_threshold(df, p_node_name_val,
+                                    parent_threshold, side)
             elif parent_decision_type == '==':
                 method = CohortFilterMethods.METHOD_INCLUDES
                 arg = create_categorical_arg(parent_threshold)
@@ -708,7 +709,8 @@ def node_to_dict(df, tree, nodeid, categories, json,
                 arg = float(parent_threshold)
                 condition = "{} > {:.2f}".format(p_node_name,
                                                  parent_threshold)
-                df = df[df[p_node_name_val] > parent_threshold]
+                filter_by_threshold(df, p_node_name_val,
+                                    parent_threshold, side)
             elif parent_decision_type == '==':
                 method = CohortFilterMethods.METHOD_EXCLUDES
                 arg = create_categorical_arg(parent_threshold)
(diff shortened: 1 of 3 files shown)


erroranalysis/erroranalysis/_internal/surrogate_error_tree.py
```diff
@@ -739,6 +741,34 @@ def node_to_dict(df, tree, nodeid, categories, json,
...
+    :param side: The side of the parent node.
+    :type side: TreeSide
+    """
+    try:
+        if side == TreeSide.LEFT_CHILD:
+            df = df[df[p_node_name_val] <= parent_threshold]
+        else:
+            df = df[df[p_node_name_val] > parent_threshold]
+    except TypeError as e:
+        mismatch_err = "not supported between instances of 'str' and 'float'"
```
gaugup, line 762: Is there a better way to check for this error? It might be really flaky to check for the above error message. If the underlying message changes then this logic will break.

gaugup: May be check for the dtype of the that column. it shouldn't be integer or float in this case.

imatiach-msft: Done.  The dtype is object in this case (and integer values can have object dtype as well), so I checked the value manually instead.

erroranalysis/erroranalysis/_internal/surrogate_error_tree.py
```diff
@@ -739,6 +741,36 @@ def node_to_dict(df, tree, nodeid, categories, json,
...
+def filter_by_threshold(df, p_node_name_val, parent_threshold, side):
+    """Filters the DataFrame by the threshold of the parent node.
+
+    :param df: The DataFrame to filter.
+    :type df: pandas.DataFrame
+    :param p_node_name_val: The name of the parent node.
+    :type p_node_name_val: str
+    :param parent_threshold: The threshold of the parent node.
+    :type parent_threshold: float
+    :param side: The side of the parent node.
```
gaugup, line 753: Return type of pandas dataframe missing.

erroranalysis/erroranalysis/_internal/surrogate_error_tree.py
```diff
@@ -739,6 +741,36 @@ def node_to_dict(df, tree, nodeid, categories, json,
...
+        has_vals = df[p_node_name_val].shape[0] > 0
+        if has_vals and isinstance(df[p_node_name_val][0], str):
+            err = ("Column {0} of type string is incorrectly treated "
+                   "as numeric with threshold value {1}. "
+                   "Please make sure it is marked as categorical instead.")
+            err = err.format(p_node_name_val, parent_threshold)
+            raise TypeError(err, e)
+        else:
+            raise e
+    return df
```
gaugup, line 771: Maybe return df after line 760. Looking a little misplaced here. The except case is raising an exception in both branches anyways.

erroranalysis/erroranalysis/_internal/surrogate_error_tree.py
imatiach-msft: good catch, done!

erroranalysis/erroranalysis/_internal/surrogate_error_tree.py
imatiach-msft: moved the return up before the except, done!

gaugup (approved):

PR #1568, opened 2022-07-20
Add hierarchy to rai exceptions

gaugup:
## Description

There are broadly two kinds of the errors that could happen in the `responsibleai` system. One set of error can be caused by bad inputs from the users which we could classify as `User Errors`. The second set of errors could happen due to error conditions in the `responsibleai` system which could mean that we halt the execution. These could be classified as `System Errors`. Ideal goal of our product should to minimize systems errors and be able to progressively capture cases where users can run into errors with the system,

This PR defines two base exceptions for `UserErrorException` and `SystemErrorException` and inherits the child exception classes, like `DuplicateManagerConfigException`, from the above two classes. This could in future help us identify the user errors and systems errors clearly and help us narrow down the system errors.

## Checklist


responsibleai/tests/test_exceptions.py
```diff
@@ -0,0 +1,46 @@
...
+
+class TestExceptions:
+    def test_user_exceptions(self):
+        ucve = UserConfigValidationException(TEST_MESSAGE)
+        assert ucve._error_code == 'Invalid config'
+        assert str(ucve) == TEST_MESSAGE
+        assert isinstance(ucve, UserErrorException)
+
+        de = DuplicateManagerConfigException(TEST_MESSAGE)
+        assert de._error_code == 'Duplicate RAI configuration detected.'
```
imatiach-msft, line 23: nit: I think you could refactor this slightly by calling a method that does these asserts, and passing to the method the exception object, expected error code, the TEST_MESSAGE and the type of exception

imatiach-msft (approved):

vinuthakaranth (approved):

responsibleai/tests/test_exceptions.py
gaugup: Done

PR #1572, opened 2022-07-21
create other cohort in model overview when there are more than 10 cohorts

imatiach-msft:
## Description

Create other cohort in model overview when there are more than 10 cohorts.

image

## Checklist

### DIFF
--- a/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/DisaggregatedAnalysisUtils.ts
+++ b/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/DisaggregatedAnalysisUtils.ts
@@ -11,6 +11,7 @@ import {
   JointDataset,
   Operations
 } from "@responsible-ai/core-ui";
+import { localization } from "@responsible-ai/localization";
 
 import { defaultNumberOfContinuousFeatureBins } from "./Constants";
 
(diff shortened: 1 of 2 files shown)


libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/DisaggregatedAnalysisUtils.ts
```diff
@@ -82,11 +84,45 @@ export function generateOverlappingFeatureBasedCohorts(
     }
   });
 
-  return (
-    generateCohortsCartesianProduct(filters, jointDataset, globalCohort)
-      // filter the empty cohorts resulting from overlapping dimensions
-      .filter((errorCohort) => errorCohort.cohortStats.totalCohort > 0)
-  );
+  const nonZeroCohorts = generateCohortsCartesianProduct(
```
gaugup, line 87: Is the above filtering needed as we don't allow users to create cohorts which are empty?

imatiach-msft: Yes, we are automatically generating cohorts from the features with all possible combinations.  Then we filter out the empty ones.  It was like this before. I added the logic now to group the cohorts with smaller sample sizes into one "Other" cohort.

gaugup (approved):

PR #1573, opened 2022-07-21
fix release pipeline failing on UI widget tests

imatiach-msft:
## Description

Fix release pipeline failing on UI widget tests.  The fix here is to run the tests without flights prior to release.
Alternatively, we could run all UI tests as an extra step, but this seems unnecessary for release pipeline.

## Checklist

### DIFF
--- a/.github/workflows/release-rai.yml
+++ b/.github/workflows/release-rai.yml
@@ -60,17 +60,10 @@ jobs:
         run: |
           yarn buildall
 
-      - if: ${{ matrix.operatingSystem != 'macos-latest' }}
-        name: Install pytorch on non-MacOS
+      - name: Install pytorch
         shell: bash -l {0}
         run: |
           conda install --yes --quiet pytorch torchvision captum cpuonly -c pytorch
-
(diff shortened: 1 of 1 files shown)


.github/workflows/release-rai.yml
```diff
@@ -60,18 +60,11 @@ jobs:
...
+      - name: Install pytorch
         shell: bash -l {0}
         run: |
           conda install --yes --quiet pytorch torchvision captum cpuonly -c pytorch
 
-      - if: ${{ matrix.operatingSystem == 'macos-latest' }}
-        name: Install Anaconda packages on MacOS, which should not include cpuonly according to official docs
-        shell: bash -l {0}
-        run: |
-          conda install --yes --quiet pytorch torchvision captum -c pytorch
```
vinuthakaranth, line 73: we don't need this anymore?

vinuthakaranth (approved):

.github/workflows/release-rai.yml
imatiach-msft: yes we still install pytorch but for ubuntu only since release script only runs in one OS, this statement was just not run

PR #1575, opened 2022-07-21
Add more classification metrics to erroranalysis

gaugup:
## Description

Add computation of `false_positive_rate`, `false_negative_rate` and `selection_rate` to metrics module in erroranalysis.

## Checklist


erroranalysis/erroranalysis/_internal/metrics.py
```diff
@@ -13,6 +14,48 @@
...
+    :param y_pred: Predicted labels.
+    :type y_pred: numpy.ndarray
+    :return: False positive rate.
+    :rtype: float
+    """
+    _, fp, _, tn = confusion_matrix(y_true, y_pred).ravel()
+    return fp / (fp + tn)
+
+
+def selection_rate(y_true, y_pred):
```
imatiach-msft, line 45: Interesting, where did you see this formula defined?  Is this the same as prevalence here in wikipedia:

https://en.wikipedia.org/wiki/Precision_and_recall
https://en.wikipedia.org/wiki/Prevalence

P / P+N, where I assume P refers to the ground truth or the actual condition being all positive, so P = fn+tp, since FN means the ground truth was positive, unless I misunderstood something.
Just wondering about the terminology used here.

imatiach-msft (approved):

erroranalysis/erroranalysis/_internal/metrics.py
gaugup: Not sure about the theory from your links, I copied the formula from https://github.com/microsoft/responsible-ai-toolbox/blob/0e128fab4a3baf85f51658284473e79aa32e011a/libs/core-ui/src/lib/util/StatisticsUtils.ts#L90. Is it similar to what you are describing?

imatiach-msft: I see, it looks like it was added in PR https://github.com/microsoft/responsible-ai-toolbox/pull/1174 by @romanlutz , I'm not sure what the source for the metric was.  Maybe we can leave as is for now then and try to revisit this later once he is back.

gaugup: Thanks @imatiach-msft.

PR #1576, opened 2022-07-21
Add multiclass classification test in raiwidgets

gaugup:
## Description

It seems we don't test multiclass scenario in `raiwidgets`. It might be good to have as we add more end points for widgets flask server.

## Checklist

### DIFF
--- a/raiwidgets/tests/conftest.py
+++ b/raiwidgets/tests/conftest.py
@@ -5,7 +5,7 @@ import pandas as pd
 import pytest
 import shap
 import sklearn
-from sklearn.datasets import fetch_california_housing
+from sklearn.datasets import fetch_california_housing, load_iris
 from sklearn.ensemble import RandomForestRegressor
 from sklearn.model_selection import train_test_split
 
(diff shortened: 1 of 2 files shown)


raiwidgets/tests/conftest.py
```diff
@@ -73,3 +73,28 @@ def create_rai_insights_object_regression():
                   skip_cat_limit_checks=True)
     ri.compute()
     return ri
+
+
+@pytest.fixture(scope='session')
+def create_rai_insights_object_multiclass_classification():
```
imatiach-msft, line 79: nit: maybe shorten the name to:

create_rai_insights_multiclass_classification

I don't think object is needed in the name

raiwidgets/tests/test_responsibleai_dashboard_input.py
```diff
@@ -220,3 +220,23 @@ def test_rai_dashboard_input_housing_causal_whatif_failure(
         assert "Failed to generate causal what-if," in \
             flask_server_prediction_output[
                 WidgetRequestResponseConstants.error]
+
+
+class TestResponsibleAIDashboardInputMultiClassClassification:
+    def test_rai_dashboard_input_iris_on_predict_success(
+            self, create_rai_insights_object_multiclass_classification):
+        ri = create_rai_insights_object_multiclass_classification
```
imatiach-msft, line 228: tiny nit: maybe if we shorten the name the re-assign here won't be needed?
```suggestion
        rf = create_rai_insights_multiclass_classification.model
        test_data = create_rai_insights_multiclass_classification.test
```

raiwidgets/tests/test_responsibleai_dashboard_input.py
```diff
@@ -220,3 +220,23 @@ def test_rai_dashboard_input_housing_causal_whatif_failure(
...
+        rf = ri.model
+        test_data = ri.test
```
imatiach-msft, line 230: tiny nit: maybe if we shorten the name the re-assign here won't be needed?
```suggestion
        rf = create_rai_insights_multiclass_classification.model
        test_data = create_rai_insights_multiclass_classification.test
```

imatiach-msft (approved):

raiwidgets/tests/conftest.py
gaugup: The fixture indeed returns an RAIInsights object. Not sure if there is need to shorten the fixture name unless it violates the linter rules.

gaugup: the other fixtures also carry the object. So keeping it the same,

raiwidgets/tests/test_responsibleai_dashboard_input.py
gaugup: closing this as responded to earlier comment,

PR #1578, opened 2022-07-23
Remove strict pinnings from rai-core-flask requirements.txt.

janjagusch:
Hi!

We're working on making the `responsible-ai-toolbox` available on conda-forge. [We've noticed that the `rai-core-flask` package has very strict pinnings in its `install_requires`](https://github.com/conda-forge/staged-recipes/pull/19445#discussion_r924282764), making it difficult to install it with other packages in the same environment. [This is because the contents of the `requirements.txt` file are used to set the `install_requires` section in the `setup.py`.](https://github.com/microsoft/responsible-ai-toolbox/blob/e3d0f24166247e601a129cd437889600a500893b/rai_core_flask/setup.py#L41)

In this PR, we propose removing the strict pinnings (replacing `==` with `<=` or `>=`). We adjusted the pinnings to the best of our knowledge. Still, some changes might be incorrect, so please review this carefully.

If you agree with the proposed changes, it would be much appreciated if you could release of new version of `rai-core-flask`, so we can also loosen the strict pinnings in the conda-forge feedstock.

Thanks a lot for your help! 🙏

### DIFF
--- a/rai_core_flask/requirements.txt
+++ b/rai_core_flask/requirements.txt
@@ -1,9 +1,9 @@
 Flask
 Flask-Cors
-ipython==7.16.3; python_version <= '3.6'
-ipython==7.31.1; python_version > '3.6'
-itsdangerous==2.0.1
-greenlet==1.1.2
-gevent==21.12.0
+ipython<=7.16.3; python_version <= '3.6'
+ipython>=7.31.1; python_version > '3.6'
+itsdangerous>=2.0.1
+greenlet>=1.1.2
+gevent>=21.12.0
 markupsafe<2.1.0
 Werkzeug<2.1.0


janjagusch:
I'm pretty sure the CI failure is unrelated to my changes and mostly likely due to a missing secret in my fork:

[`There are no credentials provided in your command and environment, we will query for the account key inside your storage account. `](https://github.com/microsoft/responsible-ai-toolbox/runs/7480751111?check_suite_focus=true#step:23:31)

imatiach-msft (approved):

gaugup (approved):

imatiach-msft:
merging this PR on behalf of submitter since I don't see any objections to this PR

imatiach-msft:
"If you agree with the proposed changes, it would be much appreciated if you could release of new version of rai-core-flask, so we can also loosen the strict pinnings in the conda-forge feedstock."
Sounds like I need to create a PR to do a version bump of rai-core-flask and then do a release to pypi.

PR #1583, opened 2022-07-25
[Bugfix] Fix Dropdown bug and Update styling for feature balance measures

ms-kashyap:
## Description

Fixes bug where the 0th element in a dropdown can't be selected because it fails the `if (item?.key)` check.

Addresses the open comments in PR #1512, namely:

- Moving the `<Dropdown>` styles into the styles file
- Adding a comment for why `<Label>` can't be used in feature balance measures description

When a `<Text>` wraps to a newline (due to there being lots of text), the label and text don't render properly.

The `Bolded text` is a `<Label>` and to the right of it is a `<Text>`:
MicrosoftTeams-image (4)

## Checklist


libs/dataset-explorer/src/lib/FeatureBalanceMeasuresChart.tsx
```diff
@@ -176,6 +179,8 @@ export class FeatureBalanceMeasuresChart extends React.PureComponent<
 
         {/* Renders the description of the chosen measure */}
         <Stack.Item>
+          {/* The description is a paragraph with the format "<measure name bolded> <measure desc>".
+              Because the measure desc wraps to a new line, Label cannot be used for the measure name. */}
           <Text style={{ fontWeight: 600 }}>{selectedMeasure} </Text>
```
imatiach-msft, line 184: In this case, I think you can move the fontWeight to a styles file as well.  It still seems strange to me that we can't change label to not wrap for this scenario.

imatiach-msft (approved):

imatiach-msft (approved):

PR #1593, opened 2022-07-28
fix bug in treeview filtering logic in error analysis package

imatiach-msft:
## Description

Fix bug introduced in recent PR:
https://github.com/microsoft/responsible-ai-toolbox/pull/1562
During refactoring of a method the filtering logic broke.  This PR fixes the filtering logic again and adds a test to validate the nodes from error analysis constructed json are the same size as the original lightgbm tree exported model file.

## Checklist


gaugup (approved):

gaugup:
looks like linting failed with error. Please check.

PR #1597, opened 2022-08-01
Make generic class interface for cohort based filtering

gaugup:
## Description

Introducing the class `FilterDataWithCohortFilters` to provide a generic data filtering using cohorts. This has all the current functionality in cohort_filter.py which are not available under the class `FilterDataWithCohortFilters`. User needs to setup the class using dataset, model, feature types etc and they can then filter the supplied dataset using the function `filter_data_from_cohort`. 

## Checklist


erroranalysis/erroranalysis/_internal/cohort_filter.py
```diff
@@ -26,259 +29,346 @@ def filter_from_cohort(analyzer, filters, composite_filters):
...
-        if true_y[i] == pred_y[i]:
-            if true_y[i] == classes[0]:
-                # True positive == 0
-                classification_outcome.append(0)
+        :param filters: The filters.
+        :type filters: list[dict]
+        :param composite_filters: The composite filters.
+        :type composite_filters: list[dict]
+        :return: The filtered dataset converted to a pandas DataFrame.
+        :param original_data_columns: Whether to just include the original
```
imatiach-msft, line 103: nit: maybe include_original_columns might be a better variable name for a boolean type?  original_data_columns makes me think that I need the list of columns themselves.

erroranalysis/erroranalysis/_internal/cohort_filter.py
```diff
@@ -26,259 +29,346 @@ def filter_from_cohort(analyzer, filters, composite_filters):
...
-    # compute the necessary column data on the fly
-    if model_task == ModelTask.CLASSIFICATION and filters:
-        for filter in filters:
-            if COLUMN in filter:
-                column = filter[COLUMN]
-                if column == CLASSIFICATION_OUTCOME:
-                    return True
-    return False
+class FilterDataWithCohortFilters:
+    def __init__(self, model: Any, dataset: pd.DataFrame,
```
imatiach-msft, line 54: nice refactor!

imatiach-msft (approved):

erroranalysis/erroranalysis/_internal/cohort_filter.py
imatiach-msft: same above on line 22

gaugup: Thanks for the suggestion. Changed it to include_original_columns_only.

erroranalysis/erroranalysis/_internal/cohort_filter.py
gaugup: Thanks

PR #1599, opened 2022-08-02
Fix tooltip location showing up incorrectly due to conflicting ids

imatiach-msft:
## Description

When clicking info icon in the local importances plot, the tooltip shows up in the wrong location due to duplicate ids "cross-class-weight-info" in the code.  That id in the LabelWithCallout component seems to actually be inconsistent with what the component does, so I assume it was a copy-paste error.  Giving the component a different id resolved the issue.

Before fix:
image

After fix:
image

## Checklist


gaugup:
Can you add the output after your fix?

gaugup (approved):

PR #1600, opened 2022-08-02
Refactor categorical statistics computation in separate module

gaugup:
## Description

 Refactored the categorical stats computation logic into its own module in error analysis so that RAIInsights can leverage the same.

## Checklist


erroranalysis/erroranalysis/_internal/process_categoricals.py
```diff
@@ -0,0 +1,54 @@
...
+                               remainder='drop')
+        string_ind_data = ct.fit_transform(dataset)
+        transformer_categories = ct.transformers_[0][1].categories_
+        for category_arr, category_index in zip(transformer_categories,
+                                                categorical_indexes):
+            category_values = category_arr.tolist()
+            categories.append(category_values)
+            category_dictionary[category_index] = category_values
+
+    return categories, categorical_indexes, category_dictionary,\
```
imatiach-msft, line 53: nit: you can also put them in parens () to avoid the \ , but this is ok too

imatiach-msft (approved):

erroranalysis/erroranalysis/_internal/process_categoricals.py
gaugup: Will leave this line as is.

PR #1604, opened 2022-08-03
responsibleai: Add FeatureMetadata class and related tests

gaugup:
## Description

This PR adds the `FeatureMetadata` class and the related unit tests. This class will allow users specify what are the various types of features that are present in their data. This will allow the `responsibleai` SDK to better handle the different types of features. 
In the PR, the following types of feature list have been added. More feature type lists can be added to this class later.

```
    def __init__(self, identity_feature_name: Optional[str] = None,
                 sensitive_features: Optional[List[str]] = None,
                 datetime_features: Optional[List[str]] = None,
                 categorical_features: Optional[List[str]] = None,
                 dropped_features: Optional[List[str]] = None):
        """Placeholder class for feature metadata provided by the user.

        :param identity_feature_name: Name of the feature which helps to
                                      uniquely identify a row or instance
                                      in user input dataset.
        :type identity_feature_name: Optional[str]
        :param datetime_features: List of datetime features in the
                                  user input dataset.
        :type datetime_features: Optional[List[str]]
        :param categorical_features: List of categorical features in the
                                     user input dataset.
        :type categorical_features: Optional[List[str]]
        :param dropped_features: List of features that were dropped by the
                                 the user during training of their model.
        :type dropped_features: Optional[List[str]]
```
 
## Checklist


responsibleai/responsibleai/feature_metadata.py
```diff
@@ -0,0 +1,97 @@
...
+        :param dropped_features: List of features that were dropped by the
+                                 the user during training of their model.
+        :type dropped_features: Optional[List[str]]
+        """
+        self.identity_feature_name = identity_feature_name
+        self.sensitive_features = sensitive_features
+        self.datetime_features = datetime_features
+        self.categorical_features = categorical_features
+        self.dropped_features = dropped_features
+        if self.sensitive_features is not None:
```
imatiach-msft, line 40: if sensitive_features are not none, why do we print that they are not in use?  Wouldn't they not be in use if they were none instead?  And then, why do we need to print a warning if that is the case?

responsibleai/tests/test_feature_metadata.py
```diff
@@ -0,0 +1,161 @@
...
+        expected_feature_metadata_dict = {
+            'identity_feature_name': None,
+            'sensitive_features': None,
+            'datetime_features': None,
+            'categorical_features': None,
+            'dropped_features': None
+        }
+        assert feature_metadata_dict == expected_feature_metadata_dict
+
+    def test_feature_metadata_with_identity_feature(self):
```
imatiach-msft, line 29: nice tests!

imatiach-msft (approved):

responsibleai/responsibleai/feature_metadata.py
gaugup: These are not in use currently. Just the identity feature will be in use in the dashboard. These can be used in future. So putting a warning, if a user tries to set them.

responsibleai/tests/test_feature_metadata.py
gaugup: Thanks

responsibleai/responsibleai/feature_metadata.py
gaugup: Removed sensitive_features for now. This can be added back later.

PR #1614, opened 2022-08-04
Add feature metadata to IDataset and JointDataset

gaugup:
## Description

This change add the `FeatureMetaData` to `IDataset` and `JointDataset`. This is the foundation work for adding feature metadata for UI and will be followed by integration of identity feature into feature importance, counterfactual and causal scatter plots. Also, included is the unit test for the `JointDataset` class for this change. 

## Checklist


libs/core-ui/src/lib/util/JointDatasetUtils.ts
```diff
@@ -54,3 +56,7 @@ export interface IJointMeta {
   category: ColumnCategories;
   index?: number;
 }
+
+export interface IDatasetMeta {
+  featureMetaData?: IFeatureMetaData;
```
vinuthakaranth, line 61: to confirm, will we have more metadata in future other than featureMetaData??

vinuthakaranth (approved):

vinuthakaranth:
Thanks for adding unit tests!!

libs/core-ui/src/lib/util/JointDatasetUtils.ts
gaugup: yes. likely we will have more such metadata that can come in. For example the feature ranges etc, could also be added here.

PR #1623, opened 2022-08-09
Handle 'Classification outcome' cohort filter for multiclass scenario in SDK

gaugup:
## Description

It looks like we have "Classification Error" (aka "Classification outcome") filter supported for multiclass scenario. This was added as a part of the PR https://github.com/microsoft/responsible-ai-toolbox/pull/1371. Hence, supporting the "Classification outcome" filter for multiclass case. 

## Checklist


erroranalysis/tests/test_cohort_filter.py
```diff
@@ -137,6 +137,40 @@ def test_cohort_filter_in_the_range_of(self):
...
+    def test_cohort_filter_multiclass_classification_outcome(
+            self, arg, correct_prediction):
+        X_train, X_test, y_train, y_test, feature_names = create_iris_pandas()
+        model = create_sklearn_svm_classifier(X_train, y_train)
+        model_task = ModelTask.CLASSIFICATION
+        categorical_features = []
+
+        # the index 1, corresponds to incorrect prediction
+        # the index 0 correspond to correct prediction
+        filters = [{'arg': arg,
```
imatiach-msft, line 151: I thought I had constants for all of these - arg, column, method

imatiach-msft: ah yes, there is:
```
from erroranalysis._internal.constants import (ARG, COLUMN, METHOD)
```
but it looks like the other tests also need to be updated in this file to use them

imatiach-msft (approved):

erroranalysis/tests/test_cohort_filter.py
```diff
@@ -275,5 +309,5 @@ def run_error_analyzer(validation_data,
                                        composite_filters)
 
     # validate there is some data selected for each of the filters
-    assert validation_data.shape[0] > 0
+    assert validation_data.shape[0] >= 0
```
imatiach-msft, line 312: This means the filter was so strict that all data was filtered out, if ```validation_data.shape[0] == 0```  - was there a specific test which this was the case for?  I'm just worried some code might change and our filtering breaks to filter everything out.  Perhaps, if this is needed, we can have a flag for the == 0 case?

erroranalysis/tests/test_cohort_filter.py
gaugup: Yeah, I have create a task internally to migrate these to module constants.

erroranalysis/tests/test_cohort_filter.py
gaugup: The multiclass classification model was giving exactly the same predictions as the test labels. That's why I had to update the assert condition. I am not passing a flag for cases when we want to assert on zero.

PR #1624, opened 2022-08-09
[300 line limit] Refactor CounterfactualChart

tongyu-microsoft:
This PR refactors CounterfactualChart, so line of code for CounterfactualChart is reduced from 938 lines to 249 lines.

## Description

For CounterfactualList.tsx, this PR creates a few new components as children of CounterfactualList component. So each file now has less than 300 lines of code.

This PR basically does the following:
1.Every new CounterfactualXXX components in the PR is originally from CounterfactualChart
2. Rename CounterfactualChart to CounterfactualComponent, as it does include CounterfactualChart, LocalImportanceChart and error dialog etc -> so CounterfactualComponent is the parent component in my PR.
3. CounterfactualComponent is parent of CounterfactualChartWithLegend, CounterfactualLocalImportanceChart and CounterfactualErrorDialog
4. CounterfactualChartWithLegend is parent of CounterfactualChart and CounterfactualChartLegend
5. Moved all getXXX functions which doesn't contain any setState in the function, to src/util folder.

## Checklist


libs/counterfactuals/src/lib/CounterfactualChartLegend.tsx
```diff
@@ -0,0 +1,152 @@
...
+              };
+            })}
+          />
+        )}
+      </Stack>
+    );
+  }
+
+  private getTargetDescription(): string {
+    if (this.context.dataset.task_type === "regression") {
```
imatiach-msft, line 106: can we move task type values like "regression" to global strings, or do such constants exist already?

libs/counterfactuals/src/lib/CounterfactualChartWithLegend.tsx
```diff
@@ -0,0 +1,189 @@
...
+      this.logTelemetryEvent(
+        TelemetryEventName.CounterfactualCreateWhatIfCounterfactualClick
+      );
+    }
+    this.setState((preState) => {
+      return { isPanelOpen: !preState.isPanelOpen };
+    });
+  };
+
+  private logTelemetryEvent = (eventName: TelemetryEventName) => {
```
imatiach-msft, line 183: I see this method copied several times, should it live in some global shared utils?

libs/counterfactuals/src/lib/CounterfactualComponent.tsx
```diff
@@ -0,0 +1,295 @@
...
+  ): void => {
+    if (!this.temporaryPoint || (!value && !index)) {
+      return;
+    }
+    const editingData = this.temporaryPoint;
+    if (index !== undefined) {
+      // User selected/de-selected an existing option
+      editingData[key] = index;
+    }
+    this.forceUpdate();
```
imatiach-msft, line 292: why is the forceUpdate needed before fetchData?  I see this above as well.

libs/counterfactuals/src/util/getCurrentLabel.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export function getCurrentLabel(
+  taskType: "classification" | "regression" | "text_classification",
```
imatiach-msft, line 5: is there some way to move these to constants?

libs/counterfactuals/src/util/getCurrentLabel.ts
```diff
@@ -0,0 +1,14 @@
...
+  desiredRange?: [number, number],
+  desiredClass?: string
+): string {
+  if (taskType === "regression") {
```
imatiach-msft, line 9: same here, move task type "regression" to some global constants

imatiach-msft (approved):

libs/counterfactuals/src/lib/CounterfactualChartLegend.tsx
```diff
@@ -0,0 +1,152 @@
...
+      this.props.toggleSelectionOfPoint(index);
+      this.logTelemetryEvent(
+        TelemetryEventName.CounterfactualNewDatapointSelectedFromDropdown
+      );
+    }
+  };
+
+  private disableCounterfactualPanel = (): boolean => {
+    return (
+      this.props.selectedPointsIndexes[0] === undefined ||
```
imatiach-msft, line 128: I see comparison to undefined in several places, but I recall Ke telling me we should not compare to null or undefined.  Perhaps it's fine to keep it as is, especially since this is just moving code around, but have you heard of this general rule?

vinuthakaranth (approved):

libs/counterfactuals/src/lib/CounterfactualChartLegend.tsx
tongyu-microsoft: Created an enum DatasetTaskTypes in core-ui. Thanks!

libs/counterfactuals/src/util/getCurrentLabel.ts
tongyu-microsoft: Updated. Created an enum DatasetTaskType in core-ui. Thanks!

libs/counterfactuals/src/util/getCurrentLabel.ts
tongyu-microsoft: Updated to DatasetTaskType.Regression

libs/counterfactuals/src/lib/CounterfactualComponent.tsx
tongyu-microsoft: This PR simply refactors the CounterfactualChart component without editing the detailed methods in the component, so we can make sure we don't block the logic while refactoring the component. The forceUpdate was there before.

But yeah it's worthwhile to look into how we can improve this by removing forceUpdate if it doesn't do anything here.
Created a task and I will look into this later: https://msdata.visualstudio.com/Vienna/_queries/edit/1923957/?triage=true
Thanks!!

libs/counterfactuals/src/lib/CounterfactualChartWithLegend.tsx
tongyu-microsoft: Great comment! Yeah there is already a work item to track it: [Task 1893214](https://msdata.visualstudio.com/Vienna/_workitems/edit/1893214): Create a context for telemetryHook

libs/counterfactuals/src/lib/CounterfactualChartLegend.tsx
tongyu-microsoft: Yeah we have this @typescript-eslint/no-null-keyword rule: https://palantir.github.io/tslint/rules/no-null-keyword/
This rule asks us to only use undefined, instead of null

`JavaScript originally intended undefined to refer to a value that doesn’t yet exist, while null was meant to refer to a value that does exist but points to nothing. To remove confusion over the two similar values, it’s better to stick with just undefined.`

PR #1626, opened 2022-08-10
Add identity feature value to counterfactual hover over

gaugup:
## Description

This PR adds the value in the identity feature column to the hover over in the counterfactual scatter plot chart.

## Checklist

image


libs/core-ui/src/lib/util/JointDataset.ts
```diff
@@ -645,6 +645,34 @@ export class JointDataset {
...
+    }
+    let jointDatasetFeatureName = undefined;
+    let userDatasetFeatureName = undefined;
+    for (jointDatasetFeatureName in this.metaDict) {
+      if (
+        this.metaDict[jointDatasetFeatureName]
+          .abbridgedLabel === userFeatureName
+      ) {
+        userDatasetFeatureName =
+          this.metaDict[jointDatasetFeatureName]
```
tongyu-microsoft, line 661: Can this be shorten by `userFeatureName` as `this.metaDict[jointDatasetFeatureName].abbridgedLabel === userFeatureName`

libs/counterfactuals/src/lib/CounterfactualChart.tsx
```diff
@@ -779,6 +780,40 @@ export class CounterfactualChart extends React.PureComponent<
         trace.y = rawY;
       }
     }
+    if (
+      this.context.jointDataset.datasetMetaData !== undefined &&
+      this.context.jointDataset.datasetMetaData.featureMetaData !== undefined
```
tongyu-microsoft, line 785: can this be shorten by `this.context.jointDataset.datasetMetaData?.featureMetaData !== undefined`? So we don't need line 784

tongyu-microsoft (approved):

libs/core-ui/src/lib/util/JointDataset.ts
```diff
@@ -645,6 +645,34 @@ export class JointDataset {
     });
   }
 
+  public getJointDatasetFeatureName(userFeatureName: string | undefined): string | undefined {
+    // Return the joint dataset feature name for the given user feature name.
+    if (userFeatureName === undefined) {
+      return undefined;
```
imatiach-msft, line 651: Maybe just not accept undefined then?  It seems strange to accept it in this method and return undefined.  Perhaps it might be better to just force the caller to make sure the parameter is defined before passing it in here, and handle the case with error/etc if the feature name is not defined in the code prior to this method?

libs/core-ui/src/lib/util/JointDataset.ts
```diff
@@ -645,6 +645,34 @@ export class JointDataset {
...
+      }
+    }
+
+    if (
+      userDatasetFeatureName !== undefined &&
+      jointDatasetFeatureName !== undefined
+    ) {
+      return jointDatasetFeatureName;
+    }
+    return undefined;
```
imatiach-msft, line 673: Perhaps it might be better to just throw in that case, since this is not expected when calling this utility method?

imatiach-msft (approved):

libs/counterfactuals/src/lib/CounterfactualChart.tsx
gaugup: Done. Thanks for the suggestion.

libs/core-ui/src/lib/util/JointDataset.ts
gaugup: Thanks for the suggestion. Simplified the flow of the function to not take undefined value as input parameter.

libs/core-ui/src/lib/util/JointDataset.ts
gaugup: I don't want throw error as this will leave the dashboard in an unusable state. The undefined is to be handled by the caller gracefully.

libs/core-ui/src/lib/util/JointDataset.ts
gaugup: Simplified the logic in the new iteration.

PR #1631, opened 2022-08-11
These two should not be comments.

DavisTJoseph:
It's a better user experience without commenting these two lines as you can copy and paste them directly in your jupyter notebook.

## Checklist

### DIFF
--- a/notebooks/individual-dashboards/fairness-dashboard/fairness-dashboard-loan-allocation.ipynb
+++ b/notebooks/individual-dashboards/fairness-dashboard/fairness-dashboard-loan-allocation.ipynb
@@ -26,8 +26,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


notebooks/individual-dashboards/fairness-dashboard/fairness-dashboard-loan-allocation.ipynb
```diff
@@ -26,8 +26,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
```
gaugup, line 28: Could you instead add a comment asking users to uncomment the two lines before running the notebook? Not sure how our test infrastructure responds if these packages are installed from the notebook. And wouldn't users need to restart the kernel when these packages are installed from the notebook?

imatiach-msft:
closing old PR as there was no response - please re-open if this should be reviewed again

PR #1632, opened 2022-08-11
[Accessibility] fix several accessibility bugs

RubyZ10:

## Description

This PR fix below bugs

[Bug 1919764](https://msdata.visualstudio.com/Vienna/_workitems/edit/1919764): A11yUsable - Feature importance - mouse centric language
**Before change:**
image

**After change:**
image

[Task 1919020](https://msdata.visualstudio.com/Vienna/_workitems/edit/1919020): 3. Bubble state colors & outline
**Before change:**
image

**After change:**
image

[Task 1919029](https://msdata.visualstudio.com/Vienna/_workitems/edit/1919029): 4. Feature/Value pill
**Before change:**
image

**After change**
image

[Task 1925557](https://msdata.visualstudio.com/Vienna/_workitems/edit/1925557): Error analysis tree keyboard accessible
**Before:**
the error analysis tree is not keyboard accessible, when navigating with tab, it pass the tree, goes to Model overview directly
errorAnalysisTabNav2

**After:**
errorAnalysisTabNav1

[Bug 1919759](https://msdata.visualstudio.com/Vienna/_workitems/edit/1919759): A11yMAS - [Keyboard accessible]Add component dialog on dashboard isn't keyboard accessible
**Before change:**
after the add component callout is open, click on tab, it goes to model overview, instead of opening the componet seletion
add1

**After change:**
add2

[Task 1923223](https://msdata.visualstudio.com/Vienna/_workitems/edit/1923223): Error analysis screen reader

## Checklist


libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/TreeViewRenderer/TreeViewNode.tsx
```diff
@@ -33,13 +33,23 @@ export class TreeViewNode extends React.Component<ITreeViewNodeProps> {
     const classNames = treeViewRendererStyles();
     const gradientFillId = getRandomId();
     const theme = getTheme();
+    const nodeFilterProps = node.data.filterProps;
+    const ariaLabel = `${nodeFilterProps.numCorrect} ${
+      nodeFilterProps.numIncorrect
+    } ${nodeFilterProps.errorCoverage.toFixed(
+      2
+    )} ${nodeFilterProps.metricValue.toFixed(2)}`; // TODO: waiting for PM input
```
RubyZ10, line 41: This ariaLabel string is temporary, waiting for PM's input on what the screen reader should announce when traversing each tree node

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/TreeViewRenderer/TreeViewNode.tsx
```diff
@@ -33,13 +33,31 @@ export class TreeViewNode extends React.Component<ITreeViewNodeProps> {
...
+    }
     return (
       <>
         <g
+          onFocus={this.onSelect}
           style={node.data.nodeState.style}
           onClick={this.onSelect}
           pointerEvents="all"
           ref={this.ref}
+          tabIndex={0}
```
imatiach-msft, line 58: It seems this is currently tabbing through the nodes, but I think we want to just go to the current selected node and then, if the user wants to change nodes, they can use up/down arrows - however we shouldn't be going to every single node in the tree when pressing tab key, which is what I am seeing in the dashboard app right now.

imatiach-msft: This can also be fixed in a separate PR though, doesn't need to be addressed right away.

imatiach-msft (approved):

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/TreeViewRenderer/TreeViewNode.tsx
imatiach-msft: maybe one of these might help:
https://codereview.stackexchange.com/questions/243627/handling-arrow-keys-and-tabindex
https://stackoverflow.com/questions/39733783/tab-to-next-tabindex-field-on-down-arrow-key

RubyZ10: Thanks Ilya, Will look into arrow navigation after we finished everything(accessibility) required by GA

vinuthakaranth (approved):

PR #1643, opened 2022-08-16
add wrapping for long text documents to text highlighting component in RAI dashboard for text

imatiach-msft:
## Description

During testing it was found that large documents don't display well in the RAI text dashboard because there is no wrapping.  This PR adds text wrapping to the text highlighting component to resolve this issue for long text documents and adds a test with a long text document.
I also refactored part of applications.ts file into a new file applicationInterfaces.ts as the applications.ts file was over 300 lines after adding the additional test case to the test app.

Before, without wrapping:

image

After, with wrapping:

image

## Checklist


libs/interpret-text/src/lib/TextExplanationDashboard/Control/TextHighlighting/TextHighlighting.styles.ts
```diff
@@ -28,6 +35,7 @@ export const textHighlightingStyles: () => IProcessedStyleSet<ITextHighlightingS
         {
           color: getPrimaryChartColor(theme),
           fontSize: theme.fonts.large.fontSize,
+          margin: "2px",
```
imatiach-msft, line 38: When I added wrap to FluentUI Stack, it removed the margin for Labels but not Text, which squished the text with underlines together (but not the highlighted or plain text) - hence I needed to add the margin here.  I wonder if it might just be better to always use text and sometimes set fontweight instead of using Label, might be a good discussion point to bring up with Ke when he is back, since he preferred to use Label for bold words.

vinuthakaranth (approved):

PR #1644, opened 2022-08-16
Handle 'Regression Error' cohort filter for regression scenario in SDK

gaugup:
## Description

It seems we don't handle the "Error" (aka Regression Error) cohort filter in erroranalysis SDK. This PR adds this capability and the relevant test.

## Checklist

### DIFF
--- a/erroranalysis/erroranalysis/_internal/cohort_filter.py
+++ b/erroranalysis/erroranalysis/_internal/cohort_filter.py
@@ -16,6 +16,7 @@ from erroranalysis._internal.utils import is_spark
 
 MODEL = 'model'
 CLASSIFICATION_OUTCOME = 'Classification outcome'
+REGRESSION_ERROR = 'Error'
 
 
 def filter_from_cohort(analyzer, filters, composite_filters,
@@ -141,13 +142,33 @@ class FilterDataWithCohortFilters:
         """
         model_task = self.model_task
         # For classification task, check if classification
-        # outcome included in filters, and if it is then
+        # outcome is included in filters, and if it is then
         # compute the necessary column data on the fly
         if model_task == ModelTask.CLASSIFICATION and filters:
             for filter in filters:
                 if COLUMN in filter:
                     column = filter[COLUMN]
                     if column == CLASSIFICATION_OUTCOME:
+                        return True
+        return False
+
+    def _filters_has_regression_error(self, filters):
+        """Checks if regression error is specified as a filter.
+
+        :param filters: The filters.
+        :type filters: list[dict]
+        :return: True if regression filter is specified in the filters.
+        :rtype: bool
+        """
+        model_task = self.model_task
+        # For regression task, check if regression error
+        # is included in filters, and if it is then
+        # compute the necessary column data on the fly
+        if model_task == ModelTask.REGRESSION and filters:
+            for filter in filters:
+                if COLUMN in filter:
+                    column = filter[COLUMN]
+                    if column == REGRESSION_ERROR:
                         return True
         return False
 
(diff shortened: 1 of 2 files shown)


erroranalysis/erroranalysis/_internal/cohort_filter.py
```diff
@@ -151,6 +152,26 @@ def _filters_has_classification_outcome(self, filters):
                         return True
         return False
 
+    def _filters_has_regression_error(self, filters):
+        """Checks if regression error is specified as a filter.
+
+        :param filters: The filters.
+        :type filters: list[dict]
+        :return: True if regression filter is specified in the filters.
```
imatiach-msft, line 160: nit: (in doc string)
```regression filter -> regression error```
is that an improvement, what do you think?

File: erroranalysis/erroranalysis/_internal/cohort_filter.py [hunk 2]
```diff
@@ -188,6 +209,27 @@ def _compute_classification_outcome_data(self, true_y, pred_y,
                     classification_outcome.append(1)
         return classification_outcome
 
+    def _compute_regression_error_data(self, true_y, pred_y):
+        """Creates the regression error data.
+
```
imatiach-msft, line 214: Suggestion - maybe add to the doc:
```Computes the per instance absolute difference between the true and predicted values.```

File: erroranalysis/erroranalysis/_internal/cohort_filter.py [hunk 3]
```diff
@@ -223,6 +266,16 @@ def _add_filter_cols(self, df, filters):
...
                     self.true_y, pred_y, classes)
+        elif has_regression_error:
+            if PRED_Y in df:
+                pred_y = df[PRED_Y]
+            else:
+                # calculate directly via prediction on model
+                pred_y = self.model.predict(
+                    df.drop(columns=[TRUE_Y, ROW_INDEX]))
+            # calculate regression error and add to df
+            df[REGRESSION_ERROR] = \
```
imatiach-msft, line 277: nit: try to avoid \ perhaps this is a bit clearer:
```
df[REGRESSION_ERROR] = self._compute_regression_error_data(
    self.true_y, pred_y)
```

erroranalysis/tests/test_cohort_filter.py
```diff
@@ -226,7 +229,38 @@ def test_cohort_filter_index(self):
...
+
+    def test_cohort_filter_regression_error(self):
+        X_train, X_test, y_train, y_test, feature_names = \
+            create_diabetes_data()
+        X_train = pd.DataFrame(X_train, columns=feature_names)
+        X_test = pd.DataFrame(X_test, columns=feature_names)
+
+        # filter on regression error, which can be done from the
+        # RAI dashboard
+        filters = [{'arg': [40],
```
imatiach-msft, line 249: nit: use arg, column, method from constants in erroranalysis package:

https://github.com/microsoft/responsible-ai-toolbox/blob/main/erroranalysis/erroranalysis/_internal/constants.py#L6

imatiach-msft (approved):
LGTM!  Can you also bump up version of erroranalysis so we can release this right away?

erroranalysis/erroranalysis/_internal/cohort_filter.py
gaugup: done.

erroranalysis/erroranalysis/_internal/cohort_filter.py
gaugup: Done

erroranalysis/erroranalysis/_internal/cohort_filter.py
gaugup: Done.

erroranalysis/tests/test_cohort_filter.py
gaugup: I have a separate work item for this. Will address this in a future PR.

PR #1646, opened 2022-08-16
[Accessibility] Add headings & change counterfactual text color

RubyZ10:

## Description

This PR fix two bugs:
[Bug 1919769](https://msdata.visualstudio.com/Vienna/_workitems/edit/1919769): A11yUsable - Counterfactuals - bold green sections in the table would be less clear to a color blind user

Before fix:
image

After fix:
image

[Bug 1919775](https://msdata.visualstudio.com/Vienna/_workitems/edit/1919775): A11yMAS - Headings should be used to convey structure and help screen readers navigate

Before fix:
image
After fix:
image

## Checklist

### DIFF
--- a/libs/counterfactuals/src/lib/CounterfactualList.styles.ts
+++ b/libs/counterfactuals/src/lib/CounterfactualList.styles.ts
@@ -32,7 +32,7 @@ export const counterfactualListStyle: () => IProcessedStyleSet<ICounterfactualLi
         }
       },
       editCell: {
-        color: theme.palette.green,
+        color: theme.semanticColors.bodyText,
         fontWeight: "bold"
       },
       highlightRow: {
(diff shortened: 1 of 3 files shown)


imatiach-msft (approved):

libs/counterfactuals/src/lib/CounterfactualList.styles.ts
```diff
@@ -32,7 +32,7 @@ export const counterfactualListStyle: () => IProcessedStyleSet<ICounterfactualLi
         }
       },
       editCell: {
-        color: theme.palette.green,
+        color: theme.semanticColors.bodyText,
```
imatiach-msft, line 35: is there some more interesting color we could use here?  or is it really recommended to just use regular body text for accessibility?

gaugup (approved):

libs/counterfactuals/src/lib/CounterfactualList.styles.ts
RubyZ10: color is useless for people with vision issue, suggest in future if we need to convey different information, we can use other signals instead of color, in this PR's table it only indicates a difference between the first row and the rest of the information. So just change it to black per suggestion from designer.

tongyu-microsoft (approved):

PR #1649, opened 2022-08-17
[Accessibility] Error analysis tree node change 1. Bubble error rate colors; 2. Bubble text & fill

RubyZ10:
## Description

This PR includes two changes:

1. change the color of error analysis tree node
2. add background for the text in tree node
Before change:
image

After change:
image

## Checklist


libs/error-analysis/src/lib/ErrorAnalysisDashboard/ColorPalette.ts
```diff
@@ -20,6 +20,19 @@ export class ColorPalette {
   public static ErrorAnalysisDarkBlackText = "rgba(0,0,0,0.8)";
   public static ErrorAnalysisDarkGreyText = "#555";
   public static DisabledColor = theme.palette.neutralSecondary;
+  public static NodeFilledColor = "#F3F2F1";
```
RubyZ10, line 23: Note: per design decision, we want the node shows the same color regardless of themes, so hard-coded the bubble colors

tongyu-microsoft (approved):

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/TreeViewRenderer/TreeViewUtils.ts
```diff
@@ -0,0 +1,41 @@
+import { ColorPalette } from "./../../ColorPalette";
+
+export function nodeColor(isErrorMetric: boolean, perc: number): string {
+  if (perc == 0) {
+    return ColorPalette.white;
+  }
+  if (isErrorMetric) {
+    if (perc > 0.75) {
```
tongyu-microsoft, line 8: nit: will changing it to a `switch` statement be better?

imatiach-msft (commented):
I created a chat to discuss this.   I do think it might be more difficult to understand the tree in the case the error rates are all within a band, say the 75-100 percent error rate band - all of the nodes will be the exact same color then.  I think this might need a bit more discussion, as there are some scenarios where the new UI look might become more confusing.

libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/TreeViewRenderer/TreeViewUtils.ts
RubyZ10: thanks tong, the file is not needed anymore.

imatiach-msft (approved):

PR #1651, opened 2022-08-18
Enable validations for FeatureMetadata when passed into RAIInsights

gaugup:
## Description

It seems like the validation was happening in the function `_validate_rai_insights_input_parameters` but the `feature_metadata` was not passed to the function. Now passing this parameter and added a test.

## Checklist


responsibleai/tests/test_feature_metadata.py
```diff
@@ -32,7 +32,7 @@ def test_feature_metadata_with_identity_feature(self):
         assert feature_metadata.dropped_features is None
         with pytest.raises(
                 UserConfigValidationException,
-                match='Identity feature name id given is not present'
+                match='The given identity feature name id, is not present'
```
imatiach-msft, line 35: nit: remove the "," in the original error message and in the check here, eg:
```
The given identity feature name id is not present in user features.
```

imatiach-msft (approved):

responsibleai/tests/test_feature_metadata.py
gaugup: done.

kicha0 (approved):

PR #1667, opened 2022-08-24
[Accessibility] Data analysis scatter chart - use symbol and color to differentiate data series

RubyZ10:
This PR made change to Data analysis scatter chart, before this change, we use 13 colors to differentiate different data series, but it does not work for people with vision issues. 
With the change, we introduce five different symbols combine with three colors
image

Before change
image

After change
image
Dark theme
image
Dark theme with two data series
image

## Checklist


libs/core-ui/src/lib/components/InteractiveLegend.styles.ts
```diff
@@ -11,85 +11,125 @@ import {
...
+    diamondColorBox: {
+      backgroundColor: color,
+      cursor: "pointer",
+      display: "inline-block",
+      height: "11px",
+      margin: "11px 4px 11px 8px",
+      opacity: activated ? 1 : 0.4,
+      transform: "rotate(45deg)",
+      width: "11px"
+    },
```
gaugup, line 67: Could you place shape definitions together?

libs/core-ui/src/lib/components/InteractiveLegend.tsx
```diff
@@ -83,4 +88,27 @@ export class InteractiveLegend extends React.PureComponent<IInteractiveLegendPro
...
+      case index % 5 === 1:
+        return classes.squareColorBox;
+      case index % 5 === 2:
+        return classes.diamondColorBox;
+      case index % 5 === 3:
+        return classes.triangleColorBox;
+      case index % 5 === 4:
+        return classes.triangleDownColorBox;
+      default:
+        return classes.circleColorBox;
```
gaugup, line 111: Seems like the code on line 111 will never execute if index >= 0. Could you use if/else instead of switch case? 

Also, does passing true in switch case work?

libs/dataset-explorer/src/lib/ScatterUtils.ts
```diff
@@ -0,0 +1,18 @@
...
+  return colors;
+}
+
+export function getScatterSymbols(): string[] {
+  const symbols = [];
+  for (let i = 0; i < 3; i++) {
+    symbols.push("circle", "square", "diamond", "triangle", "triangle-down");
+  }
+  return symbols;
+}
```
gaugup, line 18: Could you add a simple unit test file for these two functions?

libs/core-ui/src/lib/components/InteractiveLegend.tsx
```diff
@@ -83,4 +88,27 @@ export class InteractiveLegend extends React.PureComponent<IInteractiveLegendPro
       </div>
     );
   }
+
+  private getColorBoxCss = (
```
gaugup, line 92: Could you add a unit test for this function?

gaugup (changes requested):

libs/core-ui/src/lib/components/InteractiveLegend.styles.ts
RubyZ10: Could you please help elaborate what does this mean? Thanks!

libs/core-ui/src/lib/components/InteractiveLegend.tsx
RubyZ10: I can remove case 0 to deduplicate, and in this case, the guidance is using switch case rather than if/else as it has a few cases if I remember correctly. 
Passing true works in switch case

libs/dataset-explorer/src/lib/ScatterUtils.ts
RubyZ10: added, thanks!
image

libs/core-ui/src/lib/components/InteractiveLegend.tsx
RubyZ10: Moved to util file and added unit test thanks! 
image

libs/core-ui/src/lib/components/InteractiveLegendUtils.test.ts
```diff
@@ -0,0 +1,13 @@
+import { interactiveLegendStyles } from "./InteractiveLegend.styles";
+import { getColorBoxClassName } from "./InteractiveLegendUtils";
+
+describe("InteractiveLegend", () => {
+  it("should get the correct colorBox className", () => {
+    const classes = interactiveLegendStyles();
+    expect(getColorBoxClassName(0)).toEqual(classes.circleColorBox);
```
imatiach-msft, line 7: nit: might be cleaner to create array of shapes and loop over check, something like this (coded without any checking):
```
shapes = [classes.circleColorBox, classes.squareColorBox, classes.diamondColorBox, classes.triangleColorBox, classes.triangleDownColorBox]
shapes.forEach((shape, idx) => {
    expect(getColorBoxClassName(idx)).toEqual(shape)
})
```

libs/core-ui/src/lib/components/InteractiveLegendUtils.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export function getColorBoxClassName(
+  index: number,
+  color?: string,
+  activated?: boolean
+): string {
+  const classes = interactiveLegendStyles(activated, color);
+  // this is used as data series symbol in the side panel, the sequence needs to be consist with the sequence of symbols in ScatterUtils getScatterSymbols()
+  switch (true) {
+    case index % 5 === 1:
```
imatiach-msft, line 14: nit: do the index % 5 outside of the switch case, so:
```
modIndex = index % 5
switch (modIndex) {
    case 1:
       ...
    case 2:
       ....
```

libs/dataset-explorer/src/lib/ScatterUtils.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export function getScatterColors(): string[] {
+  const colors = [];
+  for (let i = 0; i < 5; i++) {
+    colors.push("#6D8EF7", "#CA397E", "#F3B33E");
```
imatiach-msft, line 7: I think these need to be consistent and come from the global FluentUI color palette here:
https://github.com/microsoft/responsible-ai-toolbox/blob/main/libs/core-ui/src/lib/util/FluentUIStyles.ts#L244
we want to make all of our charts consistent.
I tried to make all of our components as consistent as possible based on Ke's feedback but might have missed some.  We need to make all of our colors similar across charts.

imatiach-msft (approved):
Approved but the most important thing is to make sure colors are consistent and you are using the global FluentUI paletter we have in core-ui.  This was a task given to me and we need to make sure we continue to use consistent colors throughout the application's components.

imatiach-msft:
On the colors, please see this PR for more details:
https://github.com/microsoft/responsible-ai-toolbox/pull/1559
I think we should either change the current PR's colors to be consistent with that or change the global color palette to use these colors.

libs/core-ui/src/lib/components/InteractiveLegend.tsx
gaugup: Maybe you write a simple if/else. The above code doesn't look like using the switch correctly.

RubyZ10: changed to 
```
const modIndex = index % 5;
  switch (modIndex) {
    case 1:
      return classes.squareColorBox;
    case 2:
      return classes.diamondColorBox;
    case 3:
      return classes.triangleColorBox;
    case 4:
      return classes.triangleDownColorBox;
    default:
      return classes.circleColorBox;
  }
```

Does this help?

libs/core-ui/src/lib/components/InteractiveLegendUtils.test.ts
RubyZ10: updated, thanks!

libs/core-ui/src/lib/components/InteractiveLegendUtils.ts
RubyZ10: updated thanks!

libs/core-ui/src/lib/components/InteractiveLegendUtils.test.ts
```diff
@@ -0,0 +1,13 @@
+import { interactiveLegendStyles } from "./InteractiveLegend.styles";
```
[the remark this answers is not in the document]
RubyZ10: fixed, thanks!

imatiach-msft (approved):

gaugup (approved):

libs/core-ui/src/lib/components/InteractiveLegend.styles.ts
gaugup: I think linting will cause these names to be sorted alphabetically. This is ok.

vinuthakaranth (approved):

vinuthakaranth (approved):

PR #1673, opened 2022-08-29
refactor table view into data analysis component

imatiach-msft:
## Description

Refactor the table view into the data analysis component from the individual feature importances component.
Note there will be a follow-up PR to add a "mini-table" in the feature importances component to select a subset of indexes from the overall selected indexes in the tabular view.  Also, we will still show the table by default if the dataset explorer component is missing in the follow-up PR.

New data analysis view with new "Table view" tab:

image

New individual feature importances tab:

image

## Checklist


libs/dataset-explorer/src/lib/TableView/TableView.tsx
```diff
@@ -0,0 +1,294 @@
...
+            label={localization.formatString(
+              localization.ModelAssessment.FeatureImportances.SelectionCounter,
+              this.selection.count,
+              this.maxSelectable
+            )}
+            calloutTitle={undefined}
+            renderOnNewLayer
+            type="label"
+          >
+            <Text block>
```
RubyZ10, line 122: Can we not set text with block? with block,  when making screen too small , the overlfow part of the text will not be seen.

RubyZ10: Could you please try to change the screen size to 320px? If the text is still visible, we should be fine

libs/dataset-explorer/src/lib/TableView/TableView.tsx
```diff
@@ -0,0 +1,294 @@
...
+                    columns={this.state.columns}
+                    groups={this.state.groups}
+                    setKey="set"
+                    layoutMode={DetailsListLayoutMode.fixedColumns}
+                    constrainMode={ConstrainMode.unconstrained}
+                    onRenderDetailsHeader={this.onRenderDetailsHeader}
+                    selectionPreservedOnEmptyClick
+                    ariaLabelForSelectionColumn="Toggle selection"
+                    checkButtonAriaLabel="Row checkbox"
+                    // checkButtonGroupAriaLabel="Group checkbox"
```
RubyZ10, line 143: remove?

libs/dataset-explorer/src/lib/TableView/TableView.tsx
```diff
@@ -0,0 +1,294 @@
...
+                    items={this.state.rows}
+                    columns={this.state.columns}
+                    groups={this.state.groups}
+                    setKey="set"
+                    layoutMode={DetailsListLayoutMode.fixedColumns}
+                    constrainMode={ConstrainMode.unconstrained}
+                    onRenderDetailsHeader={this.onRenderDetailsHeader}
+                    selectionPreservedOnEmptyClick
+                    ariaLabelForSelectionColumn="Toggle selection"
+                    checkButtonAriaLabel="Row checkbox"
```
RubyZ10, line 142: wondering if we need to localize aria label?

libs/dataset-explorer/src/lib/TableView/TableView.tsx
```diff
@@ -0,0 +1,294 @@
...
+    return (
+      <div>
+        {defaultRender?.({
+          ...props,
+          onRenderColumnHeaderTooltip,
+          selectAllVisibility: SelectAllVisibility.hidden
+        })}
+      </div>
+    );
+  };
```
RubyZ10, line 265: Could you please create onRenderDetailsHeader.tsx and move the component there? Thanks

libs/dataset-explorer/src/lib/TableView/TableView.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+  private _onRenderGroupHeader = (props?: IDetailsGroupDividerProps) => {
+    const classNames = tableViewStyles();
```
RubyZ10, line 268: Could you please create a onRenderGroupHeader.tsx and move this component there? Thanks!

RubyZ10 (approved):

libs/dataset-explorer/src/lib/TableView/TableView.tsx
imatiach-msft: removed block, thanks!

libs/dataset-explorer/src/lib/TableView/TableView.tsx
imatiach-msft: this was in the original file where I refactored the code from, but I removed it anyway

libs/dataset-explorer/src/lib/TableView/TableView.tsx
imatiach-msft: done!

libs/dataset-explorer/src/lib/TableView/TableView.tsx
imatiach-msft: done!

libs/dataset-explorer/src/lib/TableView/TableView.tsx
imatiach-msft: done, localized just in case both aria labels

PR #1674, opened 2022-08-29
Pass classes into the sklearn confusion_matrix

gaugup:
## Description

If `pred_y` and `true_y` have only one class, then call to `confusion_matrix` fails with the following error:-

```
>       tp, fp, fn, tn = confusion_matrix(y_true, y_pred, labels=classes).ravel()
E       ValueError: not enough values to unpack (expected 4, got 1)
```

Solution is to pass the list of classes into the `confusion_matrix`. This will make sure the above error is not raised.

## Checklist


erroranalysis/erroranalysis/_internal/metrics.py
```diff
@@ -14,45 +14,68 @@
...
+    :type y_true: numpy.ndarray
+    :param y_pred: Predicted labels.
+    :type y_pred: numpy.ndarray
+    :param classes: List of classes.
+    :type classes: list
+    :return: Tuple of true positive, false positive, false negative
+             and true negative.
+    :rtype: Tuple[int, int, int, int]
+    """
+    tp, fp, fn, tn = confusion_matrix(y_true, y_pred, labels=classes).ravel()
```
imatiach-msft, line 30: should we check if classes matches the values in true y and pred y lists?  what happens in the case they don't match - eg, classes is "['dog', 'cat', 'bird']" but the values are [0,1,1,2,1,2,0,1...]?

imatiach-msft (approved):

erroranalysis/erroranalysis/_internal/metrics.py
gaugup: For the code below the `confusion_matrix` method raises an exception:-

```
>>> from sklearn.metrics import confusion_matrix
>>> y_true = [0,1,0,1]
>>> y_pred = [0,1,1,0]
>>> classes = ["dog", "cat"]
>>> tp, fp, fn, tn = confusion_matrix(y_true, y_pred, labels=classes).ravel()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\gaugup\Anaconda3\envs\rai\lib\site-packages\sklearn\metrics\_classification.py", line 316, in confusion_matrix
    raise ValueError("At least one label specified must be in y_true")
ValueError: At least one label specified must be in y_true
```

Looks like scikit-learn handles the labels by checking the values against the labels.

imatiach-msft: sounds good

PR #1679, opened 2022-08-31
JSON serialize the endpoint output and add tests

gaugup:
## Description

It seems like the end point output should be serializable by JSON. This PR does the object serialization via `serialize_json_safe` and adds a test to make sure that the end point output is JSON serialized.

## Checklist


raiwidgets/tests/common_utils.py
```diff
@@ -12,6 +14,9 @@ def check_success_criteria(self, flask_server_prediction_output):
         assert WidgetRequestResponseConstants.error not in \
             flask_server_prediction_output
 
+        json.dumps(flask_server_prediction_output[
```
vinuthakaranth, line 17: Is this test util file or actual file? Asking since it is under _tests/_ folder

vinuthakaranth (approved):

tongyu-microsoft (approved):

raiwidgets/tests/common_utils.py
gaugup: Yes, this is in the test folder just doing the validation that the output of the dashboard is json serialized.

imatiach-msft (approved):
approved but can you change code owners so I'm not mandatory?

PR #1681, opened 2022-09-01
Model assessment vision

jamesbchao:
**FYI: 2,100 of the 2,374 file changes are just adding mock data, since all the class names have to be present now that I'm using indices for predicted y and true y.**

Adds a model assessment vision tab to the test app so that users can more accurately visualize what the actual app would look like when connected to the backend. 

Additionally, converts all the vision data into the standard IDataset so that the data can be used with the Model Overview and Data Explorer tabs. Additionally adds support for external metadata, in this case mean pixel value. 

Screenshot (56)
Screenshot (57)

## Checklist


imatiach-msft (approved):

apps/dashboard/src/app/applications.ts
```diff
@@ -265,5 +269,16 @@ export const applications: IApplications = <const>{
       } as IModelAssessmentDataSet
     },
     versions: { "1": 1, "2:Static-View": 2 }
+  },
+  modelAssessmentVision: {
+    datasets: {
+      visionModelExplanationData: {
+        classDimension: 3,
+        dataset: emotion,
+        modelExplanationData: [emotionModelExplanationData],
```
imatiach-msft, line 278: emotionModelExplanationData?  this looks wrong?

imatiach-msft: Discussed, this can be removed. Also we talked about moving out the visionModelExplanationData properties into dataset instead.

PR #1683, opened 2022-09-01
for RAI text dashboard, remove radio buttons and legend text when avg of abs value local importances selected

imatiach-msft:
## Description

For the RAI text dashboard, the radio buttons to select positive/negative/all importances are not relevant when the option "Average of absolute value" is selected for class importance weights.  Hence, this PR removes those options to simplify the dashboard for users.

View of RAI text dashboard when "Average of absolute value" is selected:
image

View of RAI text dashboard when any class is selected:
image

## Checklist

### DIFF
--- a/libs/interpret-text/src/lib/TextExplanationDashboard/Control/TextExplanationView/TextExplanationView.tsx
+++ b/libs/interpret-text/src/lib/TextExplanationDashboard/Control/TextExplanationView/TextExplanationView.tsx
@@ -148,19 +148,23 @@ export class TextExplanationView extends React.PureComponent<
                   weightLabels={this.props.weightLabels}
                 />
               </Stack.Item>
-              <Stack.Item id="TextChoiceGroup">
-                <ChoiceGroup
-                  defaultSelectedKey="all"
-                  options={options}
-                  onChange={this.changeRadioButton}
-                  required
-                />
-              </Stack.Item>
-              <Stack.Item>
-                <Text variant={"small"}>
-                  {localization.InterpretText.View.legendText}
-                </Text>
-              </Stack.Item>
+              {this.props.selectedWeightVector !== WeightVectors.AbsAvg && (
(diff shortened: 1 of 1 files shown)


vinuthakaranth (approved):

libs/interpret-text/src/lib/TextExplanationDashboard/Control/TextExplanationView/TextExplanationView.tsx
```diff
@@ -148,19 +148,23 @@ export class TextExplanationView extends React.PureComponent<
...
+                  />
+                </Stack.Item>
+              )}
+              {this.props.selectedWeightVector !== WeightVectors.AbsAvg && (
+                <Stack.Item>
+                  <Text variant={"small"}>
+                    {localization.InterpretText.View.legendText}
+                  </Text>
+                </Stack.Item>
+              )}
```
RubyZ10, line 167: just curious, why not just do this?
```
{this.props.selectedWeightVector !== WeightVectors.AbsAvg && (
                <Stack.Item id="TextChoiceGroup">
                  <ChoiceGroup
                    defaultSelectedKey="all"
                    options={options}
                    onChange={this.changeRadioButton}
                    required
                  />
                </Stack.Item>
                <Stack.Item>
                  <Text variant={"small"}>
                    {localization.InterpretText.View.legendText}
                  </Text>
                </Stack.Item>
              )}
```

RubyZ10 (approved):

libs/interpret-text/src/lib/TextExplanationDashboard/Control/TextExplanationView/TextExplanationView.tsx
imatiach-msft: I can't put two <Stack.Item> without nesting them, typescript does not allow it.  However, putting them in a single div will break this parent stack/child stack.item code, and creating another stack is not desirable here.

PR #1709, opened 2022-09-13
Port over cohort filtering tests for 'Predicted Y' cohort to responsibleai

gaugup:
## Description

Port over cohort filtering tests for 'Predicted Y' cohort to responsibleai

## Checklist

### DIFF
--- a/responsibleai/tests/test_rai_insights_cohort_filter.py
+++ b/responsibleai/tests/test_rai_insights_cohort_filter.py
@@ -43,6 +43,50 @@ class TestCohortFilterRAIInsights(object):
                          model_task,
                          filters=filters)
 
+    @pytest.mark.skip("Skipping this test due to a bug condition "
+                      "in cohort filtering")
+    def test_cohort_filter_predicted_y(self):
+        X_train, X_test, y_train, y_test, feature_names = create_iris_pandas()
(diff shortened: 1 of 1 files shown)


responsibleai/tests/test_rai_insights_cohort_filter.py
```diff
@@ -43,6 +43,50 @@ def test_cohort_filter_equal(self):
...
+                         y_test,
+                         feature_names,
+                         categorical_features,
+                         model_task,
+                         filters=filters)
+
+    def test_cohort_filter_true_y(self):
+        X_train, X_test, y_train, y_test, feature_names = create_iris_pandas()
+        filters = [{'arg': [2],
+                    'column': 'True Y',
```
imatiach-msft, line 72: nit: these two test methods look exactly the same except for the column ('True Y' vs 'Predicted Y'), might it be better to refactor the code into one helper method, and call it from both tests but with the column name parameter being different?

imatiach-msft (approved):

responsibleai/tests/test_rai_insights_cohort_filter.py
gaugup: done

PR #1712, opened 2022-09-14
Add feature names to data sent to "requestPredictions" callback

gaugup:
## Description

When doing WhatIf analysis in counterfactual panel, the predict call (via `requestPredictions` callback) send raw data as plain array. This can be problematic as the order of values between the dashboard and SDK may not be consistent. Adding feature names to the data sent to the `requestPredictions` can help alleviate this problem. 

## Checklist

### DIFF
--- a/raiwidgets/tests/responsibleai_dashboard/test_responsibleai_dashboard_input_predict.py
+++ b/raiwidgets/tests/responsibleai_dashboard/test_responsibleai_dashboard_input_predict.py
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation
 # Licensed under the MIT License.
 
+import json
+
 from common_utils import CheckResponsibleAIDashboardInputTestResult
 
 from raiwidgets.responsibleai_dashboard_input import \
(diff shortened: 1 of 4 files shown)


raiwidgets/tests/responsibleai_dashboard/test_responsibleai_dashboard_input_predict.py
```diff
@@ -71,9 +86,14 @@ def test_rai_dashboard_input_iris_on_predict_success(
         test_data = ri.test
 
         dashboard_input = ResponsibleAIDashboardInput(ri)
-        test_pred_data = test_data.head(1).drop("target", axis=1).values
+
+        test_pred_data_end_point = json.loads(test_data.head(1).drop(
+            "target", axis=1).to_json(orient='records'))
+        test_pred_data = test_data.head(1).drop(
+            "target", axis=1)
```
romanlutz, line 93:
```suggestion
        test_pred_data = test_data.head(1).drop(
            "target", axis=1)
        test_pred_data_end_point = json.loads(test_pred_data.to_json(orient='records'))
        
```

PR #1714, opened 2022-09-14
fix lint rules

xuke444:

## Checklist

### DIFF
--- a/libs/core-ui/.eslintrc.json
+++ b/libs/core-ui/.eslintrc.json
@@ -16,9 +16,7 @@
     },
     {
       "files": ["*.ts", "*.tsx"],
-      "rules": {
-        "max-lines": "error"
-      }
+      "rules": {}
     },
     {
       "files": ["*.js", "*.jsx"],
--- a/libs/e2e/src/util/BoxChart.ts
+++ b/libs/e2e/src/util/BoxChart.ts
@@ -59,10 +59,13 @@ export class BoxChart extends Chart<IBox> {
       : { ...boxCoordinate };
   };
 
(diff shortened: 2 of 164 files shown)


libs/core-ui/.eslintrc.json
```diff
@@ -16,9 +16,7 @@
     },
     {
       "files": ["*.ts", "*.tsx"],
-      "rules": {
-        "max-lines": "error"
```
tongyu-microsoft, line 20: why do we remove this rule for core-ui? I didn't see we set global max-lines rule in this pr

libs/e2e/src/util/BoxChart.ts
```diff
@@ -59,10 +59,13 @@ export class BoxChart extends Chart<IBox> {
       : { ...boxCoordinate };
   };
 
-  private readonly getBoxCoordinate = (
+  private readonly getD = (
```
tongyu-microsoft, line 62: for my learning, what is this `D`?

tongyu-microsoft (approved):

libs/core-ui/.eslintrc.json
xuke444: It is in .eslintrc.custom.eslintrc as a warn.
I was planning to fix all in this pr, but seems it is too big already, I will fix all max line issues in following pr.

libs/e2e/src/util/BoxChart.ts
xuke444: The d value of svg element.
This is a extra function added to address cyclone complexity.
Seems git hub did not recognize it correctly.
getBoxCoordinate is still below

PR #1718, opened 2022-09-14
Fix "Clear” button of Search field under 'Feature List' dialog is not Functional and Focus indicator is not visible on all four sides for “Tree map” and "heat map" tabs.

vinuthakaranth:
## Description

Fix:
1. [Functional -Azure Machine Learning - Feature List>Search box]: "Clear” button of Search field under 'Feature List' dialog is not Functional. - onsearch was being called with no params resulting in no action. Removing onclear worked.
searchAccFix

3. [Usable - Azure Machine Learning - Model demo page>Error analysis]: Focus indicator is not visible on all four sides for “Tree map” and "heat map" tabs. - Since the above margin was consuming focus zone on adding margin to tabs, focus zone is visible. Similar design is present in Model overview component as well.

image
image

## Checklist

- [ ] I have added e2e tests for all UI changes. - NA

### DIFF
--- a/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/FeatureList/FeatureList.tsx
+++ b/libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/FeatureList/FeatureList.tsx
@@ -155,7 +155,7 @@ export class FeatureList extends React.Component<
   public render(): React.ReactNode {
     return (
       <Panel
-        headerText="Feature List"
+        headerText={localization.ErrorAnalysis.FeatureList.featureList}
         isOpen={this.props.isOpen}
         focusTrapZoneProps={focusTrapZoneProps}
         // You MUST provide this prop! Otherwise screen readers will just say "button" with no label.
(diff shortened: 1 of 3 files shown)


libs/error-analysis/src/lib/ErrorAnalysisDashboard/Controls/FeatureList/FeatureList.tsx
```diff
@@ -180,7 +180,9 @@ export class FeatureList extends React.Component<
                 placeholder="Search"
                 styles={searchBoxStyles}
                 onSearch={this.onSearch}
-                onClear={this.onSearch}
+                onClear={() => {
+                  this.onSearch("");
```
RubyZ10, line 184: Remove the onClear should fix the problem, btw try not using inline function, as we are going to remove all of them.  Try below, thanks!
```
<SearchBox
                placeholder="Search"
                styles={searchBoxStyles}
                onSearch={this.onSearch}
                onChange={(_, newValue?: string): void => {
                  if (newValue === undefined) {
                    return;
                  }
                  this.onSearch(newValue);
                }}
              />
```

vinuthakaranth: thanks, updated.

RubyZ10 (approved):

PR #1720, opened 2022-09-14
Raise exception in case categorical feature list doesn't have all string features

gaugup:
## Description

All the rai insight components fail if any of the string features are not identified as `categorical_features`. This PR raises an `UserConfigValidationException` in such a scenario. 

## Checklist


responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -342,6 +343,18 @@ def _validate_rai_insights_input_parameters(
...
                         )
 
+            train_features = train.drop(columns=[target_column]).columns
+            numeric_features = train.drop(
+                columns=[target_column]).select_dtypes(
+                    include='number').columns.tolist()
+            string_features_set = set(train_features) - set(numeric_features)
+            if len(string_features_set - set(categorical_features)) > 0:
+                raise UserConfigValidationException(
+                    "Looks like some string features {0} were not "
```
imatiach-msft, line 353: nit: maybe remove "Looks like" to make the error message more succinct/direct?

imatiach-msft: nit: and maybe move the feature(s) to the end and add a colon, since the set of features could be long or the feature names could be long which might make it harder to read the message?

imatiach-msft (approved):

RubyZ10 (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
gaugup: Done

PR #1721, opened 2022-09-15
clean up dataset explorer

xuke444:
clean up dataset explorer by reduce file size
use context in axis config instead of passing values

### DIFF
--- a/libs/dataset-explorer/src/lib/DatasetExplorerTab.tsx
+++ b/libs/dataset-explorer/src/lib/DatasetExplorerTab.tsx
@@ -5,14 +5,10 @@ import {
   getTheme,
   IDropdownOption,
   Dropdown,
-  DefaultButton,
   Text,
-  IChoiceGroupOption,
   Stack
 } from "@fluentui/react";
 import {
-  AxisConfigDialog,
-  JointDataset,
   ColumnCategories,
   cohortKey,
   ChartTypes,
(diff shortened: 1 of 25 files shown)


libs/dataset-explorer/src/lib/DatasetExplorerTab.tsx
```diff
@@ -304,68 +267,20 @@ export class DatasetExplorerTab extends React.Component<
...
-    newProps.chartType = item.key as ChartTypes;
-    if (newProps.yAxis.property === ColumnCategories.None) {
-      newProps.yAxis = this.generateDefaultYAxis();
-    }
-    this.setState({ chartProps: newProps });
-    this.logButtonClick(TelemetryEventName.DatasetExplorerNewChartTypeSelected);
-  };
-
   private logButtonClick = (eventName: TelemetryEventName): void => {
     this.props.telemetryHook?.({
```
tongyu-microsoft, line 271: Do we want to update all props.telemetryHook to context.telemetryHook? For example this line?

tongyu-microsoft (approved):

libs/dataset-explorer/src/lib/DatasetExplorerTab.tsx
xuke444: I am planning to wrap all buttons/input etc for automatically telemetry.

PR #1722, opened 2022-09-15
Enable cohort filtering for "Predicted Y"

gaugup:
## Description

This PR enables cohort filtering capability for "Predicted Y". It enables this by adding the predicted values to the dataframe before the cohort filtering logic using the pandas query function. This PR has the following changes:-

1. Add predicted y values in all cases in `cohort_filter.py`. If the predictions are available apriori then it adds those predictions.
2. Adds tests in test_cohort_filter.py to run tests from `PredictionAnalyzer` which takes predictions instead of the model.
3. Enables cohort filtering for `Predicted Y` case.
4. Other changes in erroranalysis package to handle scenario when predicted y is present on the filtered data.

## Checklist


RubyZ10 (dismissed):

erroranalysis/erroranalysis/_internal/cohort_filter.py
```diff
@@ -279,18 +279,21 @@ def _add_filter_cols(self, df, filters):
...
-
         if not is_spark(df):
             df[ROW_INDEX] = np.arange(0, len(self.true_y))
+
+        if self.pred_y is not None:
+            df[PRED_Y] = self.pred_y
+        else:
+            if not is_spark(self.dataset):
+                if not isinstance(self.dataset, pd.DataFrame):
+                    df[PRED_Y] = self.model.predict(self.dataset)
```
imatiach-msft, line 290: Here we may be calculating PRED_Y even if we don't need it for filtering?
That is why the calculation was done in has_classification_outcome instead previously, so we would only call predict (which is expensive) if we really need it:

```
        if has_classification_outcome:
            if PRED_Y in df:
                pred_y = df[PRED_Y]
            else:
                # calculate directly via prediction on model
                pred_y = self.model.predict(
                    df.drop(columns=[TRUE_Y, ROW_INDEX]))
```

imatiach-msft: We can just have a similar check to has_classification_outcome, such as ```has_predicted_Y```, if there is a "Predicted Y" filter passed here - and only then call the predict function.

erroranalysis/tests/test_cohort_filter.py
```diff
@@ -29,156 +30,180 @@ def test_cohort_filter_equal(self):
...
+        run_different_error_analyzers(validation_data,
+                                      model,
+                                      X_test,
+                                      y_test,
+                                      feature_names,
+                                      categorical_features,
+                                      model_task,
+                                      filters=filters)
+
+    # @pytest.mark.skip("Skipping this test due to a bug condition "
```
imatiach-msft, line 50: remove commented code here (to re-enable test)

erroranalysis/tests/test_cohort_filter.py
```diff
@@ -355,15 +392,47 @@ def create_iris_pandas():
...
     validation_data = X_test.copy()
     validation_data[TRUE_Y] = y_test
     validation_data[ROW_INDEX] = np.arange(0, len(y_test))
-    if pred_y is not None:
-        validation_data[PRED_Y] = pred_y
+    validation_data[PRED_Y] = pred_y
     return validation_data
 
 
+def run_different_error_analyzers(validation_data,
```
imatiach-msft, line 403: nit: maybe rename this to just
```run_error_analyzers```
I think this is shorter and more succinct, since "different" doesn't add more information to the method name

imatiach-msft (approved):

PR #1724, opened 2022-09-16
clean up fairness

xuke444:

## Checklist


libs/fairness/src/lib/Controls/ModelComparisonChartLoadData.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    return {
+      fairnessArray,
+      fairnessBounds,
+      performanceArray,
+      performanceBounds
+    };
+  } catch {
+    return undefined;
+    // todo;
```
tongyu-microsoft, line 86: remove?

libs/fairness/src/lib/WizardReportHelper.tsx
```diff
@@ -0,0 +1,202 @@
...
+      overpredictions,
+      performance,
+      predictions,
+      underpredictions
+    };
+
+    return metrics;
+  } catch {
+    return undefined;
+    // todo;
```
tongyu-microsoft, line 157: remove?

tongyu-microsoft (approved):

libs/fairness/src/lib/Controls/ModelComparisonChartLoadData.ts
xuke444: It is moved from old code, I don't know exactly what needs "to do" lol
Seems like we need error handling here.

PR #1727, opened 2022-09-16
[A11y]Show Error analysis tree focus && Add Screen reader for 1. New cohort panel, 2. Add component callout 3. Callout tooltip

RubyZ10:
## Description

1. Add Error analysis tree focus
image
2. for New cohort panel, after clicking on clear all filters, announce "no filters applied", after click on Add filter, announce, "filter added"
3. Add screen reader for add component callout
4. Update InfoCallout to support screen reader, this affect all places(basically all the tooltips) leverage InfoCallout

## Checklist


libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortEditorPanelContent.tsx
```diff
@@ -128,6 +129,12 @@ export class CohortEditorPanelContent extends React.PureComponent<
           <Link className={styles.clearFilter} onClick={this.clearAllFilters}>
             {localization.Interpret.CohortEditor.clearAllFilters}
           </Link>
+          <div
```
vinuthakaranth, line 132: workspace-portal repo uses something like below as a component to announce. Very similar to this but as a component. Do you think this could be better for code readability and understandability?
https://msdata.visualstudio.com/Vienna/_git/workspace-portal?path=%2Fcommon%2Fcomponents%2Fsrc%2Fcustom%2Fannounce%2FAnnounce.tsx&_a=contents&version=GBmain
or Moving div to a common Announce file.

vinuthakaranth (approved):

libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortEditorPanelContent.tsx
RubyZ10: Hi Vinu, it's actually different way of announcing the message, I can change to align with aml

RubyZ10: updated, thanks

PR #1729, opened 2022-09-17
Rename 'Error' filter name to 'Regression error'

gaugup:
## Description

Rename 'Error' filter name to 'Regression error'. UI sends 'Regression error' to send SDK.

image

## Checklist


imatiach-msft (approved):
Just to make sure, this doesn't break the UI in any way, as it looks like it might be a breaking change?  As in the UI doesn't expect to see Error anywhere instead of Regression error in the result right now?

gaugup:
> Just to make sure, this doesn't break the UI in any way, as it looks like it might be a breaking change? As in the UI doesn't expect to see Error anywhere instead of Regression error in the result right now?

I thought that UI sends "Error" but it was not the case. So this was not working as expected previously. With this change it works now as expected. There is not breaking change on UI side.

PR #1733, opened 2022-09-20
fix RAI dashboard for vision models crashing when selecting individual datapoints in data analysis

imatiach-msft:
## Description

Fix RAI dashboard for vision models crashing when selecting individual datapoints in data analysis.

This PR refactors the UI backend code to handle tabular, image and text multiclass scenarios more consistently with a helper method IsMulticlass.  We were previously only doing certain transformations for multiclass tabular case which was causing the crash to appear.

## Checklist

### DIFF
--- a/libs/core-ui/src/lib/Interfaces/IExplanationContext.ts
+++ b/libs/core-ui/src/lib/Interfaces/IExplanationContext.ts
@@ -9,7 +9,8 @@ export enum ModelTypes {
   Regression = "regression",
   Binary = "binary",
   Multiclass = "multiclass",
-  Image = "image"
+  ImageMulticlass = "imagemulticlass",
+  TextMulticlass = "textmulticlass"
 }
 
 export interface IExplanationContext {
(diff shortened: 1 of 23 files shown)


libs/core-ui/src/lib/Interfaces/IExplanationContext.ts
```diff
@@ -9,7 +9,25 @@ export enum ModelTypes {
...
+}
+
+export function IsClassifier(modelType: ModelTypes): boolean {
+  return (
+    modelType === ModelTypes.Binary ||
+    modelType === ModelTypes.Multiclass ||
+    modelType === ModelTypes.ImageMulticlass ||
+    modelType === ModelTypes.TextMulticlass
+  );
 }
```
xuke444, line 31: do you mind move these to util?
This interface folder suppose only have interface definition

xuke444 (approved):

libs/core-ui/src/lib/Interfaces/IExplanationContext.ts
imatiach-msft: done!

PR #1738, opened 2022-09-21
Fix cohort deletion not working

vinuthakaranth:
Fix cohort deletion not working
## Description

Since item key was passed as key prop, it was always undefined as React uses key as prop.

cohortdelete

## Checklist


RubyZ10 (approved):

imatiach-msft (approved):

libs/e2e/src/lib/describer/modelAssessment/dataExplorer/describeCohortFunctionality.ts
```diff
@@ -31,8 +31,21 @@ export function describeCohortFunctionality(
       cy.get(Locators.CancelButton).click();
       cy.get(Locators.YesButton).click();
     });
+
+    it("should be able to delete cohort", () => {
```
tongyu-microsoft, line 35: thanks for adding the E2E tests while fixing the bug!!

tongyu-microsoft (approved):

gaugup (approved):

PR #1742, opened 2022-09-23
in error analysis, refactor error rate metric into function

imatiach-msft:
## Description

In erroranalysis package, refactor the error rate metric computation into a function, to make it more similar to other metrics.

## Checklist

### DIFF
--- a/erroranalysis/erroranalysis/_internal/metrics.py
+++ b/erroranalysis/erroranalysis/_internal/metrics.py
@@ -171,6 +171,29 @@ def mean_prediction(y_true, y_pred):
     return np.mean(y_pred)
 
 
+def error_rate(y_true, y_pred, diff=None):
+    """Compute error rate for classification and regression tasks.
+
+    :param y_true: True labels.
+    :type y_true: numpy.ndarray
+    :param y_pred: Predicted values.
+    :type y_pred: numpy.ndarray
+    :param diff: Difference between true and predicted values.
+    :type diff: numpy.ndarray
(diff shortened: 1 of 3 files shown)


erroranalysis/erroranalysis/_internal/metrics.py
```diff
@@ -171,6 +171,24 @@ def mean_prediction(y_true, y_pred):
     return np.mean(y_pred)
 
 
+def error_rate(diff, total):
```
gaugup, line 174: Can we make its signature similar to other metric functions? 

```
def error_rate(y_pred, y_true):
```

tongyu-microsoft (approved):

erroranalysis/erroranalysis/_internal/metrics.py
imatiach-msft: diff is already pre-computed, so I think it would waste computation time to do this with y_pred and y_true?

gaugup: From an API point of view the signature of the method is consistent with other metric functions and there is more logic needed in the calling code to use this function. If you really want to use `diff` (which is pre-computed) why don't we change the signature of the method to something like below:-

```
def error_rate(y_pred, y_true, diff=None):
```
That way you can use diff when available and compute error rate from first principle when diff is not available.

imatiach-msft: done, I updated to use this new suggested API instead

PR #1753, opened 2022-09-28
Fetching global causal effects and global policy from SDK backend

gaugup:
## Description

Currently the causal functionality shows causal effects and policy for the entire data. Selecting a cohort doesn't have an impact of the causal tab. This change adds the functionality to fetch the global causal effects and policy from the SDK backend. 

Screenshot of global causal effects requested from SDK backend.
image

Screenshot of global causal policy requested from SDK backend.
image

## Checklist


libs/core-ui/src/lib/Context/ModelAssessmentContext.tsx
```diff
@@ -59,6 +59,18 @@ export interface IModelAssessmentContext {
     request: any,
     abortSignal: AbortSignal
   ) => Promise<IHighchartBoxData>;
+  requestGlobalCausalEffects?: (
```
romanlutz, line 62: We're accumulating rather many requestXYZ functions here. Is there perhaps a way we might bundle them together?
We might otherwise end up with dozens of these as input params. 

@xuke444 any ideas? Off the top of my head, I guess we could define a class that has all these as params, but that is also not particularly pretty.

romanlutz: Chatted with @xuke444 briefly on this and he agrees there should be some structure. Perhaps a tree (like a dict of dicts?). Wdyt @gaugup ?

gaugup: Sure, I am happy to introduce such a structure. Let's brainstorm how to put a structure. Nevertheless, the aforementioned structure is not in scope of this PR.

libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAggregateView/CausalAggregateView.tsx
```diff
@@ -30,7 +30,7 @@ export class CausalAggregateView extends React.PureComponent<ICausalAggregateVie
 
   public render(): React.ReactNode {
     const styles = CausalAggregateStyles();
-    this.props.data.global_effects.sort((d1, d2) => d2.point - d1.point);
+    this.props.globalEffects.sort((d1, d2) => d2.point - d1.point);
```
xuke444, line 33: this will do a in-place sort which impact the parent component pass this value in, will it cause any issue?
for intense method like sorting here, ideally not put in render function, move to componentDidMount/Update could be better.

xuke444 (approved):

libs/core-ui/src/lib/Context/ModelAssessmentContext.tsx
gaugup: I have created a work item for this ask internally.

libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAggregateView/CausalAggregateView.tsx
gaugup: The sorting seems to be present from even from before. I have created a work item for this ask internally to re-locate the sorting function.

PR #1765, opened 2022-10-05
Fix dependency plot showing Average of abs value

vinuthakaranth:

## Description
we initialize with 'Average of absolute value' and corresponding data is shown in plot but just the dropdown doesn't show the above option.
One of the way to fix is to show 'Average of absolute value' as one of the drop down option as below along with other options.
Reviewed by Mehrnoosh.

depPlotAvgAbs

## Checklist


gaugup:
Is there an E2E test for this which can validate this scenario?

tongyu-microsoft (approved):

vinuthakaranth:
> Is there an E2E test for this which can validate this scenario?
I am planning to add it in coming PR. Just wanted to merge this for GA.

PR #1766, opened 2022-10-06
fix error analysis cohort filter for string label values

imatiach-msft:
## Description

Fix the cohort filter to support string label values.  Otherwise, when creating a filter based on "True Y" or "Predicted Y", and changing the base cohort, error analysis can fail to filter the data correctly because the filter from UI is in terms of indexes while the data itself has string values.

## Checklist

### DIFF
--- a/erroranalysis/tests/test_cohort_filter.py
+++ b/erroranalysis/tests/test_cohort_filter.py
@@ -64,13 +64,22 @@ class TestCohortFilter(object):
                            model_task,
                            filters=filters)
 
-    def test_cohort_filter_true_y(self):
-        X_train, X_test, y_train, y_test, feature_names = create_iris_pandas()
+    @pytest.mark.parametrize('use_str_labels', [True, False])
+    def test_cohort_filter_true_y(self, use_str_labels):
+        X_train, X_test, y_train, y_test, feature_names = create_iris_pandas(
+            use_str_labels)
+        classes = None
+        if use_str_labels:
+            classes = create_iris_data()[5]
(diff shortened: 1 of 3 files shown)


erroranalysis/tests/test_cohort_filter.py
```diff
@@ -355,6 +376,19 @@ def create_iris_pandas():
     return X_train, X_test, y_train, y_test, feature_names
 
 
+def create_iris_str_y():
```
gaugup, line 379: could you reuse the create_iris_pandas? Seems like duplicate code here. Maybe pass in a boolean flag in that function to create string labels.

erroranalysis/tests/test_cohort_filter.py
```diff
@@ -83,6 +83,27 @@ def test_cohort_filter_true_y(self):
                            model_task,
                            filters=filters)
 
+    def test_cohort_filter_string_encoded_true_y(self):
```
gaugup, line 86: This test looks very identical to `test_cohort_filter_true_y`. Could you reuse the test `test_cohort_filter_true_y` and parametrize it?

erroranalysis/tests/test_cohort_filter.py
imatiach-msft: done!

erroranalysis/tests/test_cohort_filter.py
imatiach-msft: done, parameterized it

gaugup (approved):

PR #1769, opened 2022-10-10
Implement Confusion Matrix Logic

t-wangjohn:
Implements Confusion Matrix logic, inspired by sklearn confusion matrix implementation.

### DIFF
--- a/libs/core-ui/src/lib/Interfaces/IConfusionMatrixData.ts
+++ b/libs/core-ui/src/lib/Interfaces/IConfusionMatrixData.ts
@@ -3,6 +3,5 @@
 
 export interface IConfusionMatrixData {
   confusionMatrix: number[][];
-  labels: string[];
   selectedLabels: string[];
 }
--- a/libs/core-ui/src/lib/Interfaces/IConfusionMatrixData.ts
+++ b/libs/core-ui/src/lib/Interfaces/IConfusionMatrixData.ts
@@ -1,0 +1,8 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export interface IConfusionMatrixData {
+  confusionMatrix: number[][];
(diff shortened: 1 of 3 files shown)


gaugup:
please change the target postga

libs/core-ui/src/lib/util/calculateConfusionMatrixData.ts
```diff
@@ -0,0 +1,52 @@
...
+  let idx = 0;
+
+  allLabels.forEach((element, index) => {
+    if (selectedLabelSet.has(element)) {
+      labelMap.set(index, idx);
+      idx += 1;
+    }
+  });
+
+  const confusionMatrixSize = labelMap.size;
```
gaugup, line 34: Looks like you just need to read labelMap.size down in the logic.

libs/core-ui/src/lib/util/calculateConfusionMatrixData.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  const selectedLabelSet = new Set<string>(
+    selectedLabels !== undefined ? selectedLabels : allLabels
+  );
+
+  const labelMap = new Map<number, number>();
+
+  let idx = 0;
+
```
gaugup, line 26: Put in a comment why you need to relabel.

libs/core-ui/src/lib/Interfaces/IConfusionMatrixData.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export interface IConfusionMatrixData {
+  confusionMatrix: number[][];
+  labels: string[];
```
gaugup, line 6: Why do you need `labels`?

PR #1770, opened 2022-10-13
Change description of cohort selection panel in Aggregate Feature Importance

gaugup:
## Description

Before:-

After:-
image

## Checklist

### DIFF
--- a/libs/localization/src/lib/en.json
+++ b/libs/localization/src/lib/en.json
@@ -1070,6 +1070,7 @@
       "infoTitle": "Additional information on aggregate feature importance",
       "legendHelpText": "Toggle cohorts on and off in the plot by clicking on the legend items.",
       "missingParameters": "This tab requires the local feature importance parameter be supplied.",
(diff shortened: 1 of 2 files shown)


libs/localization/src/lib/en.json
```diff
@@ -1070,6 +1070,7 @@
       "infoTitle": "Additional information on aggregate feature importance",
       "legendHelpText": "Toggle cohorts on and off in the plot by clicking on the legend items.",
       "missingParameters": "This tab requires the local feature importance parameter be supplied.",
+      "sortByCohort": "Sort by cohort",
       "sortBy": "Sort by datapoint",
```
tongyu-microsoft, line 1074: is sortBy used anywhere else? If not, can we remove it

tongyu-microsoft (approved):

libs/localization/src/lib/en.json
gaugup: Yes, it is used in local importance chart.

PR #1783, opened 2022-10-20
[Highcharts] Enable csv download, xls download and view data table in highcharts

tongyu-microsoft:
This PR enable csv & xls download in hightcharts. This is requested from an OCV where user has the comment: "Would be really cool and helpful to add an option to download as data chart. The PDF and image options are nice, but I can't work with the data."

## Description
We add three more options in highcharts, so users can download csv/xls file and interact with data:
image

Download csv file:
image

Click on 'View data table' button:
image

localization:
image

## Checklist


gaugup:
Could you localize the new download options by looking at the PR https://github.com/microsoft/responsible-ai-toolbox/pull/1691?

gaugup (approved):

PR #1784, opened 2022-10-21
fix isnan exception in erroranalysis for newer versions of pandas and numpy when calling qcut

imatiach-msft:
## Description

Fix isnan exception in erroranalysis for newer versions of pandas and numpy when calling qcut.

When going to heatmap, in some environments with python 3.8 installed and latest versions of pandas and numpy, erroranalysis matrix_filter method would fail with error:

\\responsibleai\\responsible-ai-widgets\\erroranalysis\\erroranalysis\\_internal\\matrix_filter.py\", line 365, in bin_data\n    bindf = pd.qcut(feat_col, bins, precision=PRECISION, duplicates=DROP)\n  \\.conda\\envs\\big_data\\lib\\site-packages\\pandas\\core\\reshape\\tile.py\", line 375, in qcut\n    x_np = x_np[~np.isnan(x_np)]\nTypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''\n"

The solution was to change to more specific dtypes using convert_dtypes() for pandas dataframe after making a copy, since the copy was changing all types to object dtype.

Since qcut was failing on object dtype column when creating the dataframe copy, changing to more specify number type via convert_dtypes() fixed the error.

## Checklist

### DIFF
--- a/erroranalysis/erroranalysis/_internal/matrix_filter.py
+++ b/erroranalysis/erroranalysis/_internal/matrix_filter.py
@@ -21,6 +21,7 @@ from erroranalysis._internal.metrics import (get_ordered_classes,
 BIN_THRESHOLD = MatrixParams.BIN_THRESHOLD
 CATEGORY1 = 'category1'
 CATEGORY2 = 'category2'
+CONVERT_DTYPES = 'convert_dtypes'
 COUNT = 'count'
 FALSE_COUNT = 'falseCount'
 INTERVAL_MIN = 'intervalMin'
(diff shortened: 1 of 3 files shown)


gaugup:
can you describe what the fix does?

imatiach-msft:
"can you describe what the fix does?"
I added another sentence at the end to help clarify.

vinuthakaranth (approved):

tongyu-microsoft (approved):

PR #1792, opened 2022-10-27
Update e2e visit() function to not refresh the page if provided name doesn't exist in RAINotebookNames

tongyu-microsoft:
This PR updates e2e visit() function to not refresh the page if provided name doesn't exist in RAINotebookNames.

## Description
We don't want to refresh the page in AzureML to test for interactive dashboard features.

## Checklist


gaugup:
Should this change go into postga?

tongyu-microsoft:
> Should this change go into postga?

I need to check in this to main so AML can consume the new Rai version with this change. Since this change is only for E2E, I think it’s safe to check in to main

PR #1801, opened 2022-11-04
Fix Model overview - Dataset cohorts]: At 320*256 viewport content is getting truncated and graph is not visible.

vinuthakaranth:
Fix Model overview - Dataset cohorts]: At 320*256 viewport content is getting truncated and graph is not visible.

## Description
graphTrunc2

## Checklist

### DIFF
--- a/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.styles.ts
+++ b/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.styles.ts
@@ -60,7 +60,10 @@ export const modelOverviewStyles: () => IProcessedStyleSet<IModelOverviewStyles>
         selectors: {
           "[role='tablist'].ms-Pivot": {
             display: "flex",
-            flexWrap: "wrap"
+            flexWrap: "wrap",
+            overflow: "hidden",
+            whiteSpace: "nowrap",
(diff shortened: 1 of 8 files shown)


libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.styles.ts
```diff
@@ -60,7 +60,10 @@ export const modelOverviewStyles: () => IProcessedStyleSet<IModelOverviewStyles>
         selectors: {
           "[role='tablist'].ms-Pivot": {
             display: "flex",
-            flexWrap: "wrap"
+            flexWrap: "wrap",
```
imatiach-msft, line 63: I think you could refactor this into a common place and then use mergeStyles in here and other .styles.ts files to remove some duplicate lines, but maybe it's not worth the effort

imatiach-msft (approved):

tongyu-microsoft (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.styles.ts
vinuthakaranth: Thanks Ilya, great comment! I will check this in coming PR.

PR #1802, opened 2022-11-04
[LogarithmicScaling] Enable log scaling for data explorer, feature importance, counterfactual and casual

tongyu-microsoft:
This PR enables log scaling for data explorer, feature importance, counterfactual and casual. Feature request from customer: "Many features that I would like to investigate have values that span several orders of magnitude (especially when including outliers). Plotting these features on a logarithmic scale (base 10) would allow for better analysis. A toggle to switch between linear and logarithmic scales would be ideal." 
Attached before & after screenshots for each component.

## Description
**data explorer:**
Before:
image

After:
image
image

**feature importance:**
Before:
image

After:
image

**counterfactual:**
Before:
image

After:
image

**casual:**
Before:
image

After:
image

## Checklist


libs/core-ui/src/lib/components/AxisConfigBinOptions.tsx
```diff
@@ -26,68 +28,120 @@ export interface IAxisConfigBinOptionsProps {
...
+                  max={this.props.maxHistCols}
+                  min={this.props.minHistCols}
+                  setNumericValue={this.setNumericValue}
+                  value={this.props.selectedBinCount.toString()}
+                />
+              )}
+            {!(this.props.mustBin || this.props.selectedColumn.options.bin) &&
+              this.props.canDither &&
+              allowUserInteract(this.props.selectedColumn.property) && (
+                <Toggle
```
imatiach-msft, line 116: this Toggle looks the same, with same parameters, as the one in line 83:
https://github.com/microsoft/responsible-ai-toolbox/pull/1802/files#diff-635f8fa7b2f26365844135a68ebdb4ae8c323032fb332b0bf348c44d2154bbc8R83
consider refactoring both to a common method?

imatiach-msft (approved):

libs/core-ui/src/lib/components/AxisConfigBinOptions.tsx
tongyu-microsoft: Thanks for the comment! Updated

libs/core-ui/src/lib/util/getDependencyChartOptions.ts
```diff
@@ -57,7 +60,8 @@ export function getDependencyChartOptions(
       }
     ],
     xAxis: {
-      categories: xLabels
+      categories: xLabels,
+      type: logarithmicScaling ? AxisTypes.Logarithmic : undefined
```
vinuthakaranth, line 64: just curious, there are two types of axis. If it isn't logarithmic it would be Liner. Why undefined?

vinuthakaranth (approved):

libs/core-ui/src/lib/util/getDependencyChartOptions.ts
tongyu-microsoft: The reason to set it to undefined is, from Highcharts wiki: `The type of axis. Can be one of linear, logarithmic, datetime, category or treegrid. Defaults to treegrid for Gantt charts, linear for other chart types.` In case we set `treegrid` for `Gantt charts`, I would just keep it to `undefined`, so it can select by default value according to chart type.

libs/core-ui/src/lib/components/AxisConfigBinOptions.styles.ts
```diff
@@ -0,0 +1,19 @@
...
+import { IProcessedStyleSet, IStyle, mergeStyleSets } from "@fluentui/react";
+
+export interface IAxisConfigBinOptionStyles {
+  hideMinMax: IStyle;
+  minMax: IStyle;
+}
+
+export const axisConfigBinOptionsStyles: () => IProcessedStyleSet<IAxisConfigBinOptionStyles> =
+  () => {
+    return mergeStyleSets<IAxisConfigBinOptionStyles>({
```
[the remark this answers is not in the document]
tongyu-microsoft: Looks good!

In feature importance: 
image

In axis config panel:
image

libs/interpret/src/lib/MLIDashboard/Controls/GlobalExplanationTab/GlobalExplanationTab.tsx
```diff
@@ -70,6 +72,7 @@ interface IGlobalExplanationTabState {
   globalBarSettings?: IGlobalBarSettings;
```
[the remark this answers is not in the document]
tongyu-microsoft: Correction: this file (GlobalExplanationTab) is for the chart under Aggregate feature importance.

The reason this GlobalExplanationTab is different from others is, we cannot select x-axis value from axis config dialog (see the screenshot below, x-axis is not clickable). We need to select the x-axis value from right side dropdown list. So this couldn't aplly the common axisConfig change. And we need to have a toggle separately. This design is from Minsoo.
image

PR #1803, opened 2022-11-04
Calculate RAI insights for first `maximum_rows_for_test` samples

gaugup:
## Description

This is behavior change where instead of failing `RAIInsights` execution if the number of samples is greater than `maximum_rows_for_test`. This PR changes this behavior to not fail the execution but compute RAI for the first `maximum_rows_for_test` samples of the test dataset. 

This PR also adds two new flags in `Dataset` interface `is_large_data_scenario` and `use_entire_test_data`. The flag `use_entire_test_data` will be used in case the entire dataset needs to be used for RAI computation for the calls routed from dashboard to python backend. The flag `is_large_data_scenario` is for keeping track if the computation for first `maximum_rows_for_test` samples and the number of samples in test data was greater than configured limit of `maximum_rows_for_test`.

## Checklist


responsibleai/responsibleai/_interfaces.py
```diff
@@ -20,6 +20,8 @@ class Dataset:
     class_names: List[str]
     categorical_features: List[str]
     target_column: str
+    is_large_data_scenario: bool
```
vinuthakaranth, line 23: should this be optional? asking because if it will break existing flow?

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -104,6 +104,32 @@ def __init__(self, model: Optional[Any], train: pd.DataFrame,
         :type feature_metadata: FeatureMetadata
         """
         categorical_features = categorical_features or []
+        if len(test) > maximum_rows_for_test:
```
vinuthakaranth, line 107: Should this change be merged to postGA branch?

vinuthakaranth (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
gaugup: Thanks. I updated the target branch.

responsibleai/responsibleai/_interfaces.py
gaugup: going forward SDK will always add this so it might be ok if it is not optional. It shouldn't break any existing flow as this is not being used anywhere. 

The UI treats this as Optional property anyways in the PR:- https://github.com/microsoft/responsible-ai-toolbox/pull/1804.

PR #1806, opened 2022-11-07
Fix ‘Close button’ and ‘Spin button’ present in new cohort dialog does not have any title or aria-label

vinuthakaranth:

Fix ‘Close button’ and ‘Spin button’ present in new cohort dialog does not have any title or aria-label by adding aria label for increasing and decreasing value.

image

## Checklist

### DIFF
--- a/libs/localization/src/lib/en.json
+++ b/libs/localization/src/lib/en.json
@@ -5,7 +5,9 @@
     "identityFeature": "Identity feature",
     "infoTitle": "Additional information",
     "spinButton": "Spin",
-    "editButton": "Edit"
+    "editButton": "Edit",
+    "decreaseValue": "Decrease value",
+    "increaseValue": "Increase value"
   },
   "ChartContextMenu": {
     "viewInFullScreen": "View in full screen",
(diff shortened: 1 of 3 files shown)


tongyu-microsoft (approved):

libs/localization/src/lib/en.json
```diff
@@ -5,7 +5,9 @@
     "identityFeature": "Identity feature",
     "infoTitle": "Additional information",
     "spinButton": "Spin",
-    "editButton": "Edit"
+    "editButton": "Edit",
+    "decreaseValue": "Decrease value",
```
hawestra, line 9: not sure if where the original strings came from, but should we keep the "by one" clause ?

hawestra (approved):

imatiach-msft (approved):

libs/localization/src/lib/en.json
vinuthakaranth: Few of the increase and decrease is based on value in data. So for them we can't use 'by 1'. I updated to keep by one for [FeatureConfigurationFlyout.tsx](https://github.com/microsoft/responsible-ai-toolbox/pull/1806/files#diff-fd9c3511dddd161b6dec541d5d8f502602609085885647f7f84f613bc3e46e65) which had it previously.

PR #1809, opened 2022-11-09
update to latest 16.x node version for tests

imatiach-msft:
## Description

We previously capped node version to 16.8 due to build failures.  This PR tries to update to latest again, as it was before.

I also tried to add 18.x as well, but I saw build failures so I took it out (related to rollup plugin which I think just needs to be upgraded, but when I upgraded I saw some other failures).  Now that 18.x has an LTS version released we should at some point look into adding it as well.

## Checklist


gaugup (commented):
Should this go into postga as main is locked right now?

imatiach-msft:
@gaugup I think we can just wait until we've lifted our internal code freeze

gaugup (approved):

romanlutz (approved):

PR #1811, opened 2022-11-10
add multilabel support to RAI dashboard and multilabel text classification covid events dataset

imatiach-msft:
## Description

add multilabel support to RAI dashboard and multilabel text classification covid events dataset

Screenshot of data analysis view with multiple labels:

image

## Checklist


tongyu-microsoft:
Do we identify the prediction as incorrect when only some of the predictY are correct - when predictY is not exactly same as trueY?

tongyu-microsoft (approved):

imatiach-msft:
"Do we identify the prediction as incorrect when only some of the predictY are correct - when predictY is not exactly same as trueY?"
great question!  I think we should bring this up in the sync.  Currently I do, but maybe there are finer controls we can add to look at specific labels.

PR #1812, opened 2022-11-10
Add RAI version number to oss serialize result

tongyu-microsoft:
This PR adds RAI version number to oss serialize result.

## Description
We will generate raiversion.json file:
image
In the json file, it lists out the responsibleai package version number:
image

## Checklist

### DIFF
--- a/responsibleai/responsibleai/rai_insights/rai_base_insights.py
+++ b/responsibleai/responsibleai/rai_insights/rai_base_insights.py
@@ -17,7 +17,7 @@ from responsibleai._internal.constants import Metadata
 
 _DATA = 'data'
 _DTYPES = 'dtypes'
-_RAIVERSION = "raiversion"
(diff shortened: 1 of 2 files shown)


gaugup (commented):
If the intention is to add this for azureml command jobs, then the `responsibleai` version could be found in the environment attached to the command job. If this is indeed the case, then do we need this change? Do we require the version to be saved in the file for some feature?

xuke444:
> 

Environment sometime may not be reliable especially if we use floating version.
This one will also help on oss as well, in future when we debug or re-loading pre-generated rai insight it could also help.

gaugup:
> > 
> 
> Environment sometime may not be reliable especially if we use floating version. This one will also help on oss as well, in future when we debug or re-loading pre-generated rai insight it could also help.

Sounds good.

gaugup (approved):

PR #1829, opened 2022-12-03
Add dataset analysis chart view for large data

gaugup:
## Description

The PR adds the dataset analysis for large data by requesting bar/box plots from SDK backend. 

## Checklist


apps/widget/src/app/ModelAssessment.tsx
```diff
@@ -32,6 +32,8 @@ export class ModelAssessment extends React.Component<IModelAssessmentProps> {
       | "requestImportances"
       | "requestCausalWhatIf"
       | "requestBoxPlotDistribution"
+      | "requestDatasetAnalysisBarChart"
```
romanlutz, line 35: This is becoming a rather massive list. Should we bundle these together somehow? Otherwise we'll end up with a hundred input params. Thoughts?

libs/dataset-explorer/.eslintrc.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": ["../../.eslintrc"],
-  "ignorePatterns": ["!**/*"],
+  "ignorePatterns": ["!**/*", "LargeDatasetExplorerTab.tsx"],
```
xuke444, line 3: why do we exclude this from from linting?
Are we able to fix lint issues in the file instead of excluding checking

gaugup: The only lint problem is the number of lines is greater than 300.

gaugup: Just limiting the linting to file size now. There are no other lint issues.

xuke444 (approved):

apps/widget/src/app/ModelAssessment.tsx
gaugup: Created.

PR #1834, opened 2022-12-07
Add missing property 'id' in ICounterfactualData

gaugup:
## Description

It seems SDK is adding this property but UI doesn't have this listed in `ICounterfactualData`. 

image

## Checklist


apps/dashboard/src/app/__snapshots__/App.test.tsx.snap
```diff
@@ -12176,6 +12176,7 @@ exports[`FairnessWizardV1 should render successfully 1`] = `
               "native-country",
               "income",
             ],
+            "id": "abc",
```
romanlutz, line 12179: In a real scenario, will this still be a string like "abc" or rather a number or something entirely different?

gaugup: it will be a guid provided by SDK. Please check description.

PR #1835, opened 2022-12-07
Send the correct column name to SDK backend for model overview distribution charts

gaugup:
## Description

It seems we were sending a number as a class which was causing an exception. Now sending the correct axis via "text" field. This PR also enables the box plot for regression.

classification bug fix.
image

regression support.
image

## Checklist


libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionBoxChart.tsx
```diff
@@ -46,7 +46,7 @@ export class ProbabilityDistributionBoxChart extends React.Component<IProbabilit
             cohort,
             index,
             this.props.probabilityOption?.key || "",
-            this.props.probabilityOption?.id,
+            this.props.probabilityOption?.text,
```
xuke444, line 49: We should try avoid using text as input paramter, since it is user facing values, subject to change at anytime.
Is it possible to use some sort of map to get string values based on the key?
If key is not used anywhere, we can also change the key to be string.

romanlutz: Agreed. See my comment as well.

gaugup: The SDK side has little idea as to what id value is. The text value is a column value that SDK backend for large data recognizes. There are a lot of UI jargons that we translate for SDK like feature names for cohorts etc. This is yet another example.

I agree that we should have clearer contracts with SDK backend.

gaugup:
> I am somewhat confused by the description. How did it work so far? You write that it "enables the box plot for regression" which worked just fine as of a few days ago. I seem to be missing something.
> 
> Just taking a quick glance at the code I thought that sending the id of the option is entirely intended as it's generally preferable to use IDs than user-provided text for identifying anything.

By working for regression, I meant that it works for regression by getting the distribution from SDK backend.

For ID, please see my response on Ke's comment.

xuke444 (approved):

PR #1845, opened 2022-12-12
Disable local causal chart for large data support

gaugup:
## Description

The local causal effects will be funneled through the bubble/scatter chart experience. Disabling the local causal effects for now.

Before:-
image

After:-
image

## Checklist


romanlutz (approved):
Would be nice if we had a test case that made sure these parts only show iff they're supposed to.

imatiach-msft (approved):

gaugup:
> Would be nice if we had a test case that made sure these parts only show iff they're supposed to.

I can add a mock dataset that shows only the parts of the dashboard which should be enabled for large data. Would that work? It is harder to add an E2E test case for this scenario.

PR #1852, opened 2022-12-14
update rai-core-flask by closing WSGIServer on separate thread to fix segfault errors and get the rai-core-flask release builds working again

imatiach-msft:
## Description

Update: instead of pinning, I found that closing WSGIServer on a separate thread seems to fix the segfaults, and doing the pinning is not necessary.  See related stackoverflow about graceful shutdown of WSGIServer:

https://stackoverflow.com/questions/18277048/gevent-pywsgi-graceful-shutdown

=====

rai-core-flask package release seems to be failing on errors in tests from some segfaults: https://github.com/microsoft/responsible-ai-toolbox/actions/runs/3699568450

It seems the latest versions of greenlet v2.0 and gevent are less stable and fail with segfaults.  Pinning here to <2.0 and corresponding gevent version until newer packages are released that are more stable.

## Checklist

### DIFF
--- a/rai_core_flask/rai_core_flask/flask_helper.py
+++ b/rai_core_flask/rai_core_flask/flask_helper.py
@@ -62,6 +62,12 @@ class FlaskHelper(object):
         self._thread = threading.Thread(target=self.run, daemon=True)
         self._thread.start()
 
+        # Closes server on program exit, including freeing all sockets
+        def closeserver():
+            self.stop()
+
+        atexit.register(closeserver)
+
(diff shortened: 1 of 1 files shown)


imatiach-msft:
this pin might not be needed, just the other fix:
```
greenlet>=1.1.2,<2.0.0 # wait until 2.0 is more stable
gevent>=21.12.0,<22.10.2
```

gaugup (approved):

PR #1858, opened 2022-12-16
Remove `setState` call from render function in `ProbabilityDistributionChart.tsx`

gaugup:
## Description

Looks like we shouldn't explicitly `setState` in `render` function as per https://blog.logrocket.com/using-setstate-react-components/#:~:text=The%20render()%20function%20should,avoid%20using%20setState()%20here. So removing the call setState(). Looks like `componentDidMount` takes care of setting the state correctly anyways.

This helps in resolving the following warning trace seen in browser dev tools:-
image

## Checklist


xuke444 (approved):

romanlutz (approved):
This should be fine. I remember putting this there because we don't have the probability options on first render pre-setup of the context, joint dataset etc. This should still work as it rerenders as soon as the context is there. Perhaps it makes sense to combine with the other if right before, or maybe that's obsolete now? I think it was just precondition for this if (since you need the probability options for that setState call that's removed in this PR).

gaugup:
> This should be fine. I remember putting this there because we don't have the probability options on first render pre-setup of the context, joint dataset etc. This should still work as it rerenders as soon as the context is there. Perhaps it makes sense to combine with the other if right before, or maybe that's obsolete now? I think it was just precondition for this if (since you need the probability options for that setState call that's removed in this PR).

The render should be called again if the `componentDidMount` causes a state change which it should. That should set the state `probabilityOption` properly.

Ideally 'this.getProbabilityOptions()' should be cached in the state in the constructor and reused. Doesn't look like it will change in the lifecycle of this component. This should simplify the code further. But as you said that it depends on jointdataset and the context to be correctly setup.

PR #1868, opened 2022-12-28
Add better error message if causal treatment features contain dropped features

tongyu-microsoft:
This PR adds better error message if causal treatment features contain dropped features:
(responsibleai.exceptions.UserConfigValidationException: '{dropped_feature_name}' in treatment_features has been dropped during training the model)

## Description

We added dropped_features where users can choose to drop certain features from the model. In the case when user adds causal component, while treatment features contain dropped features, we need to have a better error message saying "'{dropped_feature_name}' in treatment_features has been dropped during training the model"

## Checklist

### DIFF
--- a/responsibleai/responsibleai/managers/causal_manager.py
+++ b/responsibleai/responsibleai/managers/causal_manager.py
@@ -155,9 +155,10 @@ class CausalManager(BaseManager):
         :type random_state: int or RandomState or None
         """
         for feature in treatment_features:
-            if feature in set(self._dropped_features):
+            if self._dropped_features and \
+                    feature in set(self._dropped_features):
                 message = ("'{}' in treatment_features has been dropped "
-                            f"during training the model").format(feature)
+                           "during training the model").format(feature)
                 raise UserConfigValidationException(message)
         difference_set = set(treatment_features) - set(self._train.columns)
         if len(difference_set) > 0:
--- a/responsibleai/responsibleai/managers/causal_manager.py
+++ b/responsibleai/responsibleai/managers/causal_manager.py
@@ -50,9 +50,10 @@ class CausalManager(BaseManager):
         :type task_type: str
         :param categorical_features: All categorical feature names.
         :type categorical_features: list
-        :param dropped_features: List of features that were dropped by the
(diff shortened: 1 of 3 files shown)


responsibleai/responsibleai/managers/causal_manager.py
```diff
@@ -57,6 +61,9 @@ def __init__(
         self._categorical_features = categorical_features
         if categorical_features is None:
             self._categorical_features = []
+        self._dropped_features = dropped_features
```
romanlutz, line 64: Super nit: if you want to shorten this you can write 

dropped_features or []

romanlutz (approved):

responsibleai/responsibleai/managers/causal_manager.py
tongyu-microsoft: Updated, thanks!

responsibleai/responsibleai/managers/causal_manager.py
```diff
@@ -494,5 +505,9 @@ def _load(path, rai_insights):
         inst.__dict__['_task_type'] = rai_insights.task_type
         inst.__dict__['_categorical_features'] = \
             rai_insights.categorical_features
+        inst.__dict__['_dropped_features'] = None
+        if rai_insights._feature_metadata:
+            inst.__dict__['_dropped_features'] = \
```
gaugup, line 510: I don't think you need to add dropped_features to the save/load path as it is just used for validation and not in actual computation of causal effects. Could you and possibly remove the above code.

tongyu-microsoft: If we don't save/load dropped_features, there is a error where CausalManager has no attribute named dropped_features

gaugup: We should just change this to be feature_metadata instead of just passing the dropped_features inside feature_metadata.

tongyu-microsoft: Good point, Updated, thanks!

gaugup (approved):

PR #1871, opened 2022-12-30
Remove warning for 'dropped_features are not in use currently.'

tongyu-microsoft:
This PR removes warning for 'dropped_features are not in use currently.', since we already support dropped_features.

## Description
PR to remove a warning 'dropped_features are not in use currently.'

## Checklist

### DIFF
--- a/responsibleai/responsibleai/feature_metadata.py
+++ b/responsibleai/responsibleai/feature_metadata.py
@@ -36,8 +36,6 @@ class FeatureMetadata:
             warnings.warn('datetime_features are not in use currently.')
         if self.categorical_features is not None:
             warnings.warn('categorical_features are not in use currently.')
-        if self.dropped_features is not None:
-            warnings.warn('dropped_features are not in use currently.')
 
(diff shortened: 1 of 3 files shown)


responsibleai/responsibleai/feature_metadata.py
```diff
@@ -36,8 +36,6 @@ def __init__(self, identity_feature_name: Optional[str] = None,
             warnings.warn('datetime_features are not in use currently.')
         if self.categorical_features is not None:
```
romanlutz, line 37: This kind of begs the question of why we accept them in the first place. Could be commented out completely, or only added whenever they are indeed used.
I guess that's not an issue with this PR at all, just the larger direction. Let's have this conversation next week when people are back.

romanlutz (approved):

gaugup (approved):

responsibleai/responsibleai/feature_metadata.py
tongyu-microsoft: Sounds good!

PR #1874, opened 2023-01-04
Add forecasting dashboard - UI - transformation creation, table, and comparison

romanlutz:
## Description

This PR adds the UI code for the forecasting dashboard. This means it is visible in the test environment but no changes to Python code are included here.

Documentation (mostly README, notebooks) and e2e tests will be added later this month and in early February before this is made available to any users.

image

image

image

image

image

## Checklist


libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationCreationDialog.tsx
```diff
@@ -0,0 +1,242 @@
...
+                .invalidCombinationErrorMessage;
+          }
+        });
+      }
+    }
+    return (
+      <Dialog
+        hidden={!this.props.isVisible}
+        dialogContentProps={{
+          closeButtonAriaLabel: "Close",
```
tongyu-microsoft, line 128: localization for aria label?

libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationsTable.tsx
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  public render(): React.ReactNode {
+    if (
+      this.props.transformations.size === 0 ||
+      this.state === undefined ||
+      this.state.rows === undefined ||
+      this.state.rows.length === 0
+    ) {
+      return;
```
tongyu-microsoft, line 75: Do we have any placeholder in the case there is no row in the table?

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
```diff
@@ -25,6 +37,22 @@ export class ForecastingDashboard extends React.Component<
...
+    };
+  }
+
+  public componentDidMount(): void {
+    const transformations = new Map<number, Map<string, Transformation>>();
+    transformations.set(
+      this.context.baseErrorCohort.cohort.getCohortID(),
+      new Map<string, Transformation>()
+    );
+  }
```
tongyu-microsoft, line 54: is this componentDidMount in use? I think `transformations ` here is not used?

tongyu-microsoft (approved):

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
romanlutz: I had wanted to capture the initial state and put an empty map in place for the first cohort, but it turns out that's not needed given how I wrote the `render` method. Nice catch! I'll remove the entire `componentDidMount`

libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationsTable.tsx
romanlutz: Great question! No, because we just don't show the table component at all in that case.

libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationCreationDialog.tsx
romanlutz: Oh, I had not heard of localization for aria labels. After a quick search I found examples for both localization and lack thereof. I will not address it in this PR, but I'll tackle all aria labels in a separate PR. I hope that's fine by you.

PR #1875, opened 2023-01-04
Improve spacing in RAI dashboard

romanlutz:
## Description

This is hopefully a visual improvement. I've added some additional padding in some of our components since several of them were really close to the boundary at the bottom. Before/after screenshots below:

| component | Before | After |
| --- | --- | --- |
| model overview | image | image |
| data explorer | image | image |
| counterfactual | image | image |

## Checklist


romanlutz:
Changed to 32px after discussing with Owen who approves of the change. Screenshots above not updated to reflect 32px, but it looks almost the same (just with 8px less, so slightly less padding)

gaugup (approved):

gaugup:
Do other components like EA and causal also need this change?

libs/dataset-explorer/src/lib/ChartView/utils/DatasetExplorerTab.styles.ts
```diff
@@ -153,7 +153,7 @@ export const datasetExplorerTabStyles: () => IProcessedStyleSet<IDatasetExplorer
       page: {
         color: theme.semanticColors.bodyText,
         height: "100%",
-        padding: "0 40px 10px 40px",
+        padding: "0 40px 32px 40px",
```
gaugup, line 156: Shall we move these to some constant strings?

romanlutz: It's not uniform everywhere. I applied the same for causality and it didn't make sense there because the component is comprised of different pieces that have their own spacing. I don't want to overthink this one and keep it as a simple fix. If we get time to rethink how these are structured we can do something more holistic and at that time, I agree, this should be done with constants and uniformly.

romanlutz: Other than causality, data balance and data explorer (tabular view, but not chart view) were also examples where it didn't make sense to increase.

hawestra (approved):

PR #1883, opened 2023-01-05
Populate probability box chart data on mount

romanlutz:
## Description

This fixes a bug where the probability box chart doesn't show up when switching from the dataset cohort view to the feature based cohort view. This bug does not occur for the regression box chart so no adjustments are required.

## Checklist


gaugup (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionBoxChart.tsx
```diff
@@ -30,39 +30,21 @@ export class ProbabilityDistributionBoxChart extends React.Component<IProbabilit
...
-        this.props.boxPlotState,
-        this.props.onBoxPlotStateUpdate
-      );
-    }
+    this.calculateBoxDataIfNeeded(prevProps);
   }
 
   public render(): React.ReactNode {
     const theme = getTheme();
+    if (this.props.boxPlotState.boxPlotData.length === 0) {
```
hawestra, line 45: nit: if(!this.props.boxPlotState.boxPlotData.length) {...}, same for the if statement below

hawestra (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionBoxChart.tsx
romanlutz: Indeed. As long as we don't have the box plot data we can't display anything. The other if is in a separate method, though. The box plot data calculation may end up getting rather expensive with larger datasets, which is why I want to be careful when calling it and avoid unnecessarily calculating it.

romanlutz: Ah you meant the === 0 part! Done!

PR #1884, opened 2023-01-05
Separate out rai_core_flask CI and add paths filter

romanlutz:
## Description

With this change the rai_core_flask CI will only run if we make changes to that folder which is very rare. This should reduce the ridiculous amount of requeuing pipelines significantly.

## Checklist

### DIFF
--- a/.github/workflows/CI-python.yml
+++ b/.github/workflows/CI-python.yml
@@ -9,9 +9,6 @@ on:
 jobs:
   ci-python:
     strategy:
-      # keep running remaining matrix jobs even if one fails
-      # to avoid having to rerun all jobs several times
-      fail-fast: false
       matrix:
         packageDirectory:
           [
--- a/.github/workflows/CI-python.yml
+++ b/.github/workflows/CI-python.yml
@@ -11,12 +11,7 @@ jobs:
     strategy:
       matrix:
         packageDirectory:
-          [
-            "responsibleai",
-            "erroranalysis",
-            "raiutils",
-            "rai_test_utils"
-          ]
+          ["responsibleai", "erroranalysis", "raiutils", "rai_test_utils"]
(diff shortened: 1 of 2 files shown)


.github/workflows/CI-python.yml
```diff
@@ -25,9 +24,6 @@ jobs:
           [ubuntu-20.04, ubuntu-latest, macos-latest, windows-latest]
         pythonVersion: ["3.6", "3.7", "3.8", "3.9"]
         exclude:
-          - packageDirectory: "rai_core_flask"
-            operatingSystem: macos-latest
-            pythonVersion: "3.9"
           - operatingSystem: ubuntu-20.04
```
gaugup, line 27: Can we remove the setting of fail_fast flag as well? I don't think we should run tests that will fail anyways.

gaugup (commented):
Can we remove the setting of fail_fast flag as well? I don't think we should run tests that will fail anyways.

.github/workflows/CI-python.yml
romanlutz: I suppose for this particular one we can turn it off. I would still like to keep it for the notebook e2e tests, though. Those take ages (1h+) and one of them fails semi-frequently. I'll remove it for the python one.

gaugup: great. Thanks.

gaugup (approved):

romanlutz:
This only touches the CI python pipelines which are passing. Everything else looks to be passing, too, but I'd have to rerun them now again since another PR got merged. Since I've spent way too much time on rerunning builds today (and waiting for this to merge so that at least the other 5 PRs can run their builds somewhat faster) I'll make an exception and merge this by bypassing branch protections.

PR #1887, opened 2023-01-06
Forecasting: boilerplate only

romanlutz:
## Description

This takes the boilerplate parts out of #1874 and should make it a lot easier to review!

## Checklist

### DIFF
--- a/apps/dashboard/src/app/App.tsx
+++ b/apps/dashboard/src/app/App.tsx
@@ -13,6 +13,7 @@ import { App as ErrorAnalysis } from "../error-analysis/App";
 import { App as Fairness } from "../fairness/App";
 import { App as InterpretText } from "../interpret-text/App";
 import { App as Interpret } from "../interpret/App";
+import { App as ModelAssessmentForecasting } from "../model-assessment-forecasting/App";
 import { App as ModelAssessmentText } from "../model-assessment-text/App";
 import { App as ModelAssessmentVision } from "../model-assessment-vision/App";
 import { App as ModelAssessment } from "../model-assessment/App";
@@ -162,6 +163,21 @@ export class App extends React.Component<IAppSetting, IAppState> {
           )}
           {this.state.application === "modelAssessmentVision" && (
             <ModelAssessmentVision
+              {...applications[this.state.application].datasets[
+                this.state.dataset
+              ]}
+              theme={themes[this.state.theme]}
+              language={Language[this.state.language]}
+              version={
+                applications[this.state.application].versions[
+                  this.state.version
+                ]
+              }
+              featureFlights={parseFeatureFlights(this.state.featureFlights)}
+            />
+          )}
+          {this.state.application === "modelAssessmentForecasting" && (
+            <ModelAssessmentForecasting
               {...applications[this.state.application].datasets[
                 this.state.dataset
               ]}
--- a/apps/dashboard/src/model-assessment-forecasting/App.tsx
+++ b/apps/dashboard/src/model-assessment-forecasting/App.tsx
@@ -1,0 +1,49 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { ITheme } from "@fluentui/react";
+import { HelpMessageDict } from "@responsible-ai/error-analysis";
+import { Language } from "@responsible-ai/localization";
+import {
+  ModelAssessmentDashboard,
+  IModelAssessmentDashboardProps,
+  IModelAssessmentData
+} from "@responsible-ai/model-assessment";
+import React from "react";
+
+interface IAppProps extends IModelAssessmentData {
+  theme: ITheme;
+  language: Language;
+  version: 1 | 2;
+  classDimension?: 1 | 2 | 3;
+  featureFlights?: string[];
+}
+
+export class App extends React.Component<IAppProps> {
+  private messages: HelpMessageDict = {
+    LocalExpAndTestReq: [{ displayText: "LocalExpAndTestReq", format: "text" }],
+    LocalOrGlobalAndTestReq: [
+      { displayText: "LocalOrGlobalAndTestReq", format: "text" }
+    ],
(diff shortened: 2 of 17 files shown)


libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
```diff
@@ -0,0 +1,19 @@
...
+} from "@responsible-ai/core-ui";
+import React from "react";
+
+export class ForecastingDashboard extends React.Component {
+  public static contextType = ModelAssessmentContext;
+  public context: React.ContextType<typeof ModelAssessmentContext> =
+    defaultModelAssessmentContext;
+
+  public render(): React.ReactNode {
+    return <Text>Placeholder</Text>;
```
gaugup, line 17: Is `Placeholder` a string or a variable?

gaugup (approved):

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
romanlutz: A string. And it's going to vanish in the next PR.

imatiach-msft (approved):

imatiach-msft:
can you mention the command used to generate the files in the PR description?  Just for posterity.

romanlutz:
> can you mention the command used to generate the files in the PR description?  Just for posterity.

I'd love to but this was done by John. I think it's an nx command. @xuke444 set it up do he might know.

vinuthakaranth (approved):

libs/forecasting/.eslintrc.json
```diff
@@ -0,0 +1,18 @@
...
+  "extends": ["plugin:@nrwl/nx/react", "../../.eslintrc.json"],
+  "ignorePatterns": ["!**/*"],
+  "overrides": [
+    {
+      "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
+      "rules": {}
+    },
+    {
+      "files": ["*.ts", "*.tsx"],
+      "rules": {}
```
tongyu-microsoft, line 11: can we add `"max-lines": "error"` in the rules for ts, tsx files
In the refactor work that I was working on, we are trying to limit file lines, so we reduce number of very large files

tongyu-microsoft (approved):

libs/forecasting/.eslintrc.json
tongyu-microsoft: this could be in next PR

romanlutz: For new projects we should absolutely do that! It was actually here because it complained at some point when I went past 300 lines 😆 I subsequently broke up my files into smaller ones. Not sure where that rule went, let me check!

apps/dashboard/src/model-assessment-forecasting/App.tsx
```diff
@@ -0,0 +1,49 @@
...
+    LocalOrGlobalAndTestReq: [
+      { displayText: "LocalOrGlobalAndTestReq", format: "text" }
+    ],
+    PredictorReq: [{ displayText: "PredictorReq", format: "text" }],
+    TestReq: [{ displayText: "TestReq", format: "text" }]
+  };
+
+  public render(): React.ReactNode {
+    if (this.props.modelExplanationData) {
+      for (const exp of this.props.modelExplanationData) {
```
hawestra, line 34: nit: this if is not needed right ? you could just iterate through this.props.modelExplanationData which if it's not defined wouldn't do anything

apps/dashboard/src/app/App.tsx
```diff
@@ -175,6 +176,21 @@ export class App extends React.Component<IAppSetting, IAppState> {
...
+                this.state.dataset
+              ]}
+              theme={themes[this.state.theme]}
+              language={Language[this.state.language]}
+              version={
+                applications[this.state.application].versions[
+                  this.state.version
+                ]
+              }
+              featureFlights={parseFeatureFlights(this.state.featureFlights)}
```
hawestra, line 191: I'm guessing you copied this from the lines above but we're calling parseFeatureFlights a lot; could we save as a local variable and use that instead ? or is there any reason that this.state.featureFlights is not the parsed featureFlights ?

hawestra (approved):

apps/dashboard/src/app/App.tsx
romanlutz: This is a complicated one 🙂 
If I remember correctly it would be easiest to represent it as a string[] everywhere, BUT this is one of the dropdowns in the test env so that we can try it out with or without flights (and also combinations thereof). That, in turn, means we need to be able to encode it in the URL, e.g., 
```
/modelAssessment/adultCensusIncomeData/1%20-%20light/en/1/flight1&flight2&flight3
```
So then at some point we have to parse it from this string to string[]. I think that part is intuitive, but you asked a great question: why not have it as string[] on the state? 
The answer is that the state consists of `IAppSetting`s and those defined as follows
```
export const routeKey = [
  "application",
  "dataset",
  "theme",
  "language",
  "version",
  "featureFlights"
] as const;
export type IAppSetting = {
  [key in typeof routeKey[number]]?: string;
};
```
In other words, we have no choice but to make it a string. We could, of course, have a duplicate and try to keep them in sync, but repeating the parse function a few times seems like a small price to pay in order not to have to keep them synced.

I may very well have missed something somewhere that makes this easily solvable. In that case, we should definitely adjust. Wdyt?

apps/dashboard/src/model-assessment-forecasting/App.tsx
romanlutz: This is indeed a strange way of writing this. Let me adjust that....

libs/forecasting/.eslintrc.json
romanlutz: Ah I think I get it now! This is on with "error" by default via .eslintrc\.eslintrc.custom.eslintrc, and some of the projects have it turned off, but not forecasting, so it is on.

PR #1892, opened 2023-01-07
Add `test_data` to `ICounterfactualData`

gaugup:
## Description

Add `test_data` to `ICounterfactualData` so that UI can make use of the test_data returned by SDK backend for counterfactual table. This is related to the PR:- https://github.com/microsoft/responsible-ai-toolbox/pull/1885.

## Checklist


vinuthakaranth (approved):

romanlutz:
What exactly is `test_data` here? Is that the test dataset? If so, are we duplicating it? I might be missing the larger point as this is distributed over several PRs, but passing around an entire dataset (additionally to what we're already passing around it seems?) worries me. A pointer to a centralized design document that explains the big picture while referencing these individual pieces of work may also help.

gaugup:
> What exactly is `test_data` here? Is that the test dataset? If so, are we duplicating it? I might be missing the larger point as this is distributed over several PRs, but passing around an entire dataset (additionally to what we're already passing around it seems?) worries me. A pointer to a centralized design document that explains the big picture while referencing these individual pieces of work may also help.

For large data scenario, not all the test samples will be present in UI. For local counterfactuals we would need to respond with the test dataset row. The size of the test data is not a concern as there is only one point returned for local counterfactuals. For initial run of rai, we limit the number of test samples to 5k. Please bear in mind that the generated counterfactuals per point determine the size of the output and not the test data itself.

We discovered this late for local counterfactual work that @vinuthakaranth is driving. The design doc for large is already shared with the PM and dev teams to refer to.

PR #1893, opened 2023-01-07
add real model for dev env to have better test data

xuke444:
use pyroide to train python model directly in browser.
so that dev env can have meaningful values.


xuke444:
image

PR #1894, opened 2023-01-09
Add forecasting dashboard - UI - time series selection & forecast comparison

romanlutz:
## Description

This is a copy of #1874 but trimmed down to allow for easier code reviews. The removed portion is about what-if transformations, so this PR only enables the selection of a time series (internally "cohort") and then automatically generates a forecast comparison plot:
image

Once this is merged and #1874 is updated, #1874 will provide the diff, i.e., it will add transformations, a UI to create them, a table to present them, and code to generate the corresponding forecasts to add them to the forecast comparison plot.

Tests will be added end of January/early February to accelerate the POC and get feedback from customers for validation purposes first.

## Checklist


libs/forecasting/src/lib/ForecastingDashboard/Controls/ForecastComparison.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { getForecastPrediction } from "./getForecastPrediction";
+
+export class IForecastComparisonProps {}
+
+export interface IForecastComparisonState {
+  timeSeriesId?: number;
+  baselinePrediction?: number[];
+}
+
+const stackTokens = {
```
vinuthakaranth, line 27: nit: should we move this to forecastingDashboardStyles?

libs/forecasting/src/lib/ForecastingDashboard/Controls/ForecastComparison.tsx
```diff
@@ -0,0 +1,167 @@
...
+        baselinePrediction,
+        timeSeriesId: currentlySelectedTimeSeriesId
+      });
+    }
+  }
+
+  public render(): React.ReactNode {
+    const classNames = forecastingDashboardStyles();
+
+    const indices = this.context.dataset.index;
```
vinuthakaranth, line 70: Since we are working on big data, will there be any issue with rendering big data here?

libs/forecasting/src/lib/ForecastingDashboard/Controls/ForecastComparison.tsx
```diff
@@ -0,0 +1,167 @@
...
+                    }
+                  }
+                },
+                series: seriesData,
+                title: {
+                  text: localization.Forecasting.forecastComparisonChartTitle
+                },
+                xAxis: {
+                  dateTimeLabelFormats: {
+                    // don't display the year
```
vinuthakaranth, line 134: just for my knowledge: why are we not displaying year? what does year: "%Y" do?

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.styles.ts
```diff
@@ -0,0 +1,87 @@
...
+    return mergeStyleSets<IForecastingDashboardStyles>({
+      configurationActionButton: {
+        marginTop: "25px"
+      },
+      descriptionText: {
+        color: theme.semanticColors.bodyText,
+        maxWidth: descriptionMaxWidth
+      },
+      dropdown: {
+        width: "400px"
```
vinuthakaranth, line 42: please help check reflow cases. As fixed widths can cause issues in 320*256 screens, most of the time it creates unwanted horizontal scroll bar. thanks.

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
```diff
@@ -1,19 +1,101 @@
...
+    if (
+      this.context === undefined ||
+      this.context.baseErrorCohort === undefined
+    ) {
+      return;
+    }
+
+    // "All data" cohort selected, so no particular time series selected yet.
+    // special case: only 1 time series in dataset, needs to be handled! TODO
+    const noCohortSelected =
```
vinuthakaranth, line 39: for my clarification: Will this work if user is using another language? will not be false in that case?

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
```diff
@@ -1,19 +1,101 @@
...
+            label={localization.Forecasting.timeSeries}
+            className={classNames.dropdown}
+            options={this.context.errorCohorts
+              .filter(
+                (cohort) =>
+                  cohort.cohort.name !==
+                  localization.ErrorAnalysis.Cohort.defaultLabel
+              )
+              .map((cohort) => {
+                return {
```
vinuthakaranth, line 63: nit: can we have this as a function?

libs/model-assessment/src/lib/ModelAssessmentDashboard/AvailableTabs.ts
```diff
@@ -21,24 +22,35 @@ export function getAvailableTabs(
       text: localization.ModelAssessment.ComponentNames.ErrorAnalysis
     });
   }
-
+  if (props.dataset.task_type === DatasetTaskType.Forecasting) {
+    availableTabs.push({
+      key: GlobalTabKeys.ForecastingTab,
+      text: "Forecasting"
```
vinuthakaranth, line 28: localization

libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/CohortList.tsx
```diff
@@ -178,13 +179,15 @@ export class CohortList extends React.Component<
...
                   </Stack>
                 </Stack.Item>
-                <Stack.Item>
-                  <CohortListDeleteButton
-                    disabled={this.isActiveCohort(item.key)}
-                    itemKey={item.key}
-                    onDeleteCohortClick={this.onDeleteCohortClick}
-                  />
-                </Stack.Item>
+                {this.props.enableEditing && (
```
vinuthakaranth, line 182: for my knowledge: why is this added now?

libs/forecasting/src/lib/ForecastingDashboard/Controls/ForecastComparison.tsx
romanlutz: AFAIK stack tokens don't quite work the same way, but I would love to be corrected on this as it really should live in a style file. If you have an example where we do it that way I will be happy to make the PR conform 🙂

libs/forecasting/src/lib/ForecastingDashboard/Controls/ForecastComparison.tsx
romanlutz: Potentially! In short: we need to have a means to get the true Y, predicted Y, and any ad-hoc predictions based on transformations (that's the follow-on PR), so if there's a long forecast horizon (meaning we want not just the predictions for the next 5 points in time, but, say, thousands of them) then that could very well result in a "big data" case that should be handled by the backend. I will check with forecasting folks on typical forecast horizons and get back to you. I'll consider this a larger discussion that isn't blocking this PR if that's okay with you.

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.styles.ts
romanlutz: Couldn't agree more! In my experience, dropdowns have proven tricky in this regard as they don't seem to be flexible based on their input params. If I didn't know fluent UI I would have expected something like min, max, and default width, but I have only the one input param. Do you have any suggestions on how to handle that? I don't want it to be too short by default because it results in poor readability of the potentially long time series names.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/CohortList.tsx
romanlutz: Great question! Cohorts (other than "All data") can generally be edited. Forecasting is a bit of an oddball there in that not all cohorts are really valid as time series. Time series require filters based on "time series id columns." So rather than making users create these by hand (which is prone to errors) we generate all of them ahead of time and pass them into the dashboard. That means they don't need to have editing capabilities as that wouldn't accomplish anything. There is a special case where one might want to merge some of these together, but that's not in scope for the MVP. [I'm planning to pass in some merged time series of interest at a later point if users are interested in that.]

libs/forecasting/src/lib/ForecastingDashboard/Controls/ForecastComparison.tsx
romanlutz: This is a leftover from John's POC. I actually think we do want to show year, at least when the time granularity isn't in something smaller than hours.

libs/forecasting/src/lib/ForecastingDashboard/Controls/ForecastComparison.tsx
romanlutz: Update: we expect on the order of tens and hundreds of predictions per time series, but not thousands because forecasting (intuitively) gets worse and worse the further you go. So I don't see this as a big issue in terms of data getting too big. I've made a note about input validation to warn users when they provide too big of a forecast horizon. This will be in my follow-up follow-up PR 🙂

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
romanlutz: The cohort name of the All data cohort will then match defaultLabel in that language, so for example, in German it will be "Alle Daten" === "Alle Daten" in case it is truly the All data cohort which in the forecasting scenario means that we haven't selected a time series. I will handle the single time series case mentioned in the comment in a follow-up PR

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
romanlutz: Certainly!

libs/forecasting/src/lib/ForecastingDashboard/Controls/ForecastComparison.tsx
romanlutz: I checked a dozen examples and this seems to be the way we're doing it everywhere else for stack tokens. I'm marking this one as resolved, but do feel free to reopen if you have a suggestion for a cleaner way!

libs/forecasting/src/lib/ForecastingDashboard/Controls/ForecastComparison.tsx
romanlutz: Highcharts has some logic to decide the appropriate granularity when presenting dates so I'm removing this entirely and will just let the type "datetime" stand so that it can make a good choice. The tooltip has the full information regardless.

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
romanlutz: ... not that that should mean we have to keep doing it that way. Perhaps a flag on the cohort to indicate that it's the "all data" cohort would be useful to make this cleaner. I'm not a fan of this either... If you like this idea please upvote and I'll create a separate PR fixing all instances. If you have a better idea I'm all ears 🙂

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
romanlutz: Actually, moving this out of the JSX portion helps a ton in making it more readable. Thanks for pointing me to this unsightly code 🙂 . I don't think breaking it up further will help in that direction, but do reopen if you disagree.

libs/model-assessment/src/lib/ModelAssessmentDashboard/AvailableTabs.ts
romanlutz: Done

libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/CohortList.tsx
romanlutz: Resolving this one, but do reopen if you have questions.

libs/core-ui/src/lib/Context/ModelAssessmentContext.tsx
```diff
@@ -112,6 +112,10 @@ export interface IModelAssessmentContext {
   requestExp?:
     | ((index: number, abortSignal: AbortSignal) => Promise<any[]>)
     | undefined;
+  requestForecast?: (
+    request: any[],
+    abortSignal: AbortSignal
+  ) => Promise<number[]>;
```
xuke444, line 118: Isn't this same as request prediction?

libs/core-ui/src/lib/util/TimeUtils.ts
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT License.
+
+interface ITimeBasedValue {
+  time: number;
+  value: number;
+}
+
+export function orderByTime(values: number[], rowIndices: string[]): any[][] {
+  return values
+    .map((predictedValue: number, idx: number) => {
```
xuke444, line 11: this part seems able to be optimized, instead map to obj, sort then map to array, can map to array directly and sort by a[0] - b[0], lodash may even have default behave to sort array by first element, so we can remove call back in sort function

libs/core-ui/src/lib/util/string.ts
```diff
@@ -10,3 +10,7 @@ export function limitStringLength(
   }
   return str;
 }
+
+export function isUndefinedOrEmpty(s: string | undefined): boolean {
+  return s === undefined || s === "";
+}
```
xuke444, line 16: this can simply replaced by (!s)

libs/forecasting/src/lib/ForecastingDashboard/Controls/ForecastComparison.tsx
```diff
@@ -0,0 +1,161 @@
...
+    // In that case, we need to update our state accordingly.
+    const currentlySelectedTimeSeriesId =
+      this.context.baseErrorCohort.cohort.getCohortID();
+    if (currentlySelectedTimeSeriesId !== this.state.timeSeriesId) {
+      const baselinePrediction = await this.getBaselineForecastPrediction();
+      this.setState({
+        baselinePrediction,
+        timeSeriesId: currentlySelectedTimeSeriesId
+      });
+    }
```
xuke444, line 64: You will probably need to handle api call serilization in following pr.
When user change time series before api returns data, and first api call takes longer than second one, we may end up with having wrong data/chart.

libs/forecasting/src/lib/ForecastingDashboard/Controls/ForecastComparison.tsx
```diff
@@ -0,0 +1,161 @@
...
+    const rowIndices = this.context.baseErrorCohort.cohort.filteredData.map(
+      (datum) => indices[datum.Index]
+    );
+
+    let trueY: SeriesOptionsType[] = [];
+    if (this.context.jointDataset.numLabels === 1) {
+      trueY = [
+        {
+          data: orderByTime(
```
xuke444, line 82: sorting is usually intense, is it worthy to use state to cache?

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
xuke444: Are we able to check filter.length == 0?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/ShiftCohort.tsx
```diff
@@ -41,9 +42,14 @@ export class ShiftCohort extends React.Component<
     defaultModelAssessmentContext;
 
   public componentDidMount(): void {
-    const savedCohorts = this.context.errorCohorts.filter(
-      (errorCohort) => !errorCohort.isTemporary
-    );
+    const savedCohorts = this.context.errorCohorts
+      .filter((errorCohort) => !errorCohort.isTemporary)
```
xuke444, line 46: seems 2 filters can be combined

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/TabsView/TabsView.tsx
```diff
@@ -180,6 +186,19 @@ export class TabsView extends React.PureComponent<
                     />
                   </>
                 )}
+              {t.key === GlobalTabKeys.ForecastingTab && (
+                <>
+                  <h3
```
xuke444, line 191: we try to move away from browser default h1,h2...
is <Label> from fluent ui good enough?

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
romanlutz: I like that! I'll write it as a utility function somewhere in core-ui and use it from there.

libs/core-ui/src/lib/Cohort/Cohort.ts
```diff
@@ -15,7 +15,8 @@ export enum CohortSource {
   None = "None",
   TreeMap = "Tree map",
   HeatMap = "Heat map",
-  ManuallyCreated = "Manually created"
+  ManuallyCreated = "Manually created",
+  UserProvidedTimeSeries = "User-provided time series"
```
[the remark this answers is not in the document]
romanlutz: They are provided via the pre-built cohort feature in the SDK right now, yes. It's just that forecasting is somewhat more restrictive in that you can't add new cohorts/time series.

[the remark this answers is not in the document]
romanlutz: Which reminds me: why don't pre-built cohorts show up here?  @gaugup

libs/core-ui/src/lib/Context/ModelAssessmentContext.tsx
romanlutz: Not quite. We have to be able to provide transformations on a time series. In other words, we provide the transformation operation (let's say method = "Add", value = 5, feature = "abc") so this needs to be applied to all data points that are part of this time series and then we need to get the predictions. Doing this in the backend seems smarter to me than duplicating the data in the UI and applying the transformation before sending all that data across. 

The "other" what-if (for tabular data) creates individual transformations that don't generalize beyond a single point so perhaps there isn't much of a difference there and that's why it uses `requestPredictions`.
Wdyt?

libs/forecasting/src/lib/ForecastingDashboard/Controls/ForecastComparison.tsx
romanlutz: Do we have an example of that somewhere?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/TabsView/TabsView.tsx
romanlutz: I just copied what the other tabs are doing, e.g., Error Analysis right below. What would you prefer? `<Text variant="xLarge">...</Text>` ? I can change all other occurrences, too.

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.styles.ts
vinuthakaranth: We can add this conditionally for large screens as in WhatIfTab.styles
 selectors: {
          "@media screen and (min-width: 1024px)": {
            paddingRight: legendWidth
          }
        }

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
romanlutz: Actually, it's more complicated. Users could create cohorts without filters that have a different name. Perhaps with injection of pre-built cohorts it's even possible to add duplicates with the same name (?)
I'm adding a utility function that does the filter comparison AND the string comparison because I'm afraid of introducing regressions in some of our older code if I remove this. Once we're more confident (e.g., by adding some additional identifier or flag indicating that it's the All data cohort, or by enforcing that there's only one cohort without filters) we can get rid of the string comparison part. 

The forecasting code doesn't care about all that and can work with just the filter-based comparison.

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.styles.ts
romanlutz: Great! I made it 
```
dropdown: {
        width: "auto",
        selectors: {
          "@media screen and (min-width: 1024px)": {
            width: "600px"
          }
        }
      },
```
so for small screens it's just the full width and as soon as we have a larger screen it's 600px. I know that's pretty wide but the time series are named after their id column names + values, e.g.,
image
so this can potentially get quite long. I've made it so that it's not longer than the text above which makes it look nice visually IMO.

libs/core-ui/src/lib/Cohort/Cohort.ts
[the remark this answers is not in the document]
romanlutz: I did as you suggested and "merged" it with prebuilt as that's a more general name for what we're doing here. I also made sure that it's actually used (in `ProcessPreBuiltCohort.ts`)

libs/core-ui/src/lib/util/TimeUtils.ts
romanlutz: Fair! I fixed that.

libs/forecasting/src/lib/ForecastingDashboard/Controls/ForecastComparison.tsx
romanlutz: Done!

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/TabsView/TabsView.tsx
xuke444: <Text variant="xLarge"> is good.
EA could be caused by legacy code.

romanlutz: Sounds good. I'll create a separate PR since there are quite a few of them.

libs/core-ui/src/lib/Context/ModelAssessmentContext.tsx
romanlutz: Settled on this offline. Marking as resolved.

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.styles.ts
romanlutz: I'm assuming this solves the problem, marking as resolved, but feel free to reopen if it's not what you were thinking of!

vinuthakaranth (approved):

libs/forecasting/src/lib/ForecastingDashboard/Controls/ForecastComparison.tsx
romanlutz: Checked: `AbortController` code will have an example. Will add in next PR 🙂

libs/core-ui/src/lib/Cohort/isAllDataCohort.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation.
```
imatiach-msft, line 1: what do you think about a more generic filename like allDataCohortUtils.ts?  Or defaultCohortUtils.ts?
I feel like more utils could eventually go into this file related to the "All Data" cohort eventually?

romanlutz: I don't really care either way. But I can change that in the follow-up PR that I'm publishing later today. Marking as resolved but rest assured I'll do it there 😉

libs/forecasting/src/lib/ForecastingDashboard/Controls/getForecastPrediction.ts
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { Cohort, JointDataset } from "@responsible-ai/core-ui";
+
+export async function getForecastPrediction(
```
imatiach-msft, line 6: why do we need a special one for forecasting instead of using the regular prediction one?

imatiach-msft (approved):

libs/forecasting/src/lib/ForecastingDashboard/Controls/getForecastPrediction.ts
romanlutz: Because it's more efficient. We don't send data across, just the transformation description that applies to the whole time series.

PR #1897, opened 2023-01-11
Fix unwrap error in individual feature importance

hawestra:
## Description
Use all data cohort as the default selected cohort if the currently selected cohort was deleted while the individual feature importance was using it. This bug was caused by React lifecycle events; the render function is using this.state.selectedCohortIndex on render but when the selected cohort gets deleted, this.props.cohorts is updated causing a re-render but this.state.selectedCohortIndex isn't updated until componentDidUpdate causing an out of index error. 

## Checklist


gaugup (approved):

imatiach-msft (approved):

libs/interpret/src/lib/MLIDashboard/Controls/GlobalExplanationTab/GlobalExplanationTab.tsx
```diff
@@ -181,6 +181,10 @@ export class GlobalExplanationTab extends React.PureComponent<
       this.context.jointDataset.datasetMetaData?.featureMetaData
         ?.dropped_features
     );
+    const selectedCohort =
+      this.state.selectedCohortIndex >= this.props.cohorts.length
+        ? this.props.cohorts[0]
+        : this.props.cohorts[this.state.selectedCohortIndex];
```
romanlutz, line 187: So this captures the case of going beyond the bounds, but if that's possible that indicates to me that we're not updating properly. So let's say there are cohorts `[0,1,2]` and we delete 2 then that's your case here, we fix that by going to 0. Great! But if we delete 1 then that would put us to.... 2? That seems unintuitive, doesn't it? 
There's a very decent chance that I'm misunderstanding what's going on here, though 😄

tongyu-microsoft: Agree that this happens because we are not updating properly. This also happened in other component for undefined cohorts. I am thinking if it is more user-friendly by disabling deletion of current selected cohort? So user won't be confused by we auto select another cohort when they delete the current one.

vinuthakaranth: I think we are already doing this. We won't let users delete current cohort.

hawestra: good points! I updated the files so that we calculate the selected cohort when a cohort is deleted. This would take care of the case you mentioned @romanlutz so the selected cohort persists. In the case that the user deleted the cohort that they are currently viewing, right now I have the selected cohort to default to the All data default cohort since cohort is required to be defined in FeatureImportanceDependence

romanlutz (approved):
Thanks for addressing!

xuke444 (approved):

PR #1903, opened 2023-01-13
Funneling metric calculation via asynchronous metric functions

gaugup:
## Description

Funneling metric calculation via asynchronous metric functions. The `generateMetricsCohortsCombined` is a asynchronous wrapper which calls backend SDK to fetch metrics in case such an endpoint is available.

## Checklist

### DIFF
--- a/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
+++ b/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
@@ -34,9 +34,6 @@ import {
 } from "@responsible-ai/core-ui";
 import { localization } from "@responsible-ai/localization";
 import React from "react";
-import {
-  generateMetricsCohortsCombined
-} from "./Metrics";
 
 import { ChartConfigurationFlyout } from "./ChartConfigurationFlyout";
 import { defaultNumberOfContinuousFeatureBins } from "./Constants";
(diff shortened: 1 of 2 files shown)


libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -62,6 +61,7 @@ interface IModelOverviewState {
   datasetCohortViewIsVisible: boolean;
   datasetCohortChartIsVisible: boolean;
   featureConfigurationIsVisible: boolean;
+  isDatasetCohortMetricCalcaulationInProgress: boolean;
```
romanlutz, line 64:
```suggestion
  isDatasetCohortMetricCalculationInProgress: boolean;
```

romanlutz, line 64: are we tackling feature cohort metric calculation separately?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -168,30 +169,36 @@ export class ModelOverview extends React.Component<
...
-        (x) => !newDatasetCohortIDs.includes(x)
-      );
-      if (addCohortIDs.length > 0) {
-        this.setState(
-          {
-            selectedDatasetCohorts:
-              this.state.selectedDatasetCohorts?.concat(addCohortIDs)
-          },
-          () => this.updateDatasetCohortStats()
+    if (!this.state.isDatasetCohortMetricCalcaulationInProgress) {
```
romanlutz, line 172: So while this is in progress we don't allow changes to dataset cohorts if I read this right? What if the user creates or deletes a new cohort in the meanwhile? Or are we blocking the entire UI for the duration of the calculation?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -205,6 +212,17 @@ export class ModelOverview extends React.Component<
       );
     }
 
+    if (this.state.datasetCohortLabeledStatistics.length === 0) {
```
romanlutz, line 215: This makes sense. I do wonder what would happen if the request failed and we can't get the data. Will we just ask the user to wait forever?

PR #1905, opened 2023-01-13
Enable logarithmic scaling, reflect cohort change and add add Revert to bubble chart button

vinuthakaranth:
## Description

This PR contains changes in counterfactuals for below:
1. Enable logarithmic toggle for scatter lot
2. Add revert button to render bubble chart from scatter plot to select different cluster
3. On cohort change update bubble chart

enaLogRevertBuCohort

## Checklist


libs/counterfactuals/src/lib/CounterfactualComponent.tsx
```diff
@@ -53,6 +54,7 @@ export interface ICounterfactualComponentState {
   indexSeries: number[];
   isCounterfactualsDataLoading?: boolean;
   localCounterfactualErrorMessage?: string;
+  isRevertButtonClicked: boolean;
```
romanlutz, line 57: I'm slightly confused why a button click has an entire state variable representing it. Is there some async call that happens after the button click that we need to wait for?

If not, is it possible to collapse the changes happening to set and unset this flag into a single set of updates (thus rendering this state variable unnecessary)?

libs/counterfactuals/src/lib/CounterfactualComponentUtils.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+enum FieldChangeUpdate {
+  Dither = "dither",
+  Property = "property",
+  Type = "type"
+}
+
+export function isJustTypeChange(changedKeys: string[]): boolean {
```
romanlutz, line 10: I'm confused by the "just" 🙂 Based on the comment below would "hasAxisTypeChanged" be a suitable name?

libs/counterfactuals/src/lib/CounterfactualComponentUtils.ts
```diff
@@ -0,0 +1,25 @@
...
+  ) {
+    return true;
+  }
+  return false;
+}
+
+function removeParentKeys(changedKeys: string[]): string[] {
+  const valuesToRemove = new Set(["options", "xAxis", "yAxis"]); // Since chartProps is a nested object, these are parent keys which are usually changed if inner keys are changed.
+  return changedKeys.filter((item) => !valuesToRemove.has(item));
+}
```
romanlutz, line 25: What does "parent" refer to here?

vinuthakaranth: It is a nested obeject. For ex:
const a = {name: "xyz",
address: {
street: "123 st",
area: "wer"},
dob:"123223"}

here address is parent key, as if area has changed, then address has also changed.

libs/counterfactuals/src/lib/CounterfactualComponent.tsx
vinuthakaranth: I need this variable to propagate to LargeCounterfactualChart to update bubble chart. I am happy to update if there is any other recommended way to handle this.

libs/counterfactuals/src/lib/CounterfactualComponentUtils.ts
vinuthakaranth: thank you, updated

libs/counterfactuals/src/lib/CounterfactualComponent.tsx
romanlutz: I see. And LargeCounterfactualChart is within this one. I guess the pattern I've seen for this type of situation is to have the state at a higher level (e.g., in this file) and pass it in. Can't say I like that a lot either since a lot of state can get concentrated at higher levels, but at least the state update happens in 1 step rather than 2.

romanlutz (approved):

PR #1908, opened 2023-01-14
[draft] upgrade nx

hawestra:
## Description
[work in progress] upgrade nx to version ~15 from version ~12

## Checklist


nx.json
```diff
@@ -16,8 +16,10 @@
   },
   "tasksRunnerOptions": {
     "default": {
-      "runner": "@nrwl/workspace/tasks-runners/default",
-      "options": {}
+      "runner": "@nrwl/nx-cloud",
+      "options": {
+        "accessToken": "YzhiYzM3NGUtNGUyMi00NDlmLTlhMTEtODlkNzY1MGUxN2ExfHJlYWQtd3JpdGU="
```
romanlutz, line 21: What's this for?

romanlutz (commented):
Is it fair to assume that workspace.json just got split up into the project.json files or is there anything that actually changed?

Also, what caused all the yarn.lock changes?

nx.json
hawestra: nx cloud is an optimization that was created in the more recent versions of nx. It makes it so that you can distribute computation cache across machines (here's more information about it: https://nx.dev/nx-cloud/intro/what-is-nx-cloud) but the tldr is it's an optimization to make developing faster by using a shared cache

PR #1824, opened 2022-12-01
Enable dropped_features for error analysis, data explorer and explainer, counterfactual and causal managers 

tongyu-microsoft:
This PR adds dropped_features and pass dropped_features to error analysis, data explorer and explainer managers if there is any.

GIF:
in the below regression example, we dropped ["s1", "s5"] features:
dropped_features_RA

for counterfactual:
dropped_features_cf

for causal:
dropped_features_causal

## Description
1. for error analysis, we display dropped_features, since feature importances are the correlation between the feature column and the error column, and any feature column can be computed.
2. for explainer, we hide dropped_features, since displaying dropped_features that aren't fed to the model doesn't make sense.
3. for data explorer, we still show the dropped_features.
4. for counterfactual, we hide the dropped_features.
5. for causal, we hide the dropped_features.

We will update the example notebooks, doc and add more tests in my next PR.

## Checklist

- [x] I have added screenshots above for all UI changes.
- [ ] I have added e2e tests for all UI changes.
- [ ] Documentation was updated if it was needed.

PR #1912, opened 2023-01-17
Fix model wrapper to identify model type, remove task type from error analysis

tongyu-microsoft:
This PR fixes an issue that blocks curated image release and other RAI repo releases. 

## Description

The issue is, the model wrapper created in PR https://github.com/microsoft/responsible-ai-toolbox/pull/1824 has both predict function and predict_proba function. In error analysis, we check if the model has predict_proba function to identify if it is a classification model or regression model. In the case model is wrapped, it will always identify the model as classification model. 

The fix is, we check if the model is classification model or regression model, and wrap the model in a separate function. So regression model won't have predict_proba function after wrapping. In this case, we don't need to specifically pass task_type into error_analysis_manager. So this PR also removes task_type from error_analysis_manager

## Checklist

### DIFF
--- a/responsibleai/responsibleai/managers/error_analysis_manager.py
+++ b/responsibleai/responsibleai/managers/error_analysis_manager.py
@@ -16,7 +16,7 @@ from erroranalysis._internal.error_report import as_error_report
 from erroranalysis._internal.error_report import \
     json_converter as report_json_converter
 from responsibleai._config.base_config import BaseConfig
-from responsibleai._interfaces import ErrorAnalysisData, TaskType
+from responsibleai._interfaces import ErrorAnalysisData
 from responsibleai._internal.constants import ErrorAnalysisManagerKeys as Keys
 from responsibleai._internal.constants import ListProperties, ManagerNames
 from responsibleai._tools.shared.state_directory_management import \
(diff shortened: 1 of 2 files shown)


gaugup (commented):
Are there any tests which validate these changes?

responsibleai/responsibleai/managers/error_analysis_manager.py
```diff
@@ -111,6 +124,30 @@ def _apply_func(self, func, dataset):
         return func(dataset.drop(columns=self.dropped_features, axis=1))
 
 
+class MetadataRemovalRegressionModelWrapper():
```
imatiach-msft, line 127: An alternative might have also been to create a base class with the predict method and have classification/regression extend it, but I'm not sure if that would have saved enough code lines anyway, although maybe it might have reduced a few duplicate lines to a common place.  I think this might still be better.

responsibleai/responsibleai/managers/error_analysis_manager.py
```diff
@@ -188,8 +225,7 @@ class ErrorAnalysisManager(BaseManager):
     def __init__(self, model: Any, dataset: pd.DataFrame, target_column: str,
                  classes: Optional[List] = None,
                  categorical_features: Optional[List[str]] = None,
-                 dropped_features: Optional[List[str]] = None,
-                 task_type: Optional[TaskType] = None):
+                 dropped_features: Optional[List[str]] = None):
```
imatiach-msft, line 228: passing task_type might still be fine, but default would need to be ModelTask.UNKNOWN instead of None.  If removing it here works that's fine too though, since this is how it was before anyway it seems.

imatiach-msft (approved):

responsibleai/responsibleai/managers/error_analysis_manager.py
tongyu-microsoft: yeah we don't pass task_type here before. So removing it as it is not needed

responsibleai/responsibleai/managers/error_analysis_manager.py
tongyu-microsoft: thanks for the great comment!
Since we want to merge this PR today, I will merge it now as all tests passed. And I will address this comment in my next PR.

tongyu-microsoft:
> Are there any tests which validate these changes?

I will add more tests in my next PR, as I test locally and I would like to merge this PR asap. Thanks!

PR #1913, opened 2023-01-17
[Large data counterfactuals] Use data from local counterfactuals to display in counterfactuals panel

vinuthakaranth:
## Description
Large data uses test_data from local counterfactuals call to make predict call and display data in panel.

oriDataFromAPI

image

## Checklist


libs/counterfactuals/src/util/getOriginalData.ts
```diff
@@ -1,26 +1,43 @@
...
+  JointDataset
+} from "@responsible-ai/core-ui";
 import { localization } from "@responsible-ai/localization";
 
 export function getOriginalData(
   index: number,
   jointDataset: JointDataset,
-  dataset: IDataset
+  dataset: IDataset,
+  ifEnableLargeData: boolean,
```
gaugup, line 15: We don't need the `ifEnableLargeData`. We can directly apply the IfEnableLargeData() on dataset.

vinuthakaranth: Great catch! updated

libs/counterfactuals/src/util/getFetchPredictionPromise.ts
```diff
@@ -6,14 +6,23 @@ import { JointDataset } from "@responsible-ai/core-ui";
...
-  const promise = invokeModel([rawData], abortController.signal);
+  let rawData = [
+    JointDataset.datasetSlice(
+      fetchingReference,
+      jointDataset.metaDict,
+      jointDataset.datasetFeatureCount
+    )
+  ];
+  if (ifEnableLargeData && testData) {
+    testData[0][0].pop(); // drop the prediction value before predict call
```
imatiach-msft, line 22: Since we are mutating this value might this affect any visualizations that might render during this call - or do we do a deep copy of testData before this somewhere/not use it directly in visualizations?

imatiach-msft (approved):

libs/counterfactuals/src/util/getFetchPredictionPromise.ts
romanlutz, line 22: I'm not quite understanding why we're dropping this value. Can you elaborate?

romanlutz (approved):

libs/counterfactuals/src/util/getFetchPredictionPromise.ts
gaugup: Should this be the last value in the list? Could you plese confirm?

gaugup (changes requested):

libs/counterfactuals/src/util/getFetchPredictionPromise.ts
vinuthakaranth: Thanks @imatiach-msft !! we are using using test data to compare and display data in panel. I updated to use cloned data to be used in predict call.
@gaugup yes, I am just removing last value from the list.

libs/counterfactuals/src/lib/CounterfactualChartLegend.tsx
```diff
@@ -168,9 +168,12 @@ export class CounterfactualChartLegend extends React.PureComponent<ICounterfactu
   };
 
   private disableCounterfactualPanel = (): boolean => {
+    const cfsList = ifEnableLargeData(this.context.dataset)
```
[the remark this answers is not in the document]
vinuthakaranth: updated :)

gaugup (approved):

PR #1921, opened 2023-01-23
Forecasting RAI dashboard: expand test data, add categorical what if

romanlutz:
## Description

This PR is the result of updates and fixes I made while building the Python backend side that matches the forecasting RAI UI portion. This includes:

- expanding the test data shown in the test env to include a case with no features (which means What-if won't work as we can't change features) and a case with only a single time series (which means we can't choose from a set of time series) in addition to the existing case with 3 separate time series. The latter two cases now also have categorical features that can be changed.
- adding support for changing categorical features to the What-if dialog. This prompted me to split up the logic into separate components since the contents are sufficiently different.
- removing leftover occurrences of `datetime_features` (plural) by changing them to `time_column_name` (singular)
- moving `getRawValue` function to `JointDataset` since it operates on the joint dataset anyway. We need this in order to convert values back and forth from UI to backend as we ask for forecasts.
- handling case when there's only a single time series (= All data)
- creating "What If section" as a separate component consisting of What-if button and the table of What-If transformations. This won't be visible if there are no features to change, so it encapsulates the What-if-only experience well.
- small adjustments to cohort settings and shift view to reflect "time series" instead of "cohort" for forecasting tasks.

As mentioned before, e2e tests to follow in early February and certainly before making this available to anyone (date TBD).

## Checklist


libs/core-ui/src/lib/util/JointDataset.ts
```diff
@@ -420,6 +420,23 @@ export class JointDataset {
     return [];
   }
 
+  public readonly getRawValue = (
+    v: number | undefined,
```
hawestra, line 424: nit: you could combine the two if statements to simplify the funciton: 
`if( v !== undefined &&  (meta.isCategorical || meta?.treatAsCategorical) && meta.sortedCategoricalValues))
 { return meta.sortedCategoricalValues[v]; }
return v;`

libs/localization/src/lib/en.json
```diff
@@ -1841,18 +1843,24 @@
   "Forecasting": {
     "target": "Target",
     "whatIfHeader": "What-if analysis",
-    "whatIfDescription": "What-if allows you to perturb features for any input and observe how the model's prediction changes. You can perturb features manually or specify the desired prediction (e.g., class label for a classifier) to see a list of closest data points to the original input that would lead to the desired prediction. Also known as prediction counterfactuals, you can use them for exploring the relationships learnt by the model; understanding important, necessary features for the model's predictions; or debug edge-cases for the model. To start, choose input points from the data table or scatter plot.",
+    "forecastHeader": "Forecast analysis",
+    "whatIfDescription": "What-if allows you to perturb features for your entire time series and observe how the model's forecast changes.",
```
hawestra, line 1847: should this be what-if analysis ? It seems a little weird to me just saying "what-if"

hawestra: or what-if forecasts

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
```diff
@@ -41,38 +38,65 @@ export class ForecastingDashboard extends React.Component<
...
+      isAllDataErrorCohort(this.context.baseErrorCohort) &&
+      this.state.timeSeriesOptions.length > 1;
 
     const cohortTransformations =
       this.state.transformations?.get(
         this.context.baseErrorCohort.cohort.getCohortID()
       ) ?? new Map<string, Transformation>();
 
+    const description =
+      (this.state.whatIfEnabled
```
hawestra, line 93: nit: a small helper function might make this description creation a little clearer

libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationCreationCategorical.tsx
```diff
@@ -0,0 +1,105 @@
...
+export class TransformationCreationCategorical extends React.Component<ITransformationCreationCategoricalProps> {
+  public static contextType = ModelAssessmentContext;
+  public context: React.ContextType<typeof ModelAssessmentContext> =
+    defaultModelAssessmentContext;
+
+  public render(): React.ReactNode {
+    const classNames = forecastingDashboardStyles();
+    let options = undefined;
+    if (this.props.transformationFeature) {
+      options = this.context.jointDataset.metaDict[
```
hawestra, line 39: could you add a comment here ? there's a lot of indexing going on

libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationCreationDialog.tsx
```diff
@@ -191,6 +181,38 @@ export class TransformationCreationDialog extends React.Component<
       </Dialog>
     );
   }
+  getTransformationValueErrorMessage() {
+    if (
+      this.state.transformationOperation &&
+      this.state.transformationFeature
+    ) {
+      const featureMeta =
+        this.context.jointDataset.metaDict[
```
hawestra, line 190: you index into this.context.jointDataset a few times; just want to make sure that this will always be defined

libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationsTable.tsx
```diff
@@ -124,24 +128,37 @@ export class TransformationsTable extends React.Component<
...
+        let value: string | number = transformation.value;
+        if (jointDataset.metaDict) {
+          const featureMeta = jointDataset.metaDict[transformation.feature.key];
+          if (
+            featureMeta.sortedCategoricalValues &&
+            (featureMeta.isCategorical || featureMeta.treatAsCategorical)
+          ) {
+            value = featureMeta.sortedCategoricalValues[value];
+          }
+          const method = `${transformation.feature.text} ${
```
hawestra, line 145: could you comment a sample of how what the final data would look like here

hawestra (approved):

libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationsTable.tsx
romanlutz: Great idea! Adding that.

libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationCreationDialog.tsx
romanlutz: JointDataset isn't optional, but the context may not be defined yet (in the first render before the context is set up. I'm adding `&& this.context` to the `if` right above. LMK if that matches your expectations here.

libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationCreationCategorical.tsx
romanlutz: Makes sense. I've added a short comment.

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
romanlutz: Fair 🤣

libs/localization/src/lib/en.json
romanlutz: We've been calling the overall feature "What If" for a long time, so I just went with that. That has no bearing on how they're rendered, of course. Following the convention here:
image

image

image

I do think What if forecasting may be useful to distinguish it from the "other" what if (the counterfactuals). Let me adjust that for the forecasting ones.

libs/core-ui/src/lib/util/JointDataset.ts
romanlutz: Indeed! Not sure how I missed that 🙂 Thanks for pointing it out.

libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationCreationDialog.tsx
romanlutz: Taking this offline for a separate discussion, as this is something affecting hundreds of places within the codebase.

PR #1923, opened 2023-01-24
Handle state change in `TreatmentView.tsx`

gaugup:
## Description

Now that causal component can update if the global cohort changes, there was a bug in `TreatmentView.tsx` where new treatment policy for a new global cohort wouldn't update the Treatment chart. The fix is to check if the global cohort changed and if it did change the causal TreatView with the new data.

## Checklist

### DIFF
--- a/libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/TreatmentView/TreatmentView.tsx
+++ b/libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/TreatmentView/TreatmentView.tsx
@@ -22,6 +22,7 @@ export interface ITreatmentViewProps {
 }
 export interface ITreatmentViewState {
   selectedPolicy?: ICausalPolicy;
+  selectedIndex: number;
 }
 
 export class TreatmentView extends React.Component<
@@ -31,6 +32,7 @@ export class TreatmentView extends React.Component<
   public constructor(props: ITreatmentViewProps) {
     super(props);
     this.state = {
+      selectedIndex: 0,
       selectedPolicy: props.data?.[0]
     };
   }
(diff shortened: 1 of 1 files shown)


libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/TreatmentView/TreatmentView.tsx
```diff
@@ -71,9 +73,28 @@ export class TreatmentView extends React.Component<
...
+  public componentDidUpdate(prevProps: ITreatmentViewProps): void {
+    if (
+      prevProps.data &&
+      this.props.data &&
+      prevProps.data[0].local_policies &&
+      this.props.data[0].local_policies
+    ) {
+      if (
+        prevProps.data[0].local_policies.length !==
+        this.props.data[0].local_policies.length
```
vinuthakaranth, line 85: Could there be a case where length is same but the policies have been updated?
can we use _.isequal(prevprops, currentprops) in that case?
Why is it always comparing with 0th index in prevProps.data[0]?

libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/TreatmentView/TreatmentView.tsx
```diff
@@ -71,9 +73,28 @@ export class TreatmentView extends React.Component<
     );
   }
 
+  public componentDidUpdate(prevProps: ITreatmentViewProps): void {
+    if (
+      prevProps.data &&
+      this.props.data &&
```
romanlutz, line 79: Shouldn't this check for the length of data being at least 1, too?

PR #1924, opened 2023-01-25
Adding a limit on the number of samples used for computing global explanations

gaugup:
## Description

Adding a limit on the number of samples used for computing global explanations. Following is the table for the run time to compute global explanations:-

| Test data rows  | Global explanation run time (seconds) |
| ------------- | ------------- |
| 5000  | 1.89  |
| 10000  | 3.89  |
| 15000  | 5.92  |
| 20000  | 8.44  |
| 30000  | 14.20  |
| 40000  | 20.49  |

Hence, setting the limit to be 5000 for now to make the global explanations end-point as responsive as possible to the requests. 

## Checklist


romanlutz (approved):

responsibleai/responsibleai/managers/explainer_manager.py
```diff
@@ -141,7 +142,18 @@ def _compute_explanations(self, local: bool, data: Any):
...
+        if len(data) <= MAXIMUM_ROWS_FOR_GLOBAL_EXPLANATIONS:
+            return explainer.explain_global(data, include_local=local)
+        else:
+            warnings.warn((
+                "LARGE-DATA-SCENARIO-DETECTED: "
+                "The data is larger than the supported limit of {0}. "
+                "Computing explanations for first {0} samples only.").format(
+                    MAXIMUM_ROWS_FOR_GLOBAL_EXPLANATIONS),
+                UserWarning)
+            return explainer.explain_global(data[
```
imatiach-msft, line 155: can we choose the points randomly instead of just the first 5k?  Sometimes data is ordered by some feature or label and this may give skewed results for the global explanation.

imatiach-msft (approved):

responsibleai/responsibleai/managers/explainer_manager.py
gaugup: Great idea. I plan to add sampling of data for another requirement and will change the above logic to perform random sampling based of that in a follow up PR.

PR #1927, opened 2023-01-28
Move sort operation of out `render()` in `CausalAggregateView`

gaugup:
## Description

@xuke444 mentioned that data operations such as sorting should not be done in render. Hence, moving the sort of the causal aggregate insights into parent component.

## Checklist

### DIFF
--- a/libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAnalysisView.tsx
+++ b/libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAnalysisView.tsx
@@ -44,7 +44,9 @@ export class CausalAnalysisView extends React.PureComponent<
   public constructor(props: ICausalAnalysisViewProps) {
     super(props);
     this.state = {
-      currentGlobalCausalEffects: this.props.data.global_effects,
+      currentGlobalCausalEffects: this.props.data.global_effects.sort(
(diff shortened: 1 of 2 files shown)


libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAnalysisView.tsx
```diff
@@ -116,10 +118,16 @@ export class CausalAnalysisView extends React.PureComponent<
...
-      this.setState({ currentGlobalCausalEffects: result.global_effects });
+      this.setState({
+        currentGlobalCausalEffects: result.global_effects.sort(
+          (d1, d2) => d2.point - d1.point
+        )
+      });
     } else {
       this.setState({
-        currentGlobalCausalEffects: this.props.data.global_effects
+        currentGlobalCausalEffects: this.props.data.global_effects.sort(
```
vinuthakaranth, line 128: can we have sorting in one function and send the global_effects as arg' while using it as we have sort in multiple places?

vinuthakaranth (approved):

libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAnalysisView.tsx
gaugup: Done

PR #1934, opened 2023-02-01
Deprecate categorical_features argument in favor of feature_metadata.categorical_features, improve metadata validation, and add forecasting-related metadata

romanlutz:
## Description

Deprecating the existing arg with a note to switch to `feature_metadata.categorical_features`. After v0.26 we can remove it entirely. The reason I'm creating this is because the duplicate way of providing categorical features can cause all sorts of trouble if unhandled (which is currently the case).

This PR also improves the validation code in `FeatureMetadata` and adds forecasting-related fields like `time_column_name` and `time_series_id_column_names`.

## Checklist


responsibleai/responsibleai/feature_metadata.py
```diff
@@ -1,57 +1,66 @@
...
         """Placeholder class for feature metadata provided by the user.
 
         :param identity_feature_name: Name of the feature which helps to
                                       uniquely identify a row or instance
                                       in user input dataset.
         :type identity_feature_name: Optional[str]
-        :param datetime_features: List of datetime features in the
-                                  user input dataset.
-        :type datetime_features: Optional[List[str]]
+        :param time_column_name: name of datetime feature in the user input
```
gaugup, line 22: Isn't datetime more common in terms of usage for classification and regression tasks? 

The time_column is very specific to forecasting. I would suggest retaining datetime_features and adding time_column_name instead.

Also we should stick to the existing nomenclature here of using *_feature_name/*_feature_names. Maybe you can use time_feature_name.

responsibleai/responsibleai/feature_metadata.py
```diff
@@ -1,57 +1,66 @@
...
         :param categorical_features: List of categorical features in the
                                      user input dataset.
         :type categorical_features: Optional[List[str]]
         :param dropped_features: List of features that were dropped by the
                                  the user during training of their model.
         :type dropped_features: Optional[List[str]]
+        :param time_series_id_column_names: List of features that are used
+                                            to uniquely identify a time
+                                            series in the user input dataset.
+        :type time_series_id_column_names: Optional[List[str]]
```
gaugup, line 34: This could be time_series_id_feature_names

responsibleai/responsibleai/feature_metadata.py
romanlutz: There may be models that can take datetime columns, but it's not at all clear how we would handle that in the UI as we assume everything is numbers only. All string columns get converted to numbers before getting passed into the UI. This time index column is handled separately for time series which is why that doesn't cause issues, but it's not at all clear how this would work for other tasks. It's also very much out of scope for my workstream. I will make a note to raise this as a discussion point with the broader team shortly. For now, I can make it an optional list, but if anyone specifies more than 1 feature or it's not forecasting then I'll raise an exception during validation saying this isn't supported yet. Wdyt?

I was going for compatible names with AutoML which uses the ones I have in the PR. If we care about doing it differently I can certainly change it. I doubt it'll confuse anyone too much.

responsibleai/responsibleai/feature_metadata.py
romanlutz: the others are all "features" rather than "feature_names" (except for the singular `identity_feature_name`) - shouldn't we stick with the pattern?
`time_series_id_features`?

responsibleai/responsibleai/feature_metadata.py
romanlutz: Made the changes I mentioned above.

romanlutz: Resolving since I made the changes and started the discussion internally.

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -431,11 +493,28 @@ def _validate_rai_insights_input_parameters(
             if feature_metadata is not None:
                 if not isinstance(feature_metadata, FeatureMetadata):
```
gaugup, line 494: This entire branch of validations should be moved under a function in this file or maybe moved into feature_metadata.py.

responsibleai/tests/rai_insights/test_rai_insights_validations.py
```diff
@@ -532,6 +531,150 @@ def test_misidentified_categorical_features(self):
...
+                train=X,
+                test=X,
+                target_column=TARGET,
+                task_type='classification',
+                feature_metadata=feature_metadata)
+
+    @pytest.mark.skip(
+        "Skip forecasting validation test until forecasting is enabled.")
+    def test_feature_metadata_forecasting_multiple_datetime_features(self):
+        X = pd.DataFrame(data=[[1, 1], [2, 3]],
```
gaugup, line 656: This seems to be a repeated code of creating a dataframe. Maybe move it into a function in common_utils.py or create a fixture.

responsibleai/responsibleai/__init__.py
```diff
@@ -7,7 +7,8 @@
 from responsibleai.rai_insights import ModelTask, RAIInsights
 
 from .__version__ import version
+from .feature_metadata import FeatureMetadata
 
 __version__ = version
 
-__all__ = ['ModelAnalysis', 'ModelTask', 'RAIInsights']
+__all__ = ['ModelAnalysis', 'ModelTask', 'RAIInsights', 'FeatureMetadata']
```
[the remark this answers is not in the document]
romanlutz: Because users will need it when they use RAIInsights. Is there a reason not to have it here?

responsibleai/responsibleai/feature_metadata.py
```diff
@@ -1,57 +1,66 @@
...
-                raise UserConfigValidationException(
-                    'The given identity feature name {0} is not present'
-                    ' in user features.'.format(
-                        self.identity_feature_name))
+        identity_feature = ([self.identity_feature_name]
+                            if self.identity_feature_name else None)
+        datetime_features = (self.datetime_features
+                             if self.datetime_features else None)
+        self._validate_columns(
+            'dropped feature', self.dropped_features, feature_names)
```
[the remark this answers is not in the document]
romanlutz: They're used exactly once (here as input for the error messages) and we don't use constants for error messages anywhere in this package so I followed the same pattern.

responsibleai/responsibleai/feature_metadata.py
```diff
@@ -1,57 +1,66 @@
...
+        self._validate_columns(
+            'categorical feature', self.categorical_features, feature_names)
+        self._validate_columns(
+            'identity feature name', identity_feature, feature_names)
+        self._validate_columns(
+            'datetime feature', datetime_features, feature_names)
+        self._validate_columns(
+            'time series ID feature',
+            self.time_series_id_features,
+            feature_names)
```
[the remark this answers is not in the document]
romanlutz: We do, but in RAI insights. We don't know the task type inside this class.

responsibleai/responsibleai/feature_metadata.py
```diff
@@ -82,4 +92,29 @@ def __eq__(self, other_feature_metadata) -> bool:
...
+            other_feature_metadata.time_series_id_features
+
+    @staticmethod
+    def _validate_columns(
+            column_purpose: str,
+            column_names: List[str],
+            feature_names: List[str]):
+        """Ensure the provided column is present in the dataset.
+
+        :param column_purpose: The purpose the column fulfills in the dataset.
```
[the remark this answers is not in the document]
romanlutz: I don't quite follow. Imagine another one called sensitive features. It's the purpose of the feature, not the data type. They could be of any data type.

responsibleai/responsibleai/feature_metadata.py
```diff
@@ -1,57 +1,66 @@
...
-        if self.datetime_features is not None:
-            warnings.warn('datetime_features are not in use currently.')
-        if self.categorical_features is not None:
-            warnings.warn('categorical_features are not in use currently.')
+        self.time_series_id_features = time_series_id_features
 
-    def validate_feature_metadata_with_user_features(
-            self, user_features: Optional[List[str]] = None):
-        """Validate the feature metadata with the user features.
+    def validate(self, feature_names: List[str]):
```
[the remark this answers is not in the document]
romanlutz: Sure. I changed it because there's no real information in that name.

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -431,11 +493,28 @@ def _validate_rai_insights_input_parameters(
...
                     raise UserConfigValidationException(
-                        "Expecting type FeatureMetadata but got {0}".format(
-                            type(feature_metadata)))
+                        "Expecting type FeatureMetadata but got "
+                        f"{type(feature_metadata)}")
+
+                feature_names = list(train.columns)
+                feature_metadata.validate(feature_names)
```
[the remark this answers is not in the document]
romanlutz: It's a matter of readability. I doubt that we need to optimize for variable assignments for a few strings. This is one I had to trace back across this file to figure out, whereas it's clear without having to think further now.

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -431,11 +493,28 @@ def _validate_rai_insights_input_parameters(
...
 
-                feature_metadata.validate_feature_metadata_with_user_features(
-                    list(train.columns))
+                if task_type != ModelTask.FORECASTING:
+                    if feature_metadata.time_series_id_features:
+                        raise UserConfigValidationException(
+                            "The specified metadata time_series_id_features "
+                            "is only supported for the forecasting task type.")
+
+                    if feature_metadata.datetime_features:
```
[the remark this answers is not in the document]
romanlutz: Yes, but it's not in scope for my workstream so I can't support it and write extensive tests. We can absolutely create a new set of items to cover this separate from my workstream. I've started the discussion accordingly internally, feel free to contribute there. I mentioned you in it.

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -431,11 +493,28 @@ def _validate_rai_insights_input_parameters(
...
+                        raise UserConfigValidationException(
+                            "The specified metadata datetime_features "
+                            "is only supported for the forecasting task type.")
+                else:
+                    if (feature_metadata.datetime_features and
+                            len(feature_metadata.datetime_features) > 1):
```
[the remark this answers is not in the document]
romanlutz: See answer above. Not in scope. I'm just using the plural name to avoid having to rename in the future but anything beyond 1 is out of scope.

responsibleai/responsibleai/rai_insights/rai_insights.py
romanlutz: Feature metadata is not award of the task type.

Agreed on moving to separate functions, but that also applies to almost the entire validation method. I suggest doing that in a separate PR to avoid unreadable sets of changes.

responsibleai/responsibleai/feature_metadata.py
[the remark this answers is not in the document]
romanlutz: Anyway, it's easy enough so I've added it.

responsibleai/responsibleai/feature_metadata.py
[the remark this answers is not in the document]
romanlutz: I've added the comment in the docstring saying that it's not meant for tasks other than forecasting. In RAIInsights you'll see the exceptions I've added for this case.

responsibleai/responsibleai/feature_metadata.py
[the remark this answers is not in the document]
romanlutz: Changed it back.

responsibleai/responsibleai/feature_metadata.py
```diff
@@ -82,4 +92,29 @@ def __eq__(self, other_feature_metadata) -> bool:
...
+            column_purpose: str,
+            column_names: List[str],
+            feature_names: List[str]):
+        """Ensure the provided column is present in the dataset.
+
+        :param column_purpose: The purpose the column fulfills in the dataset.
+        :type column_purpose: str
+        :param column_names: List of column names to validate.
+        :type column_names: List[str]
+        :param self.identity_feature_name: List of features in the input
```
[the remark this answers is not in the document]
romanlutz: Nice catch!

responsibleai/tests/rai_insights/test_rai_insights_validations.py
romanlutz: Isn't it way easier to read and understand the test case if we have it right there? Otherwise I have to find the definition of the X variable somewhere else for such a simple data frame that is of no particular significance to this test case or any other. It's 1 line of code regardless.

responsibleai/responsibleai/feature_metadata.py
```diff
@@ -1,57 +1,66 @@
...
-            return
-        if self.identity_feature_name is not None:
-            if self.identity_feature_name not in user_features:
-                raise UserConfigValidationException(
-                    'The given identity feature name {0} is not present'
-                    ' in user features.'.format(
-                        self.identity_feature_name))
+        identity_feature = ([self.identity_feature_name]
+                            if self.identity_feature_name else None)
+        datetime_features = (self.datetime_features
```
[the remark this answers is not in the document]
romanlutz: I believe this is a remnant from having it as a single time column. In that case I needed to convert it to a list to make it compatible with the validation function. So that means you're absolutely right and I will adjust the code 🙂  Nice catch!

responsibleai/responsibleai/rai_insights/rai_insights.py
romanlutz: I've moved this particular block into its own method, but we should probably do more of that in follow-up PRs. Will make a note of that...

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -886,13 +964,15 @@ def _load_metadata(inst, path):
...
-            inst.__dict__['_' + Metadata.FEATURE_METADATA] = None
+            inst.__dict__['_' + Metadata.FEATURE_METADATA] = FeatureMetadata()
         else:
             inst.__dict__['_' + Metadata.FEATURE_METADATA] = FeatureMetadata(
                 identity_feature_name=meta[Metadata.FEATURE_METADATA][
                     'identity_feature_name'],
                 datetime_features=meta[Metadata.FEATURE_METADATA][
                     'datetime_features'],
+                time_series_id_features=meta[Metadata.FEATURE_METADATA][
+                    'time_series_id_features'],
```
imatiach-msft, line 978: I think these strings should all be constants somewhere?

responsibleai/responsibleai/feature_metadata.py
```diff
@@ -82,4 +98,28 @@ def __eq__(self, other_feature_metadata) -> bool:
...
+        :type column_purpose: str
+        :param column_names: List of column names to validate.
+        :type column_names: List[str]
+        :param feature_names: List of features in the input dataset.
+        :type feature_names: List[str]
+        """
+        if column_names is None:
+            return
+        for column_name in column_names:
+            if column_name not in feature_names:
```
imatiach-msft, line 122: nit: maybe make feature_names a set before this loop?  otherwise it's O(n^2), since it's a list, no?  with feature_names as a set it will be O(n) to do this check.  Probably this won't be too long regardless since the number of features here is usually going to be pretty small though.

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -205,6 +209,64 @@ def get_test_data(self, test_data=None):
...
+        if feature_metadata is None:
+            # initialize to avoid having to keep checking if it is None
+            feature_metadata = FeatureMetadata()
+        if feature_metadata.categorical_features is None:
+            if categorical_features is None:
+                consolidated_categorical_features = []
+            else:
+                consolidated_categorical_features = categorical_features
+        else:
+            if categorical_features is None:
```
imatiach-msft, line 250: should we check None OR empty list?  Like user might have passed an empty list for the param, but that probably just means we should ignore it.

imatiach-msft (approved):

responsibleai/responsibleai/__init__.py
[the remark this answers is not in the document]
gaugup: Users don't actually use responsibleai for FeatureMetadata. Rather they use it for RAIInsights.

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -510,6 +566,33 @@ def _validate_rai_insights_input_parameters(
...
+            if task_type != ModelTask.FORECASTING:
+                if feature_metadata.time_series_id_features:
+                    raise UserConfigValidationException(
+                        "The specified metadata time_series_id_features "
+                        "is only supported for the forecasting task type.")
+
+                if feature_metadata.datetime_features:
+                    raise UserConfigValidationException(
+                        "The specified metadata datetime_features "
+                        "is only supported for the forecasting task type.")
```
gaugup, line 589: Does this mean we don't support support datetime features in classification and regression scenario?

tongyu-microsoft (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
romanlutz: Correct. That is exactly as before this PR, we're just explicit about it.

gaugup: Why can't we use datetime_features in classification and regression? It is not the question of supporting now but why do we need this explicit exception?

responsibleai/responsibleai/__init__.py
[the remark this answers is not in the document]
romanlutz: Yes, but `FeatureMetadata` is an input argument for `RAIInsights`. So by not including it here we force them to have multiple imports. Why?

responsibleai/responsibleai/rai_insights/rai_insights.py
romanlutz: Great question! The way I see it there are a few cases involving empty lists: (I'll omit the flipped case)

- `(None, [])` -> go with second one, so no check for empty list needed
- `([], [])` -> neither is None and they match! That's another case where the sets are being compared.
- `([], ['a', ...])` -> neither is None and the set comparison doesn't match, raise exception.

If the user passes one as empty list explicitly and the other with actual categorical feature names, then that should raise an exception and make them aware of the mismatch. At least that's how I was looking at this. It's a super simple fix, too. I'll mark this as resolved, but do lmk if you feel like this isn't right.

responsibleai/responsibleai/rai_insights/rai_insights.py
romanlutz: I don't quite follow. Either you support it, or you don't. If I were to specify a datetime column right now, we will error out. We're simply preserving that behavior for all cases except `task_type="forecasting"` with exactly 1 datetime feature. In all other cases we make users aware that they are configuring a scenario that isn't yet supported.

PR #1945, opened 2023-02-06
Remove unnecessary props from DatasetExplorer `SidePanel` class

gaugup:
## Description

Looks like `cohorts`, `jointDataset` and `dataset` props can be inferred from `this.context`. Hence, removing these from being passed around a props for `SidePanel`.

## Checklist

### DIFF
--- a/libs/dataset-explorer/src/lib/ChartView/DataAnalysisView/SidePanel.tsx
+++ b/libs/dataset-explorer/src/lib/ChartView/DataAnalysisView/SidePanel.tsx
@@ -10,8 +10,6 @@ import {
 } from "@fluentui/react";
 import {
   ColumnCategories,
-  JointDataset,
-  Cohort,
   ChartTypes,
   defaultModelAssessmentContext,
   ModelAssessmentContext,
@@ -23,7 +21,6 @@ import {
   TelemetryLevels,
   TelemetryEventName,
   ifEnableLargeData,
-  IDataset,
   OtherChartTypes
 } from "@responsible-ai/core-ui";
 import { localization } from "@responsible-ai/localization";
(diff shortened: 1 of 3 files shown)


libs/dataset-explorer/src/lib/ChartView/LargeDataView/LargeDatasetExplorerTab.tsx
```diff
@@ -162,13 +162,8 @@ export class LargeDatasetExplorerTab extends React.Component<
             <Stack.Item className={classNames.sidePanel}>
               <SidePanel
                 chartProps={this.state.chartProps}
-                cohorts={this.context.errorCohorts.map(
```
vinuthakaranth, line 165: Please confirm if cohort change, updates bubble chart, thanks.

libs/dataset-explorer/src/lib/ChartView/DataAnalysisView/SidePanel.tsx
```diff
@@ -62,7 +56,7 @@ export class SidePanel extends React.Component<
       text: localization.Interpret.DatasetExplorer.aggregatePlots
     },
     {
-      key: ifEnableLargeData(this.props.dataset)
+      key: ifEnableLargeData(this.context.dataset)
```
romanlutz, line 59: won't the context only get populated at a later point? As in, this will run before `componentDidMount` so `this.context` will be undefined.

gaugup: yes, this is very weird behavior. When SidePanel is initialized from Other places the values are propagated into it via this.context. Not sure why context is not initialized at this point.

libs/dataset-explorer/src/lib/ChartView/LargeDataView/LargeDatasetExplorerTab.tsx
gaugup: Yes, it doesn't work. I think we probably need props to work with render and this.context for the callbacks.

gaugup:
Closing this as the context seems to have an issue when accessed in constructor.

libs/dataset-explorer/src/lib/ChartView/DataAnalysisView/SidePanel.tsx
romanlutz: That's just how the context works. If you want to pass something in you need to use props

PR #1948, opened 2023-02-07
Support forecasting in RAIInsights

romanlutz:
## Description

Notable changes:
- `ml_wrappers` dependency added. The forecasting model wrapper is currently part of this PR, but will move to ml-wrappers shortly. At that point, I'll remove it from here. The dependency is for the base wrapper class.
- `RAIInsights` has fields `predict_output` and `predict_proba_output` as well as the same with `large_` as prefix. For forecasting, I needed something similar, although those particular names don't quite make sense. After exploring some options I set up corresponding fields `forecast_output` and `forecast_quantiles_output` (and the `large_` ones for them). I couldn't find any usage of these outside of `RAIInsights`, so I prefixed everything with an underscore. I need confirmation that that's acceptable, though. The way their contents are calculated has also been streamlined to avoid 8 sets of `model.<method>(...)` after another.
- `self._initialize_managers()` needed to be moved out of the base RAIInsights constructor because of an interdependency between the RAIInsights constructor and the validation method. The manager initialization requires a lot of the fields to be set, while the validation needs to occur for that to happen. On the other hand, the base constructor is the one that sets many of those fields, in particular the model. Since we have to wrap the model for forecasting this becomes impossible otherwise. It's now just at the end of the RAIInsights constructor.
- `predict_proba` for classification is optional in this PR. I can keep it as mandatory since that's what we had so far, but that's technically excluding lots of perfectly fine classification models.
- even after merging, this feature won't really be available yet. There's a feature flag added called `forecasting_enabled`.

About to be added:
- more test cases in `responsibleai/tests`
- I will have a separate PR with a notebook and e2e tests on that notebook

## Checklist


responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -44,6 +49,93 @@
...
+MODEL_METHODS = {
+    ModelTask.CLASSIFICATION: [
+        ModelMethod(
+            name=SKLearn.PREDICT,
+            optional=False,
+            purpose=MethodPurpose.PREDICTION),
+        ModelMethod(
+            name=SKLearn.PREDICT_PROBA,
+            # TODO: so far assumption was that it's required
+            # but that excludes lots of classification models.
```
romanlutz, line 118: We need to clarify this, although perhaps out of scope for this PR/workstream.

responsibleai/responsibleai/_internal/_forecasting_wrappers.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def _get_model_package(model):
+    module = inspect.getmodule(model)
+    module_name = module.__name__.partition('.')[0]
+    model_package = sys.modules[module_name].__package__
+    return model_package
+
+
+def _wrap_model(model, examples, time_feature, time_series_id_features):
```
imatiach-msft, line 27: this definitely looks like something that should go in the ml-wrappers repository eventually

responsibleai/responsibleai/_internal/_forecasting_wrappers.py
```diff
@@ -0,0 +1,141 @@
...
+        self._time_feature = time_feature
+        for ts_id in time_series_id_features:
+            if ts_id not in self._examples.columns:
+                raise ValueError(
+                    f"The passed time series ID feature {ts_id} is not "
+                    "in the examples DataFrame.")
+        self._time_feature = time_feature
+        self._time_series_id_features = time_series_id_features
+
+    def forecast(self, X):
```
imatiach-msft, line 75: is this the API we want to go with instead of predict?

imatiach-msft: perhaps we can also have both, with predict an alias for forecast?

responsibleai/responsibleai/_internal/_forecasting_wrappers.py
```diff
@@ -0,0 +1,141 @@
...
+            hasattr(model, _Forecasting.FORECAST_QUANTILES)):
+        return _WrappedQuantileForecastingModel(
+            model, examples, time_feature, time_series_id_features)
+    else:
+        return _WrappedForecastingModel(
+            model, examples, time_feature, time_series_id_features)
+
+
+class _WrappedForecastingModel(BaseWrappedModel):
+    def __init__(self, model, examples, time_feature,
```
imatiach-msft, line 46: I think all of these methods should have docs on them eventually.. I guess maybe it's fine for now since this is private and probably temporary anyway before it's moved to ml-wrappers but the docs should be added at some point

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -63,20 +155,21 @@ def __init__(self,
                  classes: Optional[np.ndarray] = None,
                  serializer: Optional[Any] = None,
                  maximum_rows_for_test: int = 5000,
-                 feature_metadata: Optional[FeatureMetadata] = None):
+                 feature_metadata: Optional[FeatureMetadata] = None,
```
imatiach-msft, line 158: There are a lot of changes here - I wonder if it makes more sense at this point to have an RAIForecastingInsights, similar to how we have RAITextInsights and RAIVisionInsights classes for text and vision scenarios?  Both the current code and RAIForecastingInsights work, but since forecasting is quite different I really wonder if it makes more sense to just separate it out and have the parameters and serialization logic be more focused on that scenario, especially with all of these kwargs.

imatiach-msft (approved):
Posting some general comments

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -63,20 +155,21 @@ def __init__(self,
...
+                 **kwargs):
```
imatiach-msft, line 159: it's better to usually list out the parameters if possible... I think with a separate class we won't need this kwargs, is that correct?

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -1002,55 +1279,33 @@ def _load_metadata(inst, path):
...
-            with open(large_predict_proba_output_path, 'r') as file:
-                large_predict_proba_output = json.load(file)
-            inst.__dict__["_large_predict_proba_output"] = np.array(
-                large_predict_proba_output)
-        else:
-            inst.__dict__["_large_predict_proba_output"] = None
+        for data_name, file_name in _OUTPUT_FIELDS_AND_FILENAMES:
+            print("meh")
+            file_path = prediction_output_path / file_name
+            print(file_path)
```
imatiach-msft, line 1301: remove prints

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -1002,55 +1279,33 @@ def _load_metadata(inst, path):
...
-            SerializationAttributes.LARGE_PREDICT_PROBA_JSON)
-        if large_predict_proba_output_path.exists():
-            with open(large_predict_proba_output_path, 'r') as file:
-                large_predict_proba_output = json.load(file)
-            inst.__dict__["_large_predict_proba_output"] = np.array(
-                large_predict_proba_output)
-        else:
-            inst.__dict__["_large_predict_proba_output"] = None
+        for data_name, file_name in _OUTPUT_FIELDS_AND_FILENAMES:
+            print("meh")
```
imatiach-msft, line 1299: remove "meh" print here

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -1002,55 +1279,33 @@ def _load_metadata(inst, path):
...
+            file_path = prediction_output_path / file_name
+            print(file_path)
+            print(file_path.exists())
+            if file_path.exists():
+                with open(file_path, 'r') as file:
+                    inst.__dict__[data_name] = np.array(json.load(file))
+            else:
+                inst.__dict__[data_name] = None
+        print("wait what?")
```
imatiach-msft, line 1308: remove "wait what?" print here

responsibleai/responsibleai/_internal/_forecasting_wrappers.py
romanlutz: I had a set of discussions about this with the forecasting team. They strongly recommend using forecast rather than predict. That said, you can easily wrap your model with predict function inside the wrapper. Does that make sense to you?

I'm a bit reluctant to have one as an alias for the other. A model might have both functions, and we can hardly ask users to remove one of them if they're not identical. Or they might use yet another method. Or worse, they have these and one works differently from the other. In summary, it opens up a few additional failure cases without really adding anything. I'd rather have people consciously choosing to provide forecast (with a wrapper if need be). I'll provide an example notebook so it's clear.

responsibleai/responsibleai/_internal/_forecasting_wrappers.py
romanlutz: I agree 100%
I'll add docs in this PR.

responsibleai/responsibleai/rai_insights/rai_insights.py
romanlutz: I had it separated out initially, it's 90% the same if not more. We'd have A LOT of duplication since all the tabular data code is identical. Making it inherit from RAIInsights is also not viable. With some refactoring we could make this much nicer, though.

responsibleai/responsibleai/rai_insights/rai_insights.py
romanlutz: No. This is not at all about parameters that are for any functionality in RAIInsights, but only for the feature flag. I do not want to show it to people. If they know about it they can use it at their own risk, but otherwise they shouldn't assume that forecasting is already supported. The same mechanism can work for other feature flags in the future.

I agree on not hiding kwargs that we want people to use.

responsibleai/responsibleai/rai_insights/rai_insights.py
romanlutz: Oh this is a funny one 🤣 Thank you

responsibleai/responsibleai/_internal/_forecasting_wrappers.py
romanlutz: Yes! I'll open a PR there. Once it's part of a ml-wrappers release I'll just point there and delete this file. I made sure to use all leading underscores here so that nobody starts relying on it.

responsibleai/responsibleai/_internal/_forecasting_wrappers.py
romanlutz: Done.

PR #1950, opened 2023-02-07
fix mean absolute error metric calculation in model overview panel

imatiach-msft:
## Description

During customer session, I noticed the mean absolute error metric looked incorrect for their data.  Indeed, it seems we were not taking the "mean" of the data but just summing the error.  To fix the issue, I divided by the count to get the correct metric.

Before:
image

After:
image

Note how mean absolute error and mean squared error look a lot more similar now.

## Checklist

### DIFF
--- a/libs/core-ui/src/lib/util/StatisticsUtils.ts
+++ b/libs/core-ui/src/lib/util/StatisticsUtils.ts
@@ -113,9 +113,10 @@ const generateRegressionStats: (
   errors: number[]
 ): ILabeledStatistic[] => {
   const count = trueYs.length;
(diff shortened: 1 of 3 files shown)


gaugup (approved):

romanlutz (approved):
Nice catch! Also, another good reason why we shouldn't reinvent the wheel on such things and just reuse existing definitions. This seems like a pretty bad bug. Do you have plans to ship it in Studio soon?

romanlutz:
I guess we should also consider adding at least a few basic tests if we intend to keep these definitions around.

PR #1955, opened 2023-02-08
[Large data] Map classification outcome to string values, dither, count changes

vinuthakaranth:
## Description

This PR contains changes for below:

1. Map classification outcome to string values (false positive, false negative, true positive & true negative) for data explorer, feature importance, counterfactuals.
2. Disable dither for large data as it is handled in backend
3. Remove count large data explorer individual points as size is already present as part of cluster
4. Refactor code by creating interface IClusterData which has all the details of the cluster selected.

bubble plot retaining existing experience:
image

scatter plot with mapped values:
image

dither changes:
image

remove count:
image

bubble

## Checklist


libs/core-ui/src/lib/Highchart/bubbleChart/buildScatterTemplate.ts
```diff
@@ -44,3 +48,21 @@ export function buildScatterTemplate(
   customData.Index = index;
   return customData;
 }
+
+function getAxisValueMapping(
+  axisProperty: string,
+  axisValue: number,
+  axisMap?: { [key: number]: string }
+): string | number {
+  if (axisMap && axisProperty === JointDataset.ClassificationError) {
```
gaugup, line 57: I thought this should apply to all categorical values not just classification error.

libs/core-ui/src/lib/Highchart/bubbleChart/buildScatterTemplate.ts
```diff
@@ -44,3 +48,21 @@ export function buildScatterTemplate(
...
+    const nearestValue = getNearestValue(axisValue, Object.keys(axisMap));
+    return axisMap[nearestValue];
+  }
+  return axisValue;
+}
+
+function getNearestValue(axisValue: number, axisMapKeys: any[]): number {
```
gaugup, line 64: Why do we need to get the nearest value. Can't we use the `x`/`y` to get the center and map that value?

vinuthakaranth: Scatterplot data is independent of bubble plot data at this time. As we don't persist bubble plot data, it is not possible to get x and y of clusters at this moment in scatter plot. There is a item to persist bubble plot data when axis doesn't change. This can be done as a follow up task of that item.

libs/core-ui/src/lib/Highchart/bubbleChart/buildScatterTemplate.ts
vinuthakaranth: Thanks for catching this @gaugup! Updated to add categorical features.

libs/core-ui/src/lib/Highchart/bubbleChart/buildScatterTemplate.ts
vinuthakaranth: Thanks Gaurav, updated to use x and y from cluster.

libs/core-ui/src/lib/Highchart/bubbleChart/buildScatterTemplate.ts
```diff
@@ -16,21 +16,45 @@ export interface ICustomData {
 export function buildScatterTemplate(
```
gaugup, line 16: could you maybe rename the function buildScatterTemplate? There are other functions in the repo with same name.

gaugup (approved):

libs/core-ui/src/lib/Highchart/bubbleChart/buildScatterTemplate.ts
vinuthakaranth: Thanks Gaurav, will update the name in the coming PR.

tongyu-microsoft (approved):

PR #1956, opened 2023-02-08
Migrate responsibleai tests to use utilities in rai_test_utils - Part 2

gaugup:
## Description

Update `rai_test_utils` to 0.1.0 in `erroranalysis` and `responsibleai`.

## Checklist

### DIFF
--- a/responsibleai/tests/causal/conftest.py
+++ b/responsibleai/tests/causal/conftest.py
@@ -8,7 +8,9 @@ import pandas as pd
 import pytest
 import shap
 from sklearn.model_selection import train_test_split
-from tests.common_utils import create_adult_income_dataset, create_housing_data
+from tests.common_utils import create_adult_income_dataset
+
+from rai_test_utils.datasets.tabular import create_housing_data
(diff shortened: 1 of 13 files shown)


responsibleai/tests/causal/conftest.py
```diff
@@ -8,7 +8,9 @@
 import pytest
 import shap
 from sklearn.model_selection import train_test_split
-from tests.common_utils import create_adult_income_dataset, create_housing_data
+from tests.common_utils import create_adult_income_dataset
```
vinuthakaranth, line 11: curious we still don't have create_adult_income_dataset in rai_test_utils?

vinuthakaranth (approved):

responsibleai/tests/causal/conftest.py
gaugup: Because we would need to depend on dice-ml in rai_test_utils. rai_test_utils is a simple package and I don't want complex dependencies to be present in it.

PR #1957, opened 2023-02-09
[Large counterfactuals & explanations] Do not call local counterfactuals/explanations on unselecting the point

vinuthakaranth:

Do not call local counterfactuals on unselecting the point

## Description
Fix is to call local counterfactuals/explanations api only when there is selection index.

## Checklist

### DIFF
--- a/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/IndividualFeatureImportanceView/LargeIndividualFeatureImportanceView/LargeIndividualFeatureImportanceView.tsx
+++ b/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/IndividualFeatureImportanceView/LargeIndividualFeatureImportanceView/LargeIndividualFeatureImportanceView.tsx
@@ -249,7 +249,7 @@ export class LargeIndividualFeatureImportanceView extends React.Component<
     );
   };
 
-  private toggleSelectionOfPoint = (index?: number): number[] | undefined => {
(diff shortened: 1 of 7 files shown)


libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/IndividualFeatureImportanceView/LargeIndividualFeatureImportanceView/LargeIndividualFeatureImportanceView.tsx
```diff
@@ -259,6 +259,7 @@ export class LargeIndividualFeatureImportanceView extends React.Component<
         selectedPointsIndexes: newSelections
       });
     }
+    return newSelections;
```
hawestra, line 262: nit: could we return [] instead of undefined so that we match the other function declarations of `toggleSelectionOfPoint = (index?: number): number[]`

hawestra (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/IndividualFeatureImportanceView/LargeIndividualFeatureImportanceView/LargeIndividualFeatureImportanceView.tsx
vinuthakaranth: Nice catch! removed undefined

PR #1958, opened 2023-02-09
Fix selection rate

romanlutz:
## Description

Selection rate is all predicted 1s, so FP + TP, not FN + TP.

## Checklist

### DIFF
--- a/libs/core-ui/src/lib/util/StatisticsUtils.ts
+++ b/libs/core-ui/src/lib/util/StatisticsUtils.ts
@@ -98,7 +98,7 @@ const generateBinaryStats: (outcomes: number[]) => ILabeledStatistic[] = (
     {
       key: BinaryClassificationMetrics.SelectionRate,
       label: localization.Interpret.Statistics.selectionRate,
-      stat: (falseNegCount + truePosCount) / total
+      stat: (falsePosCount + truePosCount) / total
     }
   ];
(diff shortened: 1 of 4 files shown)


imatiach-msft (approved):

libs/core-ui/src/lib/util/StatisticsUtils.ts
```diff
@@ -98,7 +98,7 @@ const generateBinaryStats: (outcomes: number[]) => ILabeledStatistic[] = (
     {
       key: BinaryClassificationMetrics.SelectionRate,
       label: localization.Interpret.Statistics.selectionRate,
-      stat: (falseNegCount + truePosCount) / total
+      stat: (falsePosCount + truePosCount) / total
```
imatiach-msft, line 101: So basically, all that were PREDICTED positive over total, not that were "in ground truth" positive (which would have been false negative instead of false positive then) - these are the "selected ones".  That makes sense to me.  Some of the "selected ones" were just incorrectly selected.

xuke444 (approved):

libs/core-ui/src/lib/util/StatisticsUtils.ts
romanlutz: Correct! It has nothing to do with ground truth.

PR #1959, opened 2023-02-10
Add spline chart data fetch from SDK backend

gaugup:
## Description

Add spline chart data fetch from SDK backend

## Checklist


libs/model-assessment/src/lib/ModelAssessmentDashboard/ModelAssessmentDashboardProps.ts
```diff
@@ -133,6 +133,10 @@ export interface IModelAssessmentDashboardProps
     request: any[],
     abortSignal: AbortSignal
   ) => Promise<any[]>;
+  requestSplinePlotDistribution?: (
+    request: any,
+    abortSignal: AbortSignal
+  ) => Promise<any>;
```
vinuthakaranth, line 139: Can we update this any with interface of spline data from api?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionSplineChart.tsx
```diff
@@ -27,14 +34,86 @@ export class ProbabilityDistributionSplineChart extends React.Component<
...
+    prevProps: IProbabilityDistributionSplineChartProps
+  ): void {
+    this.calculateSplineDataIfNeeded(prevProps);
+  }
+
+  public async calculateSplineDataIfNeeded(
+    prevProps?: IProbabilityDistributionSplineChartProps
+  ): Promise<void> {
+    if (
+      prevProps === undefined ||
```
vinuthakaranth, line 59: why do we need this?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionSplineChart.tsx
```diff
@@ -27,14 +34,86 @@ export class ProbabilityDistributionSplineChart extends React.Component<
...
+      !_.isEqual(prevProps.selectedCohorts, this.props.selectedCohorts) ||
+      !_.isEqual(
+        prevProps.probabilityOption?.id,
+        this.props.probabilityOption?.id
+      )
+    ) {
+      let splinePlotData = undefined;
+      if (
+        this.context &&
```
vinuthakaranth, line 68: can we remove this?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionSplineChart.tsx
```diff
@@ -1,23 +1,30 @@
...
 
 interface IProbabilityDistributionSplineChartProps {
   selectedCohorts: ErrorCohort[];
   probabilityOption?: IChoiceGroupOption;
 }
 
-class IProbabilityDistributionSplineChartState {}
+interface IProbabilityDistributionSplineChartState {
+  probabilityBinCountArray?: Array<IProbabilityBinCount[] | undefined>;
+  selectedCohorts: ErrorCohort[];
```
vinuthakaranth, line 26: Why do we need this as state? I see updated value is send in props and both times we set directly from props in state. Can we still have this just as prop and use in render?

gaugup: Yes, we need it in the state since the number of entries in probabilityBinCountArray need to equal to selectedCohorts. If we use selectedCohorts from props then there will be mismatch in the number of entries and I see a crash.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionSplineChart.tsx
gaugup: Removed

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ProbabilityDistributionSplineChart.tsx
gaugup: The prevProps can be undefined. The ProbabilityDistributionBoxChart.tsx follow a similar condition style.

libs/model-assessment/src/lib/ModelAssessmentDashboard/ModelAssessmentDashboardProps.ts
gaugup: Somehow adding the type here is causing the type to fail. Will leave this as any for now. Looks like most APIs have any as the return type anyways.

xuke444 (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/ModelAssessmentDashboardProps.ts
vinuthakaranth: I am still not okay with any here if it can be removed with interface similar to backend and solving following errors.

vinuthakaranth (approved):

PR #1960, opened 2023-02-10
add missing true and predicted labels to axis config for dataset explorer in multilabel text and vision scenarios

imatiach-msft:
## Description

While testing multilabel text and vision, I found that the axis config dialog in the dataset explorer was missing the true and predicted labels as options.  This PR adds those options.  Since there can be many labels, I configured this as a dropdown similar to the probabilities config dropdown in a multiclass scenario.
Note as part of this PR I also refactored the AxisConfigDialog as it was quite large and with the new changes the file was going over 300 lines in length.

Screenshot of updated axis config dialog with dropdown for true and predicted labels:

image

## Checklist

### DIFF
--- a/libs/core-ui/src/lib/components/AxisConfigDialog.tsx
+++ b/libs/core-ui/src/lib/components/AxisConfigDialog.tsx
@@ -12,7 +12,6 @@ import {
   DefaultButton
 } from "@fluentui/react";
 import { localization } from "@responsible-ai/localization";
-import _ from "lodash";
 import React from "react";
 
 import { cohortKey } from "../cohortKey";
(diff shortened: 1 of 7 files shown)


libs/core-ui/src/lib/components/AxisConfigDialog.tsx
```diff
@@ -157,24 +116,19 @@ export class AxisConfigDialog extends React.PureComponent<
...
-                    selectedKey={this.state.selectedColumn.property}
-                  />
-                )}
+                {isDataColumn &&
+                  this.renderComboBox(
+                    this.state.dataArray,
+                    localization.Interpret.AxisConfigDialog.selectFeature
+                  )}
+                {isProbabilityColumn &&
+                  this.renderClassComboBox(this.state.classArray)}
```
tongyu-microsoft, line 125: are we missing to pass label={localization.Interpret.AxisConfigDialog.selectClass} to `this.renderComboBox`? Or is this on purpose?

tongyu-microsoft (approved):

PR #1962, opened 2023-02-10
improve installation instructions for onboarding to RAI toolbox repository

imatiach-msft:
## Description

Improve installation instructions for onboarding to RAI toolbox repository.
Specifically, mention that only node v16.x is tested, as I have seen some developers try to onboard with v18.x and it failed.
Also, add instructions about some missing new pypi packages (raiutils, erroranalysis, etc) that are required for running locally.
Changed order of raiwidgets and responsibleai install since otherwise it can cause the version on pypi to get installed instead.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,6 +53,9 @@ All pull requests need to abide by the following criteria to be accepted:
 First ensure you have
 [npm](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
 installed (which in turn may require installing `node`).
+
+This repository is tested with node 14.x, 15.x and 16.x, so using the latest 16.x version is recommended.
+
 Using npm you can install `yarn` as follows:
 
 ```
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -166,14 +182,17 @@ For any new change, which involves changing any of the python SDK components and
...
-7. Run `pip install -e raiwidgets` to install raiwidgets locally.
+6. Run `pip install -e raiwidgets` to install raiwidgets locally.
+7. Run `pip install -e responsibleai` to install responsibleai locally.
 8. Run `pip install -e erroranalysis` to install erroranalysis locally.
 9. Run `pip install -e rai_core_flask` to install rai_core_flask locally.
-10. Install `jupyter` using `pip install jupyter`
-11. Open any notebook using python SDK and any widget from `responsible-ai-toolbox` and test your changes.
+10. Run `pip install -e raiutils` to install raiutils locally.
+11. Run `pip install -e nlp_feature_extractors` to install nlp_feature_extractors locally if using the RAI Text Dashboard.
+12. Run `pip install -e rai_test_utils` to install rai_test_utils locally.
```
gaugup, line 191: We don't need ra_test_utils to work with ui. Do we need to add this line?

CONTRIBUTING.md
```diff
@@ -166,14 +182,17 @@ For any new change, which involves changing any of the python SDK components and
 3. You should commit all your current set of changes for SDK and UX using `git commit`.
 4. Clean all untracked files using `git clean -fdx`
 5. Run `yarn install` and `yarn buildall` to build the UX changes.
-6. Run `pip install -e responsibleai` to install responsibleai locally.
-7. Run `pip install -e raiwidgets` to install raiwidgets locally.
+6. Run `pip install -e raiwidgets` to install raiwidgets locally.
+7. Run `pip install -e responsibleai` to install responsibleai locally.
```
gaugup, line 186: Does the order installing raiwidgets/responsibleai matter?

tongyu-microsoft (approved):

CONTRIBUTING.md
imatiach-msft: Yes, it actually does!  You need to install raiwidgets first.  Otherwise, it will overwrite the dependencies.

PR #1963, opened 2023-02-10
Fix robot page exception on data explorer

tongyu-microsoft:
There was an ICM for robot page exception on data explorer. We have seen similar issue before on undefined this.props.cohorts[this.props.selectedCohortIndex].

## Checklist


gaugup (commented):
Are there other places in the code that we will run into this error? Why not fix this in all places?

tongyu-microsoft:
> this.props.cohorts[this.props.selectedCohortIndex]

this is the only place, we have handled all scenarios for this specific issue on this.props.cohorts[this.props.selectedCohortIndex]
image

hawestra (approved):

PR #1964, opened 2023-02-11
Fix robot page exception on undefined dataset

tongyu-microsoft:
This PR is to fix the robot page exception that customer hit in `buildModelMetadata` function.

For better user experience, we should find a way to check all the props, and return meaningful error messages when certain property is undefined. Maybe we can achieve that directly inside AML studio UI code.

## Checklist


imatiach-msft (approved):

tongyu-microsoft:
This can be fixed in AzureML repo. Close this PR.

PR #1971, opened 2023-02-15
Fix issue - Filter column is localized when calling backend 

RubyZ10:
## Description

Before fix 
For filter column we are sending the localized string to backend
image

After fix
image

## Checklist

### DIFF
--- a/libs/core-ui/src/lib/util/CohortUtils.ts
+++ b/libs/core-ui/src/lib/util/CohortUtils.ts
@@ -1,0 +1,29 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { JointDataset } from "./JointDataset";
+
+enum ColumnNames {
+  Index = "Index",
+  PredictedY = "Predicted Y",
+  TrueY = "True Y",
+  ClassificationError = "Classification outcome",
+  RegressionError = "Regression error"
+}
+
+export function getColumnName(filterColumn: string): string {
+  switch (filterColumn) {
+    case JointDataset.IndexLabel:
+      return ColumnNames.Index;
+    case JointDataset.PredictedYLabel:
+      return ColumnNames.PredictedY;
+    case JointDataset.TrueYLabel:
+      return ColumnNames.TrueY;
+    case JointDataset.ClassificationError:
+      return ColumnNames.ClassificationError;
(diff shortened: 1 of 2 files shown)


imatiach-msft (approved):

libs/core-ui/src/lib/util/CohortUtils.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { JointDataset } from "./JointDataset";
+
+enum ColumnNames {
+  Index = "Index",
```
imatiach-msft, line 7: Should these come from the english json file directly?  Instead of redefining here we can just use the existing english definitions (so it's easier to update in case we decide to change the strings and also to limit duplication).

RubyZ10: makes sense, how to use the english json file directly?
this is a temporary fix, was planning to change the filter logic with the removing jointdataset work, but I am happy to try other ways now

vinuthakaranth: umm I do not recommend using from json file. Haven't seen en-us file being used other than for localization.

RubyZ10: yeah, agree on having single source of truth, but not sure if we can use directly from json file

vinuthakaranth: Can we reuse CohortColumnNames from ProcessPreBuiltCohort.ts file?

RubyZ10: Or we can fix Jointdataset itself, but I don't want to touch too much on that, as it is going to be gone. So made this temporary fix

RubyZ10: > Can we reuse CohortColumnNames from ProcessPreBuiltCohort.ts file?

RegressionError is different, we are passing "Regression error" as filter column to backend instead of "Error"(CohortColumnNames.RegressionError)

imatiach-msft: hmm ok then you can just ignore my comment!  Seems like this is a temporary fix anyway.

PR #1978, opened 2023-02-22
Forecasting: improve description in no-features case

romanlutz:
## Description

This PR simply adds more description in the no-features case to let users know that they could use what-if analysis if they provided features.

## Checklist

### DIFF
--- a/libs/localization/src/lib/en.json
+++ b/libs/localization/src/lib/en.json
@@ -1850,7 +1850,7 @@
     "forecastHeader": "Forecast analysis",
     "whatIfForecastingDescription": "What-if allows you to perturb features for your entire time series and observe how the model's forecast changes.",
     "whatIfForecastingChooseTimeSeries": "To start, choose a time series from the options below.",
-    "forecastDescription": "Forecast analysis compares your model's forecast to the actual values of your time series. To enable what-if analysis, please provide a dataset with features.",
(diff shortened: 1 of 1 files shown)


libs/localization/src/lib/en.json
```diff
@@ -1850,7 +1850,7 @@
     "forecastHeader": "Forecast analysis",
     "whatIfForecastingDescription": "What-if allows you to perturb features for your entire time series and observe how the model's forecast changes.",
     "whatIfForecastingChooseTimeSeries": "To start, choose a time series from the options below.",
-    "forecastDescription": "Forecast analysis compares your model's forecast to the actual values of your time series.",
+    "forecastDescription": "Forecast analysis compares your model's forecast to the actual values of your time series. To enable what-if analysis, please provide a dataset with features.",
```
vinuthakaranth, line 1853: v.nit: it is recommended to not use please in user facing statements. I know we have this in various places :(

vinuthakaranth (approved):

libs/localization/src/lib/en.json
romanlutz, line 1853:
```suggestion
    "forecastDescription": "Forecast analysis compares your model's forecast to the actual values of your time series. To enable what-if analysis, provide a dataset with features.",
```

gaugup (approved):

PR #1981, opened 2023-02-22
Forecasting: restrict transformation names to 50 chars

romanlutz:
## Description

Allowing arbitrarily long transformation names can cause all sorts of issues. To prevent that, this PR sets the max length to 50 characters. We're also no longer showing the error message from the start but rather upon the first change that makes the name invalid.

When opening a new dialog:
image

After entering a name and erasing it:
image

When there's a name collision:
image

When the char limit of 50 is exceeded:
image

## Checklist

### DIFF
--- a/libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationCreationDialog.tsx
+++ b/libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationCreationDialog.tsx
@@ -155,8 +155,8 @@ export class TransformationCreationDialog extends React.Component<
             disabled={
               transformationCombinationErrorMessage !== undefined ||
               this.getTransformationNameErrorMessage(
-                this.state.transformationName ?? ""
-              ) !== "" ||
+                this.state.transformationName
+              ) !== undefined ||
               transformationValueErrorMessage !== undefined ||
               this.state.transformationOperation === undefined ||
               this.state.transformationFeature === undefined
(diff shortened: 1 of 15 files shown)


gaugup (approved):

libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationCreationDialog.tsx
```diff
@@ -166,7 +154,9 @@ export class TransformationCreationDialog extends React.Component<
           <PrimaryButton
             disabled={
               transformationCombinationErrorMessage !== undefined ||
-              transformationNameErrorMessage !== undefined ||
+              this.getTransformationNameErrorMessage(
```
hawestra, line 157: nit: could getTransformationNameErrorMessage accept undefined or string and then we can move this logic into that function ?

libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationCreationDialog.tsx
```diff
@@ -216,11 +206,32 @@ export class TransformationCreationDialog extends React.Component<
...
+        .scenarioNamingCollisionMessage;
+    }
+    if (value.length > transformationNameMaxLength) {
+      return localization.formatString(
+        localization.Forecasting.TransformationCreation
+          .scenarioNamingLengthMessage,
+        value.length
+      );
+    }
+    return "";
```
hawestra, line 225: nit: return undefined as the no error scenario to match the other functions

hawestra (approved):

imatiach-msft (approved):

libs/forecasting/src/lib/ForecastingDashboard/Controls/TransformationCreationDialog.tsx
romanlutz: Good idea! I had it accept string only because it's what TextField's onGetErrorMessage expects, but the broader definition doesn't make that one fail so it works.

PR #1986, opened 2023-02-24
Forecasting: validate characters for name field in transformation creation dialog

romanlutz:
## Description

This PR adds input validation for the name field in the transformation creation dialog. Length validation already exists, so this only adds character validation.

If you write a name and fully erase it:
image

If you enter invalid characters:
image

## Checklist

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -13,6 +13,7 @@ module.exports = {
     "<rootDir>/apps/dashboard",
     "<rootDir>/libs/localization",
     "<rootDir>/libs/error-analysis",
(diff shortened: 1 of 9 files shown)


hawestra:
should there be a string length check ? or is this already existing ?

hawestra (approved):

romanlutz:
> should there be a string length check ? or is this already existing ?

That was added in #1981 🙂 It's an earlier check so it doesn't need to be part of the function with the regex.

imatiach-msft (approved):

PR #1991, opened 2023-03-01
fix error analysis showing float values with many zeros for some classification task nodes

imatiach-msft:
## Description

I noticed in several customer meetings that sometimes the nodes in error analysis would have values with many zeros.  I was actually able to reproduce this locally for one of our notebooks ([census classification scenario](https://github.com/microsoft/responsible-ai-toolbox/blob/main/notebooks/individual-dashboards/erroranalysis-dashboard/erroranalysis-interpretability-dashboard-census.ipynb)).  It seems this occurs only for classification tasks, as we don't use precision in front-end then, and we cast the error variable passed from backend to frontend cast as float.  The fix is to conditionally cast the error variable to frontend or backend depending on whether a regression metric was used.

Before fix:

image

After fix:

image

## Checklist


gaugup (commented):
Can you add a test for regression as well?

tongyu-microsoft (approved):

RubyZ10 (approved):

imatiach-msft:
> Can you add a test for regression as well?

The validation code is for both regression and classification tasks

PR #1992, opened 2023-03-02
[Refactor] 1. Cohort and filter change 2. Apply on counterfactual chart

RubyZ10:
Major changes in this PR
1. Added DatasetCohort 
When creating a new cohort from UI, according to the filters customer selected, we will create a new DatasetCohort and save the selectedIndexes[] and cohort metrics there. 
With the selectedIndexes[] and dataset, we should be able to get everything we need for each cohort
2. Added columnRanges
It calculate the range for each feature and the other cohort filter column. 
This is a temporary solution, calculate range does not make sense for big data. it should come from backend.
3. Added datasetCohort, modelType, datasetFeatureRanges in ModelAssessmentContext, as those should be global and need to be consumed by other components. 
**We will have both datasetCohort and errorCohort available in context, so when creating, updating, deleting cohort, we will make these actions on both datasetCohort and errorCohort.
When read and leverage cohort, it will depends on the flight. When flight is on, read and use datasetCohort, when flight is off, read and use errorCohort**
5. Filter
Changed filter column to use the "real" column name, exactly as what backend expected, so when calling SDK, no translate needed
Before change,
image
After change:
image
6. Added getPropertyValues
With this function, it return property values from dataset according to indexes
7. Leveraged DatasetCohort in CounterfactualChart.tsx
8. Moved FeatureFlights.ts from model-assessment package to core-ui

## Checklist


libs/core-ui/src/lib/Cohort/CohortBasedComponent.ts
```diff
@@ -21,6 +23,9 @@ export interface ICohortBasedComponentState {
   cohorts: ErrorCohort[];
   selectedCohort: ErrorCohort;
   jointDataset: JointDataset;
+  datasetCohorts?: DatasetCohort[];
```
vinuthakaranth, line 26: what is baseDatasetCohort? Can't we get this from array datasetCohorts?

libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortBar.tsx
```diff
@@ -127,11 +128,11 @@ export class CohortBar extends React.Component<
     });
   };
 
-  private onCohortChange = (newCohort: Cohort): void => {
+  private onCohortChange = (newCohort: Cohort | DatasetCohort): void => {
```
vinuthakaranth, line 131: will remove Cohort eventually?

libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortEditorFilterSection.tsx
```diff
@@ -230,7 +251,9 @@ export class CohortEditorFilterSection extends React.PureComponent<
   };
 
   private setDefaultStateForKey(key: string): void {
-    const filter = this.getFilterValue(key);
+    const filter = this.props.isRemoveJointDatasetFlightOn
+      ? this.getFilterValue2(key)
```
hawestra, line 255: because you're going to remove the old version (ie this.getFilterValue(key)) when the flight gets deprecated, I would rename the old version to be getFilterValue2 or some function name that will be deprecated and then have the new version be whatever you want the final version name to be so you don't have to do any re-naming later

libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortBar.tsx
RubyZ10: yes, as mentioned in the meeting, will remove it once explanation dashboard is deprecated

libs/core-ui/.eslintrc.json
```diff
@@ -7,7 +7,11 @@
       "rules": {}
     },
     {
-      "files": ["JointDataset.ts"],
+      "files": [
+        "JointDataset.ts",
+        "CohortEditorFilterSection.tsx",
```
hawestra, line 12: why do these two need to be ignored for the lint ?

libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortEditorFilter.tsx
```diff
@@ -42,47 +51,83 @@ export interface ICohortEditorFilterProps {
...
+  private readonly dataArray: IComboBoxOption[] = this.props.isRefactorFlightOn
+    ? this.props.dataset.feature_names.map((featureName, index) => {
+        return {
+          index,
+          key: featureName,
+          text:
+            featureName.length <= maxLength
+              ? featureName
+              : `${featureName.slice(0, maxLength)}...`
+        };
```
hawestra, line 72: nit: you could use lodash truncate which might be a little more readable here but I"m not sure if lodash is included already in the package

libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortEditorFilter.tsx
```diff
@@ -42,47 +51,83 @@ export interface ICohortEditorFilterProps {
...
+          return {
+            index,
+            key,
+            text: this.props.jointDataset.metaDict[key].abbridgedLabel
+          };
+        });
+
   public render(): React.ReactNode {
     const selectedMeta =
-      this.props.jointDataset.metaDict[this.props.openedFilter.column];
```
hawestra, line 66: because you're checking the refactor flight a few times in the render function, would it make sense to call a different render function whether the flight is on or not ? something like : `public render(): React.ReactNode { if(this.props.isRefactorFlightOn) {
renderWithFlight();}
else {
render();
}
}`

libs/core-ui/.eslintrc.json
RubyZ10: Otherwise I can try to refactor them.
Adding them here is temporary as they will be back to less than 30 lines after removing the old code

libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortEditorFilter.tsx
```diff
@@ -42,47 +51,83 @@ export interface ICohortEditorFilterProps {
...
-  )
-    .fill(0)
-    .map((_, index) => {
-      const key = JointDataset.DataLabelRoot + index.toString();
-      return {
-        key,
-        text: this.props.jointDataset.metaDict[key].abbridgedLabel
-      };
-    });
+  private readonly dataArray: IComboBoxOption[] = this.props.isRefactorFlightOn
```
imatiach-msft, line 63: I thought we agreed we won't have a flight for these changes but just either have both cohorts side by side or an adapter to convert between them?

libs/core-ui/src/lib/DatasetCohort.ts
```diff
@@ -0,0 +1,236 @@
+// Copyright (c) Microsoft Corporation.
```
imatiach-msft, line 1: This PR is really large and hard to review.  Maybe we can break it up a bit more - eg the DatasetCohort.ts could be its own PR to start with.

libs/core-ui/src/lib/DatasetCohort.ts
```diff
@@ -0,0 +1,236 @@
...
+    row: { [key: string]: unknown },
+    filters: IFilter[]
+  ): boolean {
+    return filters.every((filter) => {
+      const rowVal = row[filter.column];
+      switch (filter.method) {
+        case FilterMethods.Equal:
+          return rowVal === filter.arg[0];
+        case FilterMethods.GreaterThan:
+          return typeof rowVal == "number" && rowVal > filter.arg[0];
```
imatiach-msft, line 132: nit: refactor out the portion:
```
typeof rowVal == "number"
```
into its own small function and reuse

PR #1993, opened 2023-03-02
Forecasting: UI code adjustments to renamed datetime_features and time_series_id_features

romanlutz:
## Description

Due to a recent change in PR #1948 we need to rename `time_column_name` to `datetime_features` (plural, even though right now we only support 1, and only for forecasting) and `time_series_id_column_names` to `time_series_id_features`.

This PR makes that adjustment and adds a little bit of validation code.

## Checklist


libs/core-ui/src/lib/util/TimeUtils.ts
```diff
@@ -20,11 +20,18 @@ export function isTimeOrTimeSeriesIDColumn(
...
-  if (featureMetaData) {
-    const isDatetimeFeature = featureMetaData.time_column_name === featureName;
+  // The if-condition below should always be true since these are all required
+  // for forecasting and validated upon building the model assessment context.
+  // The if statement exists merely to satisfy the compiler.
+  if (
+    featureMetaData &&
+    featureMetaData.datetime_features &&
+    featureMetaData.datetime_features?.length > 0
+  ) {
```
RubyZ10, line 30: can remove the question mark? 
(
    featureMetaData &&
    featureMetaData.datetime_features &&
    featureMetaData.datetime_features.length > 0
  )

RubyZ10 (approved):

libs/core-ui/src/lib/util/TimeUtils.ts
hawestra: also: if(
featureMetaData?.datetime_features &&
featureMetaData.datetime_features.length > 0
)

hawestra (approved):

PR #1995, opened 2023-03-07
Object Detection template support

Advitya17:
## Description

Defined enums and template functions for object detection support in the Model Overview section (and parts of Data Characteristics view).

Wanted to get initial feedback on the ground work needed to iteratively add object detection support and identify areas for code changes.

## Checklist


libs/core-ui/src/lib/util/StatisticsUtils.ts
```diff
@@ -277,6 +278,9 @@ export const generateMetrics: (
       return generateImageStats(trueYSubset, predYSubset);
     });
   }
+  if (modelType == ModelTypes.ObjectDetection) {
```
natalie-isak, line 281: Looks like typescript uses === for boolean equals expressions

natalie-isak:
Why is the confidence_score optional?

libs/core-ui/src/lib/util/ObjectDetectionStatisticsUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export const generateObjectDetectionStats: (
+    jointDataset: JointDataset,
+    selectionIndexes: number[][]
+  ) => ILabeledStatistic[][] = (
+    jointDataset: JointDataset,
+    selectionIndexes: number[][]
+  ): ILabeledStatistic[][] => {
+    const numLabels = jointDataset.numLabels;
```
imatiach-msft, line 26: looks like there are some linting issues here - variables that are never used?

libs/core-ui/src/lib/Interfaces/IDataset.ts
```diff
@@ -11,13 +11,14 @@ export enum DatasetTaskType {
...
-  Forecasting = "forecasting"
+  Forecasting = "forecasting",
+  ObjectDetection = "object_detection"
 }
 
 export interface IDataset {
   task_type: DatasetTaskType;
-  true_y: number[] | number[][];
-  predicted_y?: number[] | number[][];
+  true_y: number[] | number[][] | number[][][];
```
imatiach-msft, line 20: I think there might be some test failures related to this change which may require other file changes - also, did you converge on using this format vs the additional object detection variable?

imatiach-msft (approved):

PR #2002, opened 2023-03-13
[Model Overview] Metrics Template support with OD-specific optional args

Advitya17:
## Description

This PR defines enums and template functions for object detection support in the Model Overview section (and parts of Data Characteristics view).

This PR also adds in optional args (i.e. `object_detection_{true/predicted}_y` to the dataset interfaces) to store the OD label representation:
image

This PR explores an alternate direction to #1995 . 

image

image

image

## Checklist


libs/core-ui/src/lib/util/StatisticsUtils.ts
```diff
@@ -277,6 +278,9 @@ export const generateMetrics: (
       return generateImageStats(trueYSubset, predYSubset);
     });
   }
+  if (modelType == ModelTypes.ObjectDetection) {
```
natalie-isak, line 281: Looks like TS uses === not == for equals

Advitya17: Omg never even noticed, thanks a lot!

libs/core-ui/src/lib/Interfaces/IDataset.ts
```diff
@@ -31,6 +31,8 @@ export interface IDataset {
   feature_metadata?: IFeatureMetaData;
   images?: string[];
   index?: string[];
+  object_detection_true_y?: number[][][];
```
natalie-isak, line 34: I'm actually a bit confused about how this works. If we have object_detection_true_y, then what is in true_y for the OD scenario? I can see in the [Option 1] PR you have it written differently (only true_y with different input types), and I'm curious why this option may be better?

Advitya17: Discussed offline. `{predicted/true}_y` will have the object counts array instead of multilabel scenario's binary array.

Advitya17, line 34: @imatiach-msft where are these python args passed to the typescript interfaces? Wish to emulate something similar:

image

image

Also the changes may involve defining a new interface, parsing python list to that interface, and including that interface in unit tests so I propose doing that in the next PR to avoid overloading this one.

natalie-isak: Thanks for investigating this! My concern here is that when implementing the explainability logic, I was planning on relying on the indexes of the bounding boxes. Would that still be possible with a dictionary representation? What would be the keys of the dictionary?

imatiach-msft (approved):

PR #2003, opened 2023-03-13
[Refactor]1.Add DatasetCohort 2. Move FeatureFlights to core 

RubyZ10:
## Description

1. Add DatasetCohort
When creating a new cohort from UI, according to the filters the customer selected, we will create a new DatasetCohort and save the selectedIndexes[] and cohort metrics there. (creating and adding datasetCohort to context will be in following PR)
With the selectedIndexes[] and dataset, we should be able to get everything we need for each cohort
2. Move FeatureFlights to core

## Checklist


libs/core-ui/src/lib/DatasetCohort.ts
```diff
@@ -0,0 +1,236 @@
...
+        val.forEach((subVal, subIndex) => {
+          dataDict[index][
+            DatasetCohortColumns.PredictedY + subIndex.toString()
+          ] = subVal;
+        });
+      } else {
+        dataDict[index][DatasetCohortColumns.PredictedY] = val;
+      }
+    });
+    // set up errors
```
tongyu-microsoft, line 92: do we consider multi label classification scenario?

libs/core-ui/src/lib/util/datasetUtils/getPropertyValues.ts
```diff
@@ -0,0 +1,117 @@
...
+    if (predictedYs) {
+      return indexes.map((index) => {
+        return predictedYs[index];
+      });
+    }
+  }
+  const classNames = dataset.class_names;
+  if (
+    classNames &&
+    property === DatasetCohortColumns.ProbabilityY + classNames[0]
```
tongyu-microsoft, line 31: for my learning, what is classNames[0] and className[1] here? will classNames always be length of 2?

tongyu-microsoft (approved):

libs/core-ui/src/lib/util/datasetUtils/getPropertyValues.ts
RubyZ10: good question! should handle not only the first two classNames. let me update,
classNames example
image

libs/core-ui/src/lib/DatasetCohort.ts
RubyZ10: is line 110 what you are asking?

RubyZ10: for multiclass error is line 110
for true Y and predict Y multi label scenario, it will be line 74 and line 85,
```
val.forEach((subVal, subIndex) => {
          dataDict[index][
            DatasetCohortColumns.PredictedY + subIndex.toString()
          ] = subVal;
```

PR #2006, opened 2023-03-14
[Refactor]Add columnRanges and modelType to ModelAssessmentContext

RubyZ10:
## Description
1. Added columnRanges
columnRanges calculate the range for each feature and the other cohort filter columns. It will be used in cohort editor panel
and as chart axis values, etc
2. Added modelType to ModelAssessmentContext

Both of them will be used in following refactor PRs.
Put them in context as they will be used cross components and should be global

## Checklist

### DIFF
--- a/libs/core-ui/src/lib/util/datasetUtils/getColumnRanges.ts
+++ b/libs/core-ui/src/lib/util/datasetUtils/getColumnRanges.ts
@@ -1,0 +1,203 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { localization } from "@responsible-ai/localization";
+import {
+  ICategoricalRange,
+  INumericRange,
+  RangeTypes
+} from "@responsible-ai/mlchartlib";
+import _ from "lodash";
+
+import { DatasetCohortColumns } from "../../DatasetCohortColumns";
+import { IDataset } from "../../Interfaces/IDataset";
+import { ModelTypes } from "../../Interfaces/IExplanationContext";
+import { IsBinary, IsMulticlass } from "../ExplanationUtils";
+
+// TODO: these ranges should come from backend, it does not make sense to calculate at FE for big data
+export function getColumnRanges(
+  dataset: IDataset,
+  modelType: ModelTypes
+): {
+  [key: string]: INumericRange | ICategoricalRange;
+} {
+  const ranges = {};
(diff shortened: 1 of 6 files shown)


libs/core-ui/src/lib/util/datasetUtils/getColumnRanges.ts
```diff
@@ -0,0 +1,203 @@
...
+  ranges: { [key: string]: INumericRange | ICategoricalRange }
+): void {
+  let categoricalValues = dataset.class_names;
+  // if it is 1D array
+  if (!Array.isArray(values[0])) {
+    if (modelType === ModelTypes.Regression) {
+      const numbers: number[] = [];
+      // this for loop is only to let it make sure values is a 1D array, so it can be used with _.max and _.min
+      values.forEach((value) => {
+        if (!Array.isArray(value)) {
```
tongyu-microsoft, line 162: do we want to raise error if some value here is an array?

tongyu-microsoft: or is there any existing check somewhere else already?

libs/model-assessment/src/lib/ModelAssessmentDashboard/utils/getModelTypeFromProps.ts
```diff
@@ -0,0 +1,55 @@
...
+} from "@responsible-ai/core-ui";
+
+import { IModelAssessmentDashboardProps } from "../ModelAssessmentDashboardProps";
+
+export function getModelTypeFromProps(
+  props: IModelAssessmentDashboardProps
+): ModelTypes {
+  let modelType: ModelTypes = ModelTypes.Multiclass;
+  const classNames = props.dataset.class_names;
+  if (props.dataset.task_type === DatasetTaskType.Regression) {
```
tongyu-microsoft, line 18: since we have many if conditions in the function, can we do early return in each if condition? Or maybe use switch?

tongyu-microsoft (approved):

libs/core-ui/src/lib/util/datasetUtils/getColumnRanges.ts
RubyZ10: it is handled in the below else statement - from line 181 to the end

PR #2007, opened 2023-03-15
[Refactor] Add preBuiltCohort to datasetCohorts in modelAssessmentContext

RubyZ10:
## Description
This PR added preBuiltCohort to datasetCohorts and added datasetCohorts into modelAssessmentContext

1. The preBuiltCohort from SDK has different filter format for categorical features from datasetCohort filter, so translate the filter of each preBuiltCohort and add to datasetCohorts
for example, in below screenshots the arg has the actual feature value, but in datasetCohort filter, it is storing the index of those feature values, so we need to translate the arg. 
image

FYI, The filter format of DatasetCohort is the same as what SDK expected when calling from UI. in below screen shots it is a actual call from UI to SDK, column "Programming language" is a categorical feature, and the arg is passing the index of selected values.  this is exactly the same as DatasetCohort filter.
image

2. Add datasetCohorts in modelAssessmentContext

## Checklist


libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/ProcessPreBuiltDatasetCohort.ts
```diff
@@ -0,0 +1,199 @@
...
+          case CohortColumnNames.TrueY: {
+            const filter = translateCohortFilterForTarget(
+              preBuiltCohortFilter,
+              props.dataset,
+              modelType,
+              CohortColumnNames.TrueY
+            );
+            filterList.push(filter);
+            break;
+          }
```
imatiach-msft, line 75:
```suggestion
          case CohortColumnNames.TrueY:
          case CohortColumnNames.PredictedY: {
            const filter = translateCohortFilterForTarget(
              preBuiltCohortFilter,
              props.dataset,
              modelType,
              preBuiltCohortFilter.column
            );
            filterList.push(filter);
            break;
          }
```

libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/ProcessPreBuiltDatasetCohort.ts
```diff
@@ -0,0 +1,199 @@
...
+          case CohortColumnNames.RegressionError: {
+            const filter: IFilter = {
+              arg: preBuiltCohortFilter.arg,
+              column: DatasetCohortColumns.RegressionError,
+              method: preBuiltCohortFilter.method
+            };
+            filterList.push(filter);
+            break;
+          }
```
imatiach-msft, line 55:
```suggestion
          case CohortColumnNames.RegressionError:
          case CohortColumnNames.Index: {
            let column = DatasetCohortColumns.Index
            if (preBuiltCohortFilter.column === CohortColumnNames.RegressionError) {
                column = DatasetCohortColumns.RegressionError
            }
            const filter: IFilter = {
              arg: preBuiltCohortFilter.arg,
              column: column,
              method: preBuiltCohortFilter.method
            };
            filterList.push(filter);
            break;
          }
```

libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/ProcessPreBuiltDatasetCohort.ts
```diff
@@ -0,0 +1,199 @@
...
+        allowedClassificationErrorValues.indexOf(classificationError)
+      )
+      .filter((index) => index !== -1);
+  }
+  index.sort((a, b) => a - b);
+  const filter: IFilter = {
+    arg: index,
+    column: DatasetCohortColumns.ClassificationError,
+    method: preBuiltCohortFilter.method
+  };
```
imatiach-msft, line 152: what do you think about extracting this out to a small method that takes in arg, column as parameters and returns a method filter?  Like "createMethodFilter"?
It could be re-used in several places in this file - lines 161, 135, 127, 39, 48, 190
So it could reduce a lot of lines in this file?

imatiach-msft (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/ProcessPreBuiltDatasetCohort.ts
RubyZ10: thanks Ilya,  I think we also need to pass in preBuiltCohortFilter.method, as the third parameters? or are you suggesting to create a new filter type? please let me know if you have other thoughts
and by extracting this out to a small method by taking those three params and return a IFilter. it  is actually taking more lines, as there will be one more function and the value for column and method are too long, when leveraging the function, it still takes 5 lines everytime. Overall the file increases 8 lines.

RubyZ10: merged the change, plz let me know if any thoughts, I can always update thanks!

PR #2008, opened 2023-03-15
 Add validations for model predictions

gaugup:
This validation prevents the downstream failure when trying to serialize the `RAIInsights` object or computing explanations if the predicted set of values aren't passed as valid classes to `RAIInsights`.

## Checklist

### DIFF
--- a/responsibleai/responsibleai/rai_insights/rai_insights.py
+++ b/responsibleai/responsibleai/rai_insights/rai_insights.py
@@ -616,7 +616,7 @@ class RAIInsights(RAIBaseInsights):
             # do not change the input data.
             if task_type != ModelTask.FORECASTING:
                 self._ensure_model_outputs(input_data=small_train_data)
-                self._ensure_model_outputs(input_data=small_test_data)
+            self._ensure_model_outputs(input_data=small_test_data)
 
             if task_type == ModelTask.REGRESSION:
                 if hasattr(model, SKLearn.PREDICT_PROBA):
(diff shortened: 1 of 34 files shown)


responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -642,6 +625,67 @@ def _validate_rai_insights_input_parameters(
...
+                        len(set(classes) -
+                            set(np.unique(train_predictions))) != 0):
+                    raise UserConfigValidationException(
+                        'The train labels and distinct values in '
+                        'predictions (train data) do not match')
+
+                if (len(set(np.unique(test_predictions)) -
+                        set(classes)) != 0 or
+                        len(set(classes) -
+                            set(np.unique(test_predictions))) != 0):
```
hawestra, line 684: nit: since you're repeating this check a couple times, you could make it a helper function that returns a boolean

hawestra: and potentially add the exception in the helper function too

tongyu-microsoft (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -642,6 +625,67 @@ def _validate_rai_insights_input_parameters(
...
+                        columns=[target_column], axis=1)
+
+                train_predictions = model.predict(train_data)
+                test_predictions = model.predict(test_data)
+
+                if (len(set(np.unique(train_predictions)) -
+                        set(classes)) != 0 or
+                        len(set(classes) -
+                            set(np.unique(train_predictions))) != 0):
+                    raise UserConfigValidationException(
```
hawestra, line 677: how would this exception be raised if we already checked the train and test data ?

gaugup: This will be raised if predictions of the model do not match with the classes passed in by the user.

PR #2012, opened 2023-03-23
[Refactor] Refactor Cohort Editor

RubyZ10:
## Description
This PR refactor the filter used in Cohort Editor. The filter column will use the "real" column name, exactly as what backend expected
Before change,
legacy filter
image
After change:
new filter
image

With this PR, the cohort editor flow will be using the new filter. 
A few things to notice:
As we want to have both DatasetCohort,(which will be using the new filter), and ErrorCohort,(which uses the legacy filter) available in context. So when save cohort, we use the new filter to create DatasetCohort, and translate it to legacy filter and use the legacy filter to create Error Cohort
When open the cohort editor panel, we pass the existing filterList to CohortEditor, which currently is from ErrorCohort.filters, which are the legacy filters, so we translate the legacy fitlers to new filters, and use the new filters in the Cohort Editor
As CohortEditor is used not only by model assessment dashboard, but also by the interpretation dashboard. So some workaround was made for in the interpretation dashboard, so it could use the refactored CohortEditor

## Checklist


libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortBar.tsx
```diff
@@ -25,6 +25,7 @@ export interface ICohortBarProps {
   cohorts: Cohort[];
   jointDataset: JointDataset;
   modelMetadata: IExplanationModelMetadata;
+  features: unknown[][] | undefined;
```
imatiach-msft, line 28: why unknown here? there is no more specific type or set of types we can give here?

RubyZ10: picking what it is in IDataset, I think it will be either a number or string two D array, if we are sure about it I can change the IDataset
image

libs/interpret/src/lib/MLIDashboard/utils/buildColumnRanges.ts
```diff
@@ -0,0 +1,65 @@
...
+export function buildColumnRanges(jointDataset: JointDataset): {
+  [key: string]: IColumnRange;
+} {
+  const range: {
+    [key: string]: IColumnRange;
+  } = {};
+  for (const [key, value] of Object.entries(jointDataset.metaDict)) {
+    const columnRange = getRange(key, value, jointDataset.dataDict);
+    if (columnRange) {
+      if (key === JointDataset.RegressionError) {
```
imatiach-msft, line 22: you could simplify here by creating a dictionary with JointDataset keys and DatasetCohortColumns values - then index into that to get the value to set the columnRange

imatiach-msft (approved):

libs/interpret/src/lib/MLIDashboard/utils/buildColumnRanges.ts
RubyZ10: sounds good. will update in a separate PR, thank you!

imatiach-msft (approved):

libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortEditorFilter.tsx
```diff
@@ -42,48 +42,50 @@ export interface ICohortEditorFilterProps {
...
+      ? this.props.columnRanges[this.props.openedFilter.column]
+      : undefined;
+    const isCategorical =
+      columnRange?.treatAsCategorical ||
+      columnRange?.rangeType === RangeTypes.Categorical;
+    if (isCategorical) {
+      categoricalOptions = columnRange?.sortedUniqueValues.map(
         (label, index) => {
-          return { key: index, text: label };
+          return { key: index, text: String(label) };
```
tongyu-microsoft, line 84: is there a difference for String(A) and A.toString()?

libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortEditorUtils.ts
```diff
@@ -0,0 +1,148 @@
...
+    return translateToNewFilters(props.filterList, metadata?.featureNames);
+  }
+  return translateToNewFilters(props.filterList, dataset?.feature_names);
+}
+
+export function translateToLegacyFilterColumn(
+  column: string,
+  featureNames: string[]
+): string {
+  if (column === DatasetCohortColumns.Index) {
```
tongyu-microsoft, line 30: feel like this is better if using "switch case", but since this will be deprecated soon, either is fine to me!

tongyu-microsoft (approved):

libs/core-ui/src/lib/Cohort/ManualCohortManagement/CohortEditorFilter.tsx
RubyZ10: the type of label here is unknown, and we cannot use toString() for unknown type as it could be null or undefined. 
fyi, https://stackoverflow.com/questions/3945202/whats-the-difference-between-stringvalue-vs-value-tostring

PR #2013, opened 2023-03-28
Add validations if target column is included in `FeatureMetadata`

gaugup:
## Description

Add validations if target column is included in `FeatureMetadata`. If we don't do this validation a downstream validation (predict validation) will fail with cryptic errors.

## Checklist

### DIFF
--- a/responsibleai/responsibleai/rai_insights/rai_insights.py
+++ b/responsibleai/responsibleai/rai_insights/rai_insights.py
@@ -576,26 +576,29 @@ class RAIInsights(RAIBaseInsights):
                 f"{non_categorical_or_time_string_columns}")
 
         self._validate_feature_metadata(
-            feature_metadata, train, task_type, model)
+            feature_metadata, train, task_type, model, target_column)
 
         if model is not None:
             # Pick one row from train and test data
             small_train_data = train[0:1]
             small_test_data = test[0:1]
             has_dropped_features = False
-            if feature_metadata is not None:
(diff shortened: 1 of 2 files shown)


responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -576,26 +576,29 @@ def _validate_rai_insights_input_parameters(
...
-                columns=[target_column], axis=1)
-            small_test_data = small_test_data.drop(
-                columns=[target_column], axis=1)
+            if feature_metadata is not None and \
+                (feature_metadata.dropped_features is not None and
+                    len(feature_metadata.dropped_features) != 0):
+                has_dropped_features = True
+                small_train_data = small_train_data.drop(
+                    columns=feature_metadata.dropped_features + [
+                        target_column], axis=1)
```
hawestra, line 592: nit: set features_to_drop above in an if statement as either feature_metadata.dropped_features + [target_column] or [target_column], then small_train_data.drop(columns=features_to_drop, axis=1) outside of if statement, also since this is getting longer, maybe making a helper function would make it more readable

hawestra (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
gaugup: done

imatiach-msft (approved):

PR #2023, opened 2023-04-03
add scrollbar to RAI text dashboard text highlighting area

imatiach-msft:
## Description

Add scrollbar to RAI text dashboard text highlighting area

image

## Checklist


libs/interpret-text/src/lib/TextExplanationDashboard/Control/TextHighlighting/TextHightlighting.tsx
```diff
@@ -28,51 +36,55 @@ export class TextHighlighting extends React.PureComponent<IChartProps> {
...
+      <Stack styles={scrollablePaneStyles}>
+        <ScrollablePane scrollbarVisibility={ScrollbarVisibility.auto}>
+          <Stack
+            id="TextHighlighting"
+            horizontal
+            horizontalAlign="start"
+            tokens={textStackTokens}
+            wrap
+            styles={textStackStyles}
+          >
```
hawestra, line 48: Does it work to only have one stack ? like <ScrollablePane> <Stack>...</Stack> </ScrollablePane>? I'm not sure what the outstack is adding that the pane couldn't

hawestra (approved):

libs/interpret-text/src/lib/TextExplanationDashboard/Control/TextHighlighting/TextHightlighting.tsx
imatiach-msft: the scrollable pane needs to go inside a stack with position: relative according to docs... but I can't do that in the inner stack and also if scroll goes in inner stack it messes up the spacing of the words for some reason

PR #2029, opened 2023-04-10
make train data optional on RAIBaseInsights class

imatiach-msft:
## Description

Make train data optional on RAIBaseInsights class.
The train data is optional for RAITextInsights and RAIVisionInsights as it is not used currently.

## Checklist

### DIFF
--- a/responsibleai/responsibleai/rai_insights/rai_base_insights.py
+++ b/responsibleai/responsibleai/rai_insights/rai_base_insights.py
@@ -29,7 +29,7 @@ class RAIBaseInsights(ABC):
     This class is abstract and should not be instantiated.
     """
 
-    def __init__(self, model: Optional[Any], train: pd.DataFrame,
+    def __init__(self, model: Optional[Any], train: Optional[pd.DataFrame],
                  test: pd.DataFrame, target_column: str, task_type: str,
                  serializer: Optional[Any] = None):
         """Creates an RAIBaseInsights object.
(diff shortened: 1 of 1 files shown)


responsibleai/responsibleai/rai_insights/rai_base_insights.py
```diff
@@ -29,7 +29,7 @@ class RAIBaseInsights(ABC):
     This class is abstract and should not be instantiated.
     """
 
-    def __init__(self, model: Optional[Any], train: pd.DataFrame,
+    def __init__(self, model: Optional[Any], train: Optional[pd.DataFrame],
```
gaugup, line 32: Does python allow this? An Optional parameter before required parameter? 

On another thought, could text/image scenario pass in None for training data and we handle it accordingly here?

imatiach-msft: That's what we do for the "model" parameter, right?  And we actually allow users not to specify it for causal?  It's also marked as optional.

gaugup: Adding Optional keyword here just indicates the static code checker that this can be optional.

Since model can be None in case of causal, we supply None. The downstream code indicates similar handling for train variable. 

Shall we update the doc strings for train in rai_insights.py? I don't think adding Optional keyword is adding useful indicator for the user.

vinuthakaranth (approved):

PR #2038, opened 2023-04-19
fix failing builds by upgrading pip-tools after latest pip release

imatiach-msft:
## Description

Latest builds are failing with error:

```
in get_dependencies
    wheel_cache = WheelCache(self._cache_dir, self.options.format_control)
TypeError: __init__() takes 2 positional arguments but 3 were given
```

The latest pip release broke compatibility with old pip-tools.  Solution seems to be to upgrade pip-tools since we are pinned to an old version currently.

See related issue: https://github.com/jazzband/pip-tools/issues/1823

## Checklist

### DIFF
--- a/.github/workflows/CI-e2e-notebooks.yml
+++ b/.github/workflows/CI-e2e-notebooks.yml
@@ -88,7 +88,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install --upgrade setuptools
-          pip install --upgrade "pip-tools<6.12.2"
+          pip install --upgrade "pip-tools<=6.13.0"
 
       - name: Install dependencies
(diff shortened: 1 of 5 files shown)


gaugup (commented):
Error: Invalid value for '[SRC_FILES]...': Path 'requirements-dev-comp.txt' does not exist.

it looks like it is running into problem with file not found error.

imatiach-msft:
> Error: Invalid value for '[SRC_FILES]...': Path 'requirements-dev-comp.txt' does not exist.
> 
> it looks like it is running into problem with file not found error.

yes I was missing a -o argument it looks like

gaugup (approved):

vinuthakaranth (approved):

PR #2039, opened 2023-04-20
Ensuring backwards compat b/w OD scenario and other vision scenarios

natalie-isak:


raiwidgets/raiwidgets/responsibleai_dashboard_input.py
```diff
@@ -48,8 +48,8 @@ def __init__(
             self.dashboard_input.cohortData = []
 
         self._feature_length = len(self.dashboard_input.dataset.feature_names)
-        if hasattr(analysis, ManagerNames.ERROR_ANALYSIS):
```
imatiach-msft, line 51: why is this part commented out? I think we should still add it back in?

imatiach-msft (approved):

raiwidgets/raiwidgets/responsibleai_dashboard_input.py
natalie-isak: fixed, that was just for testing :)

PR #2044, opened 2023-04-25
fix transient test failures by increasing timeout for gbm shap method

imatiach-msft:
## Description

Fix transient test failures by increasing timeout for gbm shap method.

## Checklist

### DIFF
--- a/erroranalysis/tests/test_importances.py
+++ b/erroranalysis/tests/test_importances.py
@@ -115,7 +115,11 @@ class TestImportances(object):
         print(execution_time)
         # assert we don't take too long and downsample the dataset
         # note execution time is in seconds
-        assert execution_time < 20
+        # note GBM_SHAP is slower than the other methods
+        if error_correlation_method == GBM_SHAP:
+            assert execution_time < 35
+        else:
+            assert execution_time < 25
 
     @pytest.mark.parametrize('num_rows', [1, 2, 3, 4])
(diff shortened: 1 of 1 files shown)


erroranalysis/tests/test_importances.py
```diff
@@ -115,7 +115,11 @@ def test_large_data_importances(self, error_correlation_method):
         print(execution_time)
         # assert we don't take too long and downsample the dataset
         # note execution time is in seconds
-        assert execution_time < 20
+        # note GBM_SHAP is slower than the other methods
+        if error_correlation_method == GBM_SHAP:
+            assert execution_time < 35
+        else:
+            assert execution_time < 25
```
gaugup, line 122: I think adding timeouts like these makes the tests very flaky. Is there any other way you validate this? maybe have a higher upper bound of like a minute or so where you know that execution time has deteriorated that it needs to be looked at.

imatiach-msft: These have been working fine previously - the issue is that this bug indeed was due to this code path taking too long.  In reality even this timeout is much longer than what I would usually see, just 1-4 seconds.  But it seems it happens once in a blue moon, and since we have so many builds and tests I just bump up the time to make it happen even less.  I don't want to bump it up too high though because sometimes indeed there can be an issue due to some code change and I want to keep the performance fast.

gaugup (approved):

RubyZ10 (approved):

PR #2034, opened 2023-04-16
QA Scenario metrics template

mehektulsyan:
Placeholder for metrics for Question Answering Scenario

## Description

This PR contains a general placeholder for metrics for question answering scenario.

Some metrics are calculated here, while there is placeholder for the rest to be integrated from responsibleai-text repo.

PR #2048, opened 2023-05-02
remove QA statistics which were mistakenly added to multilabel metrics

imatiach-msft:
## Description

In this PR to add new statistics for QA task:
https://github.com/microsoft/responsible-ai-toolbox/pull/2034
It looks like some QA metrics were mistakenly added for multilabel text and vision scenarios.
This PR removes these metrics, which currently just show dummy zero values.

image

## Checklist

### DIFF
--- a/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
+++ b/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
@@ -153,11 +153,7 @@ export class ModelOverview extends React.Component<
     ) {
       defaultSelectedMetrics = [
         MultilabelMetrics.ExactMatchRatio,
-        MultilabelMetrics.HammingScore,
-        MultilabelMetrics.MeteorScore,
-        MultilabelMetrics.BleuScore,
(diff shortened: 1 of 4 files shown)


libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -172,7 +168,9 @@ export class ModelOverview extends React.Component<
     ) {
       defaultSelectedMetrics = [
         QuestionAnsweringMetrics.ExactMatchRatio,
-        QuestionAnsweringMetrics.F1Score
```
mehektulsyan, line 175: We dont want to remove F1Score from here

tongyu-microsoft (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
imatiach-msft: I thought it's not an interesting default metric, and we don't want too many, so that's why I removed it... user can still add it in.  None of the other tasks have this many default metrics.

PR #2049, opened 2023-05-03
Reset ComboBox Option for Object Detection

natalie-isak:
This PR resets the state for the selected combo box option once the flyout is closed


imatiach-msft (approved):

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
```diff
@@ -103,9 +100,13 @@ export class FlyoutObjectDetection extends React.Component<
       });
     }
   }
-
+  public callbackWrapper = (): void => {
```
imatiach-msft, line 103: can this be made a private function instead?  it doesn't seem like it needs to be public since it's not called from another file but just within this file?

imatiach-msft (approved):

tongyu-microsoft (approved):

PR #2050, opened 2023-05-03
Prevent overwriting of dataset in `erroranalysis`

gaugup:
## Description

Fix overwriting of dataset in `erroranalysis`. This overwriting only happens in case there are only categorical features in the dataset. It doesn't trigger in case there only numerical features or mix of numerical and categorical features.

## Checklist

### DIFF
--- a/erroranalysis/erroranalysis/_internal/surrogate_error_tree.py
+++ b/erroranalysis/erroranalysis/_internal/surrogate_error_tree.py
@@ -302,7 +302,7 @@ def get_surrogate_booster_local(filtered_df, analyzer, is_model_analyzer,
     if is_pandas:
         true_y = true_y.to_numpy()
     else:
-        input_data = input_data.to_numpy()
+        input_data = input_data.to_numpy(copy=True)
     if is_model_analyzer:
         pred_y = analyzer.model.predict(input_data)
     if analyzer.model_task == ModelTask.CLASSIFICATION:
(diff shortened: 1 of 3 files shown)


imatiach-msft (approved):

erroranalysis/erroranalysis/analyzer/error_analyzer.py
```diff
@@ -442,7 +442,7 @@ def compute_importances(self, error_correlation_method=MUTUAL_INFO):
         input_data = self.dataset
         diff = self.get_diff()
         if isinstance(self.dataset, pd.DataFrame):
-            input_data = input_data.to_numpy()
+            input_data = input_data.to_numpy(copy=True)
```
imatiach-msft, line 445: is it possible to add some test to validate the scenario you were hitting won't happen in the future?

vinuthakaranth (approved):

erroranalysis/erroranalysis/_internal/surrogate_error_tree.py
```diff
@@ -302,7 +302,7 @@ def get_surrogate_booster_local(filtered_df, analyzer, is_model_analyzer,
     if is_pandas:
         true_y = true_y.to_numpy()
     else:
-        input_data = input_data.to_numpy()
+        input_data = input_data.to_numpy(copy=True)
```
vinuthakaranth, line 305: Are there any other places in the repo where data.to_numpy is used and this change needs to be done there too?

vinuthakaranth (approved):

erroranalysis/erroranalysis/analyzer/error_analyzer.py
gaugup: added. This is only triggered in case there are only categorical features in the dataset.

erroranalysis/erroranalysis/_internal/surrogate_error_tree.py
gaugup: It is hard to validate all the cases. So making changes and adding tests for the cases reported by customers.

PR #2051, opened 2023-05-03
Fix robot page exception on feature importance when dropped_feature is defined

tongyu-microsoft:
This PR fixes the bug where the index for x axis and y axisfor feature importances plot was not correctly set when dropped_feature has length > 0.

## Description
There are two issues:
1) y axis: Feature importances are only calculated for non-dropped features. So we should pass featureIndex instead of featureIndexBeforeDrop to yKey (change on line 602)
2) x axis: this.handleFeatureSelection is a shared method for both <FeatureImportanceBar/> and FeatureOptions' <ComboBox/>. We should pass indexAfterDrop to this.handleFeatureSelection method. 
feautureImportanceRobotPage

## Checklist

### DIFF
--- a/libs/interpret/src/lib/MLIDashboard/Controls/GlobalExplanationTab/GlobalExplanationTab.tsx
+++ b/libs/interpret/src/lib/MLIDashboard/Controls/GlobalExplanationTab/GlobalExplanationTab.tsx
@@ -554,8 +554,18 @@ export class GlobalExplanationTab extends React.PureComponent<
     if (typeof item?.key === "string") {
       const key = item.key;
       const index = this.context.jointDataset.metaDict[key].index;
-      if (index !== undefined) {
-        this.handleFeatureSelection(this.state.selectedCohortIndex, index);
+      let indexAfterDrop;
+      for (const i of this.featureIndexMap.keys()) {
+        if (index && this.featureIndexMap.get(i) === index) {
+          indexAfterDrop = i;
+          break;
+        }
+      }
+      if (indexAfterDrop !== undefined) {
+        this.handleFeatureSelection(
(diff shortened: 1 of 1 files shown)


libs/interpret/src/lib/MLIDashboard/Controls/GlobalExplanationTab/GlobalExplanationTab.tsx
```diff
@@ -554,8 +554,18 @@ export class GlobalExplanationTab extends React.PureComponent<
...
-      if (index !== undefined) {
-        this.handleFeatureSelection(this.state.selectedCohortIndex, index);
+      let indexAfterDrop;
+      for (const i of this.featureIndexMap.keys()) {
+        if (index && this.featureIndexMap.get(i) === index) {
+          indexAfterDrop = i;
+          break;
+        }
+      }
+      if (indexAfterDrop !== undefined) {
```
imatiach-msft, line 564: nit: why not just call this inside the if and then break?  then you won't need to keep track of indexAfterDrop at all.  Maybe I am missing something here.

imatiach-msft (approved):

libs/interpret/src/lib/MLIDashboard/Controls/GlobalExplanationTab/GlobalExplanationTab.tsx
tongyu-microsoft: Great comment. Updated!

tongyu-microsoft: Thanks!

PR #2054, opened 2023-05-05
Integrate Question answering metrics from python backend 

mehektulsyan:
image (1)

### DIFF
--- a/libs/localization/src/lib/en.json
+++ b/libs/localization/src/lib/en.json
@@ -1690,23 +1690,23 @@
         },
         "exactMatchRatio": {
           "name": "Exact match ratio",
-          "description": "The ratio of instances classified correctly for every label in question answering task."
+          "description": "The ratio of instances classified correctly for every label."
         },
         "meteorScore": {
           "name": "Meteor Score",
-          "description": "The ratio of instances classified correctly for every label in question answering task."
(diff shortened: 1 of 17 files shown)


raiwidgets/raiwidgets/responsibleai_dashboard_input.py
```diff
@@ -366,17 +366,9 @@ def get_question_answering_metrics(self, post_data):
         :rtype: Dict[str, List]
         """
         try:
-            true_y = post_data[0]
-            predicted_y = post_data[1]
-            aggregate_method = post_data[2]
-            class_name = post_data[3]
-            iou_thresh = post_data[4]
+            selection_index = post_data[0]
```
imatiach-msft, line 369: nit: selection_indexes

imatiach-msft (approved):

libs/localization/src/lib/en.json
```diff
@@ -1689,19 +1690,23 @@
         },
         "exactMatchRatio": {
           "name": "Exact match ratio",
-          "description": "The ratio of instances classified correctly for every label in multilabel task."
+          "description": "The ratio of instances classified correctly for every label in question answering task."
```
imatiach-msft, line 1693: this is used in multilabel classification too though - so shouldn't we generalize this to just say "The ratio of instances classified correctly for every label"

PR #2065, opened 2023-05-16
Raise warning in case the test data has missing values

gaugup:
## Description

Raise warning in case the test data has missing values to make user aware of missing data columns. Also, adding a test to make sure that missing values in test data doesn't affect `get_feature_ranges` function.

## Checklist

### DIFF
--- a/responsibleai/responsibleai/rai_insights/rai_insights.py
+++ b/responsibleai/responsibleai/rai_insights/rai_insights.py
@@ -575,10 +575,14 @@ class RAIInsights(RAIBaseInsights):
                 "identified as categorical features: "
                 f"{non_categorical_or_time_string_columns}")
 
+        list_of_feature_having_missing_values = []
         for feature in test.columns.tolist():
             if np.any(test[feature].isnull()):
-                warnings.warn(
-                    f"Feature {feature} has missing values in test data")
+                list_of_feature_having_missing_values.append(feature)
+        if len(list_of_feature_having_missing_values) > 0:
+            warnings.warn(
+                f"Features {list_of_feature_having_missing_values} "
+                "have missing values in test data")
 
         self._validate_feature_metadata(
             feature_metadata, train, task_type, model, target_column)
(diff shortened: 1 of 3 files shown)


responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -575,6 +575,11 @@ def _validate_rai_insights_input_parameters(
                 "identified as categorical features: "
                 f"{non_categorical_or_time_string_columns}")
 
+        for feature in test.columns.tolist():
+            if np.any(test[feature].isnull()):
+                warnings.warn(
```
imatiach-msft, line 580: can we either:
1.) break here on first feature, and only show warning for that
2.) gather a list of all features, and print the warning once.  Maybe if there are more than 5 just print first 5 features with missings and break then.
If there are like 500 features and all have missings that will create a lot of printed warnings here I think?

imatiach-msft (approved):

responsibleai/tests/rai_insights/test_rai_insights_get_feature_ranges.py
```diff
@@ -0,0 +1,18 @@
...
+class TestRAIInsightsGetFeatureRanges:
+    def test_get_feature_range(self):
+        data = {'Category': ['A', 'B', 'C', None, 'D'],
+                'Numerical': [10, 20, None, 40, 50]}
+
+        df = pd.DataFrame(data)
+
+        feature_ranges = RAIInsights._get_feature_ranges(
+            df, ['Category'], ['Category', 'Numerical'])
+        assert'Category' in feature_ranges
```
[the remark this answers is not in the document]
gaugup: Done. Thanks for pointing this out. Weird that flake8 didn't catch this.

responsibleai/responsibleai/rai_insights/rai_insights.py
gaugup: Done. Changed the logic with second suggestion.

PR #2072, opened 2023-05-22
Raise value error for min max computation of feature range

kicha0:
Raising value error for exception in converting min / max of a feature column to float, and printing the type of the argument.

## Description

Observed value error:
<class 'ValueError'>
File "/mnt/azureml/cr/j/[guid]/exe/wd/_telemetry/_loggerfactory.py", line 143 in wrapper
return func(*args, **kwargs)
File "create_rai_insights.py", line 184 in main
_ = RAIInsights(
File "/azureml-envs/responsibleai/lib/python3.8/site-packages/responsibleai/rai_insights/rai_insights.py", line 236 in __init__
self._feature_ranges = RAIInsights._get_feature_ranges(
File "/azureml-envs/responsibleai/lib/python3.8/site-packages/responsibleai/rai_insights/rai_insights.py", line 1263 in _get_feature_ranges
max_value = float(test[col].max())

## Checklist


gaugup (changes requested):
Could you please add a unit test for this change?

kicha0:
> Could you please add a unit test for this change?

Hi, In the process of trying to troubleshoot we find missing exception information about the failure, this is exception handling (logging) for a case with unknown trigger condition.  Could you suggest a suitable unit test for this as it's not clear what a dataset needs to look like in the first place to trigger a ValueError in this part.

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -1279,8 +1279,18 @@ def _get_feature_ranges(
...
             else:
-                min_value = float(test[col].min())
-                max_value = float(test[col].max())
+                col_min = test[col].min()
+                col_max = test[col].max()
+                try:
+                    min_value = float(col_min)
+                    max_value = float(col_max)
+                except Exception as e:
+                    raise ValueError(
```
gaugup, line 1288: Could you change this to SystemErrorException?

kicha0: done.

gaugup (approved):

PR #2081, opened 2023-05-23
fix outline bug and add contributors to readme

hawestra:
image

## Description

fix bug with outline on heatmap-
before: 
image
update readme maintainer list

## Checklist


tongyu-microsoft (approved):

gaugup (approved):

romanlutz (approved):
Nice! Is it fair to assume you tried dark mode, too?

I opened the link to the blob and couldn't really check if it works because there's a bug with the heatmap toggle. It is toggled (on) to start with but the heatmap isn't showing, so the font is white on white background rendering it invisible. Dark mode seemed to work.

hawestra:
@romanlutz yep! seems like there are some other issues with dark mode (which i have a bug for) but the outline is gone at least 
image

PR #2082, opened 2023-05-24
Correct test command and add clarification on GPG Key to CONTRIBUTING.md

ziqi-ma:
- Corrected e2e test command from yarn e2eall to yarn e2e
- Added clarification on setting up GPG Key


romanlutz:
This is great! I see the commit is signed. Only issue is that it's from your own fork which will make CD fail. On the other hand we know that you only added to a md file so you couldn't have broken it... I will just bypass branch protections in this case, but please open PRs from the main repo in the future. I should have explained that and forgot, my bad!

PR #2089, opened 2023-05-30
[Big Data] Fix create cohort, filter range for index and dataset

vinuthakaranth:
## Description

1. Index range for large data
2. Feature range for large data
3. Remove 'Treat as categorical' for integer features as data is large, it will be tedious fro user to select unique values in dropbox

## Checklist

### DIFF
--- a/libs/core-ui/src/lib/util/datasetUtils/getColumnRanges.ts
+++ b/libs/core-ui/src/lib/util/datasetUtils/getColumnRanges.ts
@@ -8,8 +8,8 @@ import _ from "lodash";
 import { DatasetCohortColumns } from "../../DatasetCohortColumns";
 import { IDataset } from "../../Interfaces/IDataset";
 import { ModelTypes } from "../../Interfaces/IExplanationContext";
-import { IsBinary, IsMulticlass } from "../ExplanationUtils";
 import { ifEnableLargeData } from "../buildInitialContext";
+import { IsBinary, IsMulticlass } from "../ExplanationUtils";
 
 // TODO: these ranges should come from backend, it does not make sense to calculate at FE for big data
 export function getColumnRanges(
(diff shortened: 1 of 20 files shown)


libs/core-ui/src/lib/util/datasetUtils/getColumnRanges.ts
```diff
@@ -62,12 +65,20 @@ export function getColumnRanges(
...
-  };
+  return !ifEnableLargeData(dataset)
+    ? {
+        max: dataset.features.length - 1,
+        min: 0,
+        rangeType: RangeTypes.Integer,
+        sortedUniqueValues: [...new Array(dataset.features.length).keys()]
+      }
+    : ({
+        max: dataset.tabular_dataset_metadata?.num_rows
```
imatiach-msft, line 76: nit: move this to variable
```dataset.tabular_dataset_metadata?.num_rows```

**Review by imatiach-msft — APPROVED** [review]

File: libs/core-ui/src/lib/util/datasetUtils/getColumnRanges.ts (hunk 1 above)
Comment (vinuthakaranth) [reply libs/core-ui/src/lib/util/datasetUtils/getColumnRanges.ts:76]: Thanks, Ilya. I tried to have the below variable but the issue with that is, it will be initiated even when not needed (not in large data cases). If I try to have this initiated in large data cases, it adds more lines of code. Open to change if this can be done in an easier way :)
_const numberOfRows = dataset.tabular_dataset_metadata?.num_rows;_

File: libs/core-ui/src/lib/util/datasetUtils/getColumnRanges.ts [hunk 2]
```diff
@@ -20,7 +21,9 @@ export function getColumnRanges(
   const ranges = {};
   // get dataset features' range
   dataset.feature_names.forEach((feature) => {
-    const range = getDatasetFeatureRange(dataset, feature);
+    const range = ifEnableLargeData(dataset)
```
RubyZ10, line 24: Can we use tabular_dataset_metadata for both large data and non large data scenarios?

vinuthakaranth: I and Gaurav discussed this. We can use it, but Gaurav suggested using tabular_dataset_metadata for large data for now. tagging @gaugup

RubyZ10: Could you please a task for it to track, thank you!

PR #2094, opened 2023-06-02
Fix bug -"All data" cohort detection is based on localized strings

RubyZ10:
## Checklist


libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
```diff
@@ -51,7 +50,7 @@ export class ForecastingDashboard extends React.Component<
 
   public componentDidMount(): void {
     const timeSeries = this.context.errorCohorts.filter(
-      (cohort) => !isAllDataErrorCohort(cohort)
+      (cohort) => !cohort.isAllDataCohort
```
RubyZ10, line 53: Does this change make sense for ForecastingDashboard? Or should we keep the isAllDataErrorCohort() for ForecastingDashboard?

romanlutz: As long as you're setting `isAllDataCohort` to true only for the first All data cohort and not any future cohorts (even if they have the same rows) that's exactly what we need in forecasting :-) I believe that's what your PR does, correct?

romanlutz (approved):

libs/forecasting/src/lib/ForecastingDashboard/ForecastingDashboard.tsx
RubyZ10: Yes, exactly as you said!

vinuthakaranth (approved):

PR #2097, opened 2023-06-06
[Image Explorer] CanvasTools Image Loading support for Object Detection

Advitya17:
This PR replaces the existing image loading logic in the Flyout (using `<Image>` from FluentUI) to CanvasTools.

## Description

Advantages:
1. Reusability with Data Labeling team's work, which also unlocks loading and overlaying elements on images for CV tasks.
2. Unlocks the potential to draw bounding boxes in the _frontend_, which'll enable _on-the-fly_ functionality of filtering/displaying bounding boxes based on user inputs related to error types, classes etc. Aligned with customer feedback & feature requests.
3. Piggybacks on CanvasTools' accessibility standards (including for overlaid elements)
4. Slightly higher res display compared to `<Image>`

image

image

## Checklist

### DIFF
--- a/libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
+++ b/libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
@@ -8,7 +8,6 @@ import {
   IComboBoxOption,
   Icon,
   Image as ImageTag,
-  ImageFit,
   List,
   Panel,
   PanelType,
@@ -111,7 +110,6 @@ export class FlyoutObjectDetection extends React.Component<
   }
 
   private readonly callbackRef = (editorCallback: HTMLDivElement) => (this.setState({ editorCallback: editorCallback }));
-  private readonly imageRef = (imageCallback: HTMLImageElement) => (this.setState({ imageCallback: imageCallback }));
 
   public loadImageFromBase64(base64String: string, editor: Editor) { // onReady is a function/callable
     const image = new Image();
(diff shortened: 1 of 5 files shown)


libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
```diff
@@ -100,6 +108,93 @@ export class FlyoutObjectDetection extends React.Component<
...
+    // Stops if the div container for the canvastools editor doesn't exist
+    if (!this.state.editorCallback) {
+      return;
+    }
+    // Removes any pre-existimg images due to previous render calls
+    this.state.editorCallback.innerHTML = "";
+
+    // Initializes CanvasTools-vott editor
+    var editor = new CanvasTools.Editor(this.state.editorCallback); // store as an instance variable
+    editor.enablePathRegions(true);
```
code-vicar, line 134: enablePathRegions is only necessary if you plan on using "Path" region types (you'd use these if you needed Bezier curve support).

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
```diff
@@ -109,6 +204,9 @@ export class FlyoutObjectDetection extends React.Component<
     const classNames = flyoutStyles();
     const predictedY = getJoinedLabelString(item?.predictedY);
     const trueY = getJoinedLabelString(item?.trueY);
+
+    this.drawBoundingBoxes(item);
```
code-vicar, line 208: drawBoundingBoxes is getting called on every render, which means you are re-initializing the editor every render.

Instead, move the editor initialization to a separate function and only call it when the component mounts.

In the "component did update" lifecycle method check if the image source has changed and reset the editor content source if it has.

Draw the bounding boxes once after the editor content source has loaded (i.e. the image has loaded in the editor) and also in the "component did update" method in case the bounding box prop data has changed.

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
```diff
@@ -100,6 +108,93 @@ export class FlyoutObjectDetection extends React.Component<
...
+    // Retrieving labels for the image in the Flyout
+    const predictedY : number[][] = this.props.dataset.object_detection_predicted_y[item.index];
+    const trueY : number[][]  = this.props.dataset.object_detection_true_y[item.index];
+
+    // Draws bounding boxes for each predicted object
+    for (let oidx = 0; oidx < predictedY.length; oidx++) {
+
+      // Creating box region
+      let predObject = predictedY[oidx]
+      let predBox = new RegionData(predObject[1], predObject[2],
```
code-vicar, line 159: Use the static "Build" methods of the RegionData class to create the region data (i.e. BuildRectRegionData). This will ensure the correct region type is set and create the correct position, size and point data from the arguments.
The current code is only creating a region with position/size data and not providing any point data or the region type.

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
```diff
@@ -100,6 +108,93 @@ export class FlyoutObjectDetection extends React.Component<
...
+                                   predObject[3]-predObject[1],
+                                   predObject[4]-predObject[2]);
+
+      // Retrieving label for annotation above the box
+      this.props.dataset.object_detection_predicted_y
+      let className = this.props.dataset.class_names[predObject[0]-1]
+      let confidenceScore = (predObject[5] * 100).toString()
+
+      // Initializing bounding box tag
+      const predTag = new CanvasTools.Core.Tag(className + "(" + confidenceScore + "%)", // Object(95%)
```
natalie-isak, line 169: Do we want to include the object ID?

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
```diff
@@ -100,6 +108,93 @@ export class FlyoutObjectDetection extends React.Component<
...
+                                               new Color(theme.palette.magenta))
```
natalie-isak, line 170: Do you have the hex values for this?

To ensure accessibility, we want to confirm that the contrast between magenta and midblue is very high. The Web Content Accessibility Guidelines (WCAG) recommend a contrast ratio of at least 4.5:1. If you know the hex values I can confirm!

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
```diff
@@ -100,6 +108,93 @@ export class FlyoutObjectDetection extends React.Component<
...
+      const predTagDesc = new CanvasTools.Core.TagsDescriptor([predTag]);
+
+      // Drawing bounding box with vott
+      editor.RM.addRegion(oidx.toString(), predBox, predTagDesc);
+    }
+
+    // Drawing bounding boxes for each ground truth object
+    for (let oidx = 0; oidx < trueY.length; oidx++) {
+
+      // Creating box region
```
natalie-isak, line 180: Question out of curiosity -- if we move the location of the image/change the size of the image, will this code have to change?

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
natalie-isak: Agree with this! Just as a note, there isn't any reason (in the current implementation) that the image source should change. So it might not be necessary to update "component did update"

natalie-isak:
I'm a little confused by the second UI screenshot in the description. Why are there such large bounding boxes? 

(Also, are those UI screenshots created by the changes in this PR? I thought the bboxes were magenta/blue?)

natalie-isak:
I remember in an initial demo of CanvasTools, there was the possibility of adding shading to bboxes. How difficult is this to add, maybe can be a consideration for a future PR?

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
```diff
@@ -201,11 +299,11 @@ export class FlyoutObjectDetection extends React.Component<
                     </Stack>
                   </Stack.Item>
```
natalie-isak, line 300: If I'm remembering correctly, I think the line limit is 300 for files like this :(. Might need to break up this file into separate components (maybe add helper functions to a common file)

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
Advitya17: Sorry, what exactly are you referring to?

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
Advitya17: Thanks for looking into this! Here're the hex values - https://developer.microsoft.com/en-us/fluentui#/styles/web/colors/shared

Let me know any suggestions on the color contrast, thanks a lot!

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
Advitya17: As long as the image is rendered somewhere in this flyout, it shouldn't matter due to the callback that gets the reference to this image element. This'll also enable changing the size of the image and/or rendering it below the flyout metadata for a clearer look.

Advitya17:
> I remember in an initial demo of CanvasTools, there was the possibility of adding shading to bboxes. How difficult is this to add, maybe can be a consideration for a future PR?

@natalie-isak tags support adding a color. Seems like those are reflected in the box as well, but may have to check when the boxes get rendered.

Advitya17:
> I'm a little confused by the second UI screenshot in the description. Why are there such large bounding boxes?
> 
> (Also, are those UI screenshots created by the changes in this PR? I thought the bboxes were magenta/blue?)

The bounding boxes are from the backend itself. Those big white lines are simply tracing where the users hovers on the image (seems like default canvastools behavior). Hence tried to draw a red arrow (as the arrow can't be trapped in a screenshot) to show that's where I was hovering during the screenshot :)

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
natalie-isak: That ratio is actually a bit low: https://coolors.co/contrast-checker/00188f-b4009e

Maybe try something like purpleDark and greenLight? (https://coolors.co/contrast-checker/bad80a-32145a) Or anything with a higher contrast ratio!

Advitya17: Will address in the bbox PR.

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
Advitya17: Will address in bbox PR.

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
```diff
@@ -109,6 +137,19 @@ export class FlyoutObjectDetection extends React.Component<
     const classNames = flyoutStyles();
     const predictedY = getJoinedLabelString(item?.predictedY);
     const trueY = getJoinedLabelString(item?.trueY);
+
+    if (this.state.editorCallback) {
```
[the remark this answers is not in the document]
Advitya17: Update: Was able to add support on the callback, and validated by tracing breakpoints that it only executes once per image.

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
```diff
@@ -100,15 +81,17 @@ export class FlyoutObjectDetection extends React.Component<
       });
     }
   }
+
   public render(): React.ReactNode {
     const { isOpen } = this.props;
-    const item = this.state.item;
+    const item = this.props.item; // state.item is undefined during 1st execution
```
natalie-isak, line 87: https://stackoverflow.com/questions/27991366/what-is-the-difference-between-state-and-props-in-react

Makes sense that you would want props for the 1st execution, but what happens when state.item is updated? Looks like props is immutable, so we might be missing changes to the item during a state change, right?

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
```diff
@@ -201,11 +184,11 @@ export class FlyoutObjectDetection extends React.Component<
                     </Stack>
                   </Stack.Item>
                   <Stack.Item className={classNames.imageContainer}>
-                    <Image
-                      src={`data:image/jpg;base64,${item?.image}`}
-                      className={classNames.image}
-                      imageFit={ImageFit.contain}
-                    />
+                    <Stack.Item id="canvasToolsDiv">
```
natalie-isak, line 187: nit -do we want to abstract these strings to the utils file, and replace with constants?

natalie-isak (approved):

libs/interpret-vision/.eslintrc.json
```diff
@@ -13,6 +13,12 @@
     {
       "files": ["*.js", "*.jsx"],
       "rules": {}
+    },
+    {
+      "files": ["FlyoutObjectDetection.tsx"],
```
imatiach-msft, line 18: can you refactor out the file instead to keep to the 300 line limit?

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetectionUtils.tsx
```diff
@@ -0,0 +1,51 @@
...
+export function loadImageFromBase64(
+  base64String: string,
+  editor: Editor
+): Promise<HTMLImageElement> {
+  return new Promise((resolve, reject) => {
+    const image = new Image();
+    image.addEventListener("load", (e) => {
+      editor.addContentSource(e.target as HTMLImageElement);
+      editor.AS.setSelectionMode(2);
+      resolve(image);
```
imatiach-msft, line 44: great to see this worked!

imatiach-msft (approved):
signing off except for the 300 line limit issue

imatiach-msft (approved):
re-approving

PR #2099, opened 2023-06-06
Fix dimension mismatch error when dropped_features contains cat_features

tongyu-microsoft:
This PR fixes the bug when there are categorical features in dropped_features for feature importance component and casual component.

## Description

The fix is to add a `get_categorical_features_after_drop()` method to remove dropped_features if there are any categorical features that are passed to RAIInsights().

In the example below, we drop ['education-num' and 'education'] features, 'education' is a categorical feature.
The expected behavior: For feature importances, we don't show the dropped feature, and we make correct /predict calls. For causal, we don't show dropped features and we make correct /causal calls.
metadataCat

## Checklist


responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -259,6 +259,14 @@ def __init__(self,
         self._initialize_managers()
         self._try_add_data_balance()
 
+    def get_categorical_features_after_drop(self):
```
gaugup, line 262: could add unit test for this function?

gaugup (changes requested):

responsibleai/responsibleai/rai_insights/rai_insights.py
tongyu-microsoft: Added tests

gaugup (approved):

tongyu-microsoft:
this PR passed cd-build pipeline here: https://github.com/microsoft/responsible-ai-toolbox/actions/runs/5193855143
and the only commit after that succeeded pipeline is for unit tests. So this PR should be good to be merged once all other pipelines pass.

PR #2104, opened 2023-06-08
Create python-package-conda.yml

Ninja-man07:
package main
type CategorySummary struct {
    Title       string
    Tasks       int
    AvgValue    float64
}
func createTables(db *sql.DB) {
    db.Exec("CREATE TABLE tasks (id INTEGER PRIMARY KEY, title TEXT, value INTEGER, category TEXT)")
}
func createCategorySummaries(db *sql.DB) ([]CategorySummary, error) {
    var summaries []CategorySummary
    rows, err := db.Query("SELECT category, COUNT(category), AVG(value) FROM tasks GROUP BY category")
    if err != nil {
        return nil, err
    }
    defer rows.Close()
    for rows.Next() {
        var summary CategorySummary
        err := rows.Scan(&summary.Title, &summary.Tasks, &summary.AvgValue)
        if err != nil {
            return nil, err
        }
        summaries = append(summaries, summary)
    }
    return summaries, nil
}
Copilot

## Checklist

### DIFF
--- a/.github/workflows/python-package-conda.yml
+++ b/.github/workflows/python-package-conda.yml
@@ -1,0 +1,34 @@
+name: Python Package using Conda
+
+on: [push]
+
+jobs:
+  build-linux:
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 5
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.10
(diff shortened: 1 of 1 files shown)


imatiach-msft:
closing old PR since there was no response, please let us know if this should be reopened

PR #2121, opened 2023-06-19
Add loading spinner for causal policy tab

gaugup:
## Description

For large data the policy computation takes time. Hence, adding the loading spinner. 

image

## Checklist

### DIFF
--- a/libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAnalysisView.tsx
+++ b/libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAnalysisView.tsx
@@ -180,7 +180,7 @@ export class CausalAnalysisView extends React.PureComponent<
     return globalCausalEffects?.sort((d1, d2) => d2.point - d1.point);
   }
 
-  private getPolicyElement(): any {
+  private getPolicyElement(): React.ReactNode {
     if (!this.state.isCausalPolicyDataLoading) {
       return (
         <TreatmentView
--- a/libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAnalysisView.tsx
+++ b/libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAnalysisView.tsx
@@ -10,8 +10,10 @@ import {
   ICausalPolicy,
   ITelemetryEvent,
   ModelAssessmentContext,
-  ifEnableLargeData
+  ifEnableLargeData,
+  LoadingSpinner
 } from "@responsible-ai/core-ui";
+import { localization } from "@responsible-ai/localization";
 import React from "react";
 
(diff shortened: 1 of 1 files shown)


libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAnalysisView.tsx
```diff
@@ -171,4 +179,16 @@ export class CausalAnalysisView extends React.PureComponent<
...
+  private getPolicyElement(): any {
+    if (!this.state.isCausalPolicyDataLoading) {
+      return (
+        <TreatmentView
+          data={this.state.currentGlobalCausalPolicy}
+          telemetryHook={this.props.telemetryHook}
+        />
+      );
+    }
+    return <LoadingSpinner label={localization.Common.loading} />;
```
vinuthakaranth, line 192: nit: we can have conditional operator here
`return !this.state.isCausalPolicyDataLoading ? (
      <TreatmentView
        data={this.state.currentGlobalCausalPolicy}
        telemetryHook={this.props.telemetryHook}
      />
    ): <LoadingSpinner label={localization.Common.loading} />;`

libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAnalysisView.tsx
```diff
@@ -171,4 +179,16 @@ export class CausalAnalysisView extends React.PureComponent<
   ): ICausalAnalysisSingleData[] | undefined {
     return globalCausalEffects?.sort((d1, d2) => d2.point - d1.point);
   }
+
+  private getPolicyElement(): any {
```
vinuthakaranth, line 183: any can be replaced by React.ReactNode

vinuthakaranth:
Design per se it would be good to have loading part just for the area below the dropdown instead of all tab reloading.
image

vinuthakaranth:
> Design per se it would be good to have loading part just for the area below the dropdown instead of all tab reloading. image

discussed offline. It loads just once while loading tab and not while the selection of dropdown changes.

vinuthakaranth (approved):

libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAnalysisView.tsx
gaugup: The if statement has the same effect. Will leave the logic as is.

libs/causality/src/lib/CausalAnalysisDashboard/Controls/CausalAnalysisView/CausalAnalysisView.tsx
gaugup: done

PR #2127, opened 2023-06-21
add validation target column type in the classification scenario

hawestra:
Add validation check to ensure that the target column is not continuous (ie of type float) for multiclass classification. This is to address the following bug: [Bug 2472415](https://msdata.visualstudio.com/Vienna/_workitems/edit/2472415): [Live site bug] raise ValueError("Unknown label type: %r" % y_type) in 0.8.0 which resulted from a sklearn error.
Add unit tests for this scenario

## Checklist


romanlutz (approved):

gaugup (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -566,6 +566,13 @@ def _validate_rai_insights_input_parameters(
                         f"Error finding unique values in column {column}. "
                         "Please check your test data.")
 
+        # Validate that the target column isn't continuous if the
+        # user is running classification scenario
+        if (task_type == ModelTask.CLASSIFICATION and
+                pd.api.types.is_float_dtype(train[target_column].dtype)):
+            raise UserConfigValidationException(
```
imatiach-msft, line 573: It could be float dtype but whole numbers/counting numbers though - eg 1f, 2f, etc.  This would fail but it shouldn't then.  I would be careful with this type of check.  Also, I've seen in the past customers with a number "standing in" for a class - like say 2.2 and 3.3 actually being a binary classification scenario which might have some industrial or chemical significance for the label values.  I'm not completely convinced this is the best way to handle all such scenarios.

imatiach-msft (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
hawestra: I added a bit in the comments about the sklearn error that's thrown plus adding a check for whether the float can be converted to an int (and if it can't throw the error); I see your point with float numbers not necessarily indicating regression, but seems like this is the behavior sklearn has so I was trying to duplicate their logic in ours and just add a better error message. Let me know if you think there's a better approach

PR #2143, opened 2023-06-26
raise error instead of warning when a user has missing data and add c…

hawestra:
1. Raise user validation error instead of warning when a user's test data has missing data
2. Add this same validation check for a user's train data
3. Add tests

*To address this bug: [Bug 2477311](https://msdata.visualstudio.com/Vienna/_workitems/edit/2477311): [Live site bug]Add length validation for test / train datasets

## Checklist


romanlutz (approved):
Not sure why this was ever a warning but this looks great, and thanks for adding tests 🙂👍🏻

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -586,14 +586,22 @@ def _validate_rai_insights_input_parameters(
...
             if np.any(test[feature].isnull()):
                 list_of_feature_having_missing_values.append(feature)
         if len(list_of_feature_having_missing_values) > 0:
-            warnings.warn(
+            raise UserConfigValidationException(
                 f"Features {list_of_feature_having_missing_values} "
                 "have missing values in test data")
+        for feature in train.columns.tolist():
+            if np.any(train[feature].isnull()):
+                list_of_feature_having_missing_values.append(feature)
```
gaugup, line 600: This looks like repeated code. Could you create a common function for this?

imatiach-msft (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
gaugup: Also, looks like you are using the same list for both train and test. Though functionally the logic is correct it can easily break. Creating a common function will resolve this problem.

responsibleai/tests/rai_insights/test_rai_insights_validations.py
```diff
@@ -57,7 +57,39 @@ def test_missing_data_warnings(self):
...
+            RAIInsights(
+                model=model,
+                train=train,
+                test=test,
+                target_column='Target',
+                task_type='classification')
+        assert "Features ['Target'] have missing values in " + \
+            "test data" in str(ucve.value)
+
+    def test_missing_train_data(self):
```
gaugup, line 83: Maybe rename to test_missing_train_data_warnings for consistency with the above test function test_missing_test_data_warnings

gaugup (changes requested):

gaugup (commented):
Why do we need the check on train data if it has missing values? We compute insights on test data only.

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -586,14 +586,22 @@ def _validate_rai_insights_input_parameters(
...
+        if len(list_of_feature_having_missing_values) > 0:
+            raise UserConfigValidationException(
```
imatiach-msft, line 602: Some ML models, like LightGBM, can actually handle missing values and predict on them - and certainly many pipelines can if they have some imputer as one of the steps (and we can take a whole pipeline as input).  I've had customers in the past complain that we should just be able to handle missing values.  I guess I'm ok with this change but I'm really worried what we will have to tell our customers when we get such issues - often they just want everything to 'just work' and they are reluctant to take a bunch of extra steps just to get something to work in our dashboard when their pipeline already works end to end on their data with missing values.

imatiach-msft (approved):

responsibleai/responsibleai/rai_insights/rai_insights.py
kicha0: @vinuthakaranth for the bug: https://msdata.visualstudio.com/Vienna/_workitems/edit/2349068 you might want to check here and see if this handles the case of NaN's. The error comes from this https://scikit-learn.org/stable/modules/generated/sklearn.utils.assert_all_finite.html checking for NaNs and the null check here may not be covering that.

vinuthakaranth: @Hannah, can you please check if NaN case is also handled with the same fix as it is the similar error? If not, I'll reopen the bug and follow up.

hawestra:
> Why do we need the check on train data if it has missing values? We compute insights on test data only.

looking at the error analysis manager constructor, we actually use the test data there : 
image
image


but this highlight @imatiach-msft  point, that if we fail because test data is missing but the user doesn't use error analysis, then we're failing their run for no reason. I'm considering moving this test data check to the error analysis code. @imatiach-msft, @gaugup what are your thoughts ?

responsibleai/responsibleai/rai_insights/rai_insights.py
hawestra: makes sense; tagged you in a comment below!

responsibleai/responsibleai/rai_insights/rai_insights.py
hawestra: @vinuthakaranth it does! 
@gaugup see my comment below about moving this check for test data and let me know your thoughts

responsibleai/tests/rai_insights/test_rai_insights_validations.py
hawestra: removed the "_warnings" because it's not a warning anymore

PR #2150, opened 2023-06-27
Delete the downloaded zip and folder in `retrieve_dataset`

gaugup:
## Description

Delete the downloaded zip and folder in `retrieve_dataset`

## Checklist

### DIFF
--- a/rai_test_utils/rai_test_utils/utilities/utils.py
+++ b/rai_test_utils/rai_test_utils/utilities/utils.py
@@ -2,7 +2,10 @@
 # Licensed under the MIT License.
 
 import os
+import shutil
 import uuid
+
+DOWNLOADED_DATASET_DIR = 'datasets.4.27.2021'
 
 
 def is_valid_uuid(id: str):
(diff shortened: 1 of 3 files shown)


rai_test_utils/rai_test_utils/utilities/utils.py
```diff
@@ -2,8 +2,11 @@
 # Licensed under the MIT License.
 
 import os
+import shutil
 import uuid
 
+DOWNLOADED_DATASET_DIR = 'datasets.4.27.2021'
```
romanlutz, line 8: That seems oddly specific. Isn't it going to change?

rai_test_utils/rai_test_utils/utilities/utils.py
```diff
@@ -29,7 +32,7 @@ def retrieve_dataset(dataset, **kwargs):
     :rtype: object
     """
     # if data not extracted, download zip and extract
-    outdirname = 'datasets.4.27.2021'
+    outdirname = DOWNLOADED_DATASET_DIR
     if not os.path.exists(outdirname):
```
imatiach-msft, line 36: if the dataset is downloaded we don't re-download it here... we just re-use the downloaded one

rai_test_utils/rai_test_utils/utilities/utils.py
imatiach-msft: we use a specific zip folder with the date it was created for these datasets (that are put on our blob storage for tests)

romanlutz: Right, so if the dataset changes this need to be updated. How will we remember?

PR #2157, opened 2023-06-29
add device parameter to responsibleai_vision package and pass to model wrappers

imatiach-msft:
## Description

add device parameter to responsibleai_vision package and pass to model wrappers

## Checklist

### DIFF
--- a/responsibleai_vision/responsibleai_vision/managers/explainer_manager.py
+++ b/responsibleai_vision/responsibleai_vision/managers/explainer_manager.py
@@ -17,6 +17,7 @@ import numpy as np
 import pandas as pd
 import shap
 from ml_wrappers import wrap_model
+from ml_wrappers.common.constants import Device
 from ml_wrappers.model.image_model_wrapper import (MLflowDRiseWrapper,
                                                    PytorchDRiseWrapper)
 from PIL import Image, ImageDraw, ImageFont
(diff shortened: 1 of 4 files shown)


responsibleai_vision/responsibleai_vision/managers/explainer_manager.py
```diff
@@ -249,16 +257,22 @@ def compute_single_explanation(self,
...
+                # get_drise_saliency_map only recognizes GPU and CPU
+                if device == Device.AUTO.value:
+                    device = None
                 fl, _, _, = get_drise_saliency_map(img,
                                                    self._model,
                                                    len(self._classes),
                                                    savename=str(index),
                                                    nummasks=self._num_masks,
                                                    maskres=mask_res_tuple,
+                                                   devicechoice=device,
```
Advitya17, line 275: What if `None` is passed here for device after line 268?

Advitya17 (approved):

responsibleai_vision/responsibleai_vision/managers/explainer_manager.py
imatiach-msft: see https://github.com/microsoft/vision-explanation-methods/blob/657fe461264a596c94378a6a1174a3c21d0757fe/python/vision_explanation_methods/DRISE_runner.py#L117

it just uses same logic as auto:

torch.device('cuda' if torch.cuda.is_available() else 'cpu')

PR #2158, opened 2023-06-29
[Object Detection] Cache support for Model Overview Metrics

Advitya17:
This adds support for caching object detection metrics within a dashboard session, such that the metrics will be automatically displayed if they had been computed for the same set of inputs previously.

This is also reflected with adding new cohorts:

image

image

## Checklist

### DIFF
--- a/apps/widget/src/app/ModelAssessment.tsx
+++ b/apps/widget/src/app/ModelAssessment.tsx
@@ -60,7 +60,7 @@ export class ModelAssessment extends React.Component<IModelAssessmentProps> {
         selectionIndexes: number[][],
         aggregateMethod: string,
         className: string,
-        iouThresh: number,
+        iouThreshold: number,
         objectDetectionCache: Map<string, [number, number, number]>,
         abortSignal: AbortSignal
       ): Promise<any[]> => {
@@ -70,7 +70,7 @@ export class ModelAssessment extends React.Component<IModelAssessmentProps> {
             selectionIndexes,
             aggregateMethod,
             className,
-            iouThresh,
+            iouThreshold,
             objectDetectionCache
           ],
           "/get_object_detection_metrics",
--- a/apps/widget/src/app/ModelAssessment.tsx
+++ b/apps/widget/src/app/ModelAssessment.tsx
@@ -11,47 +11,16 @@ import {
   ILocalExplanations,
   parseFeatureFlights
 } from "@responsible-ai/core-ui";
-import {
-  ModelAssessmentDashboard,
-  IModelAssessmentData,
-  IModelAssessmentDashboardProps
-} from "@responsible-ai/model-assessment";
+import { ModelAssessmentDashboard } from "@responsible-ai/model-assessment";
 import React from "react";
 
 import { callFlaskService } from "./callFlaskService";
-import { IAppConfig } from "./config";
+import { CallbackType, IModelAssessmentProps } from "./ModelAssessmentUtils";
 
-interface IModelAssessmentProps {
-  config: IAppConfig;
-  modelData: IModelAssessmentData;
-}
 export class ModelAssessment extends React.Component<IModelAssessmentProps> {
   public render(): React.ReactNode {
-    const callBack: Pick<
-      IModelAssessmentDashboardProps,
-      | "requestExp"
-      | "requestObjectDetectionMetrics"
-      | "requestPredictions"
-      | "requestQuestionAnsweringMetrics"
-      | "requestDebugML"
-      | "requestMatrix"
-      | "requestImportances"
-      | "requestCausalWhatIf"
-      | "requestBoxPlotDistribution"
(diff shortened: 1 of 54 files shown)


Advitya17:
TODO: Remove console logs & fix lints

libs/core-ui/src/lib/util/ObjectDetectionStatisticsUtils.ts
```diff
@@ -15,13 +15,29 @@ export enum ObjectDetectionMetrics {
...
 
+    const key: [number[], string, string, number] = [
+      selectionArray,
+      objectDetectionInputs[0],
+      objectDetectionInputs[1],
+      objectDetectionInputs[2]
+    ];
+    const value = objectDetectionCache.get(key.toString());
+    console.log('retrieving key');
+    console.log(key.toString());
```
imatiach-msft, line 37: should the console.log calls be removed here?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -647,6 +656,18 @@ export class ModelOverview extends React.Component<
...
 
+            const key: modelOverviewKeyType = [
+              selectionIndexes[cohortIndex], 
+              this.state.aggregateMethod, 
+              this.state.className, 
+              this.state.iouThresh
+            ];
+            if (!this.objectDetectionCache.has(key.toString())) {
+              this.objectDetectionCache.set(key.toString(), [meanAveragePrecision, averagePrecision, averageRecall]);
+              console.log('set new value');
```
imatiach-msft, line 667: remove console.log calls here?

imatiach-msft (approved):

libs/core-ui/src/lib/util/ObjectDetectionStatisticsUtils.ts
Advitya17: Yes I put in a separate comment for myself above :D

Advitya17:
TODO: Cohort creation test & attach snapshot on PR.

apps/widget/src/app/ModelAssessment.tsx
```diff
@@ -61,11 +61,18 @@ export class ModelAssessment extends React.Component<IModelAssessmentProps> {
...
         abortSignal: AbortSignal
       ): Promise<any[]> => {
         return callFlaskService(
           this.props.config,
-          [selectionIndexes, aggregateMethod, className, iouThresh],
+          [
+            selectionIndexes,
+            aggregateMethod,
+            className,
+            iouThresh,
```
romanlutz, line 73: Is this a threshold? Can we spell it out?

libs/core-ui/src/lib/util/ObjectDetectionStatisticsUtils.ts
```diff
@@ -15,13 +15,26 @@ export enum ObjectDetectionMetrics {
...
+  selectionIndexes: number[][],
+  objectDetectionCache: Map<string, [number, number, number]>,
+  objectDetectionInputs: [string, string, number]
 ): ILabeledStatistic[][] => {
   return selectionIndexes.map((selectionArray) => {
     const count = selectionArray.length;
 
+    const key: [number[], string, string, number] = [
+      selectionArray,
+      objectDetectionInputs[0],
```
romanlutz, line 31: Should it be a well-defined object (class instance) so that it's clear which is which? Right now, it's possible to return something in wrong order from the SDK without noticing here.

romanlutz (approved):

libs/core-ui/src/lib/util/ObjectDetectionStatisticsUtils.ts
Advitya17: @romanlutz Currently all the inputs are stitched and serialized to a string key of the format "<cohort indices>,agg method, class name, iou threshold"

If I define a new class with those 4 stored in instance variables, I may still end up manually specifying the order like `[object.cohort_indices, object.agg_method,...].toString()`. Perhaps this'll be more resilient that passing in the order of variables multiple times in places of typed variables.

Let me know your suggestion, thanks a lot!

romanlutz: It comes from the SDK, right? So you can use a well-defined object there, too 🙂 or I'm misunderstanding where this order comes from

romanlutz: Yeah I must have been thinking of a different PR.... this is a cache lookup so it's local enough that I'm fine with this.

Advitya17: Ahh I see, thanks for reviewing this!

PR #2166, opened 2023-07-04
[Question Answering] Cache support for Model Overview Metrics

Advitya17:
This adds support for caching object detection metrics within a dashboard session, such that the metrics will be automatically displayed if they had been computed for the same set of inputs previously.

This is also reflected with adding new cohorts:

image

## Checklist


libs/core-ui/src/lib/util/ImageStatisticsUtils.ts
```diff
@@ -1,6 +1,13 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+import { localization } from "@responsible-ai/localization";
+
+import {
+  ILabeledStatistic,
```
imatiach-msft, line 7: I remember there was some circular dependency issue with moving ILabeledStatistic here like this - hopefully I'm wrong and the builds will just pass

imatiach-msft (approved):

romanlutz (approved):

PR #2170, opened 2023-07-07
Agemawat/od cache ext

Advitya17:
This PR extends the OD cache support from #2158 as the torchmetrics API calculates the metrics for all classes at once, which are now all cached as part of this PR.

## Checklist

### DIFF
--- a/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
+++ b/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
@@ -648,27 +648,43 @@ export class ModelOverview extends React.Component<
           new AbortController().signal
         )
         .then((result) => {
-          // Assumption: the lengths of `result` and `selectionIndexes` are the same.
+          let [allCohortMetrics, cohortClasses] = result;
+
+          // Assumption: the lengths of `allCohortMetrics` and `selectionIndexes` are the same.
           const updatedMetricStats: ILabeledStatistic[][] = [];
 
           for (const [
             cohortIndex,
-            [meanAveragePrecision, averagePrecision, averageRecall]
-          ] of result.entries()) {
+            cohortMetrics
+          ] of allCohortMetrics.entries()) {
             const count = selectionIndexes[cohortIndex].length;
 
-            const key: [number[], string, string, number] = [
-              selectionIndexes[cohortIndex],
-              this.state.aggregateMethod,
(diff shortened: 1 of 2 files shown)


libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -657,27 +657,47 @@ export class ModelOverview extends React.Component<
           new AbortController().signal
         )
         .then((result) => {
-          // Assumption: the lengths of `result` and `selectionIndexes` are the same.
+          const [allCohortMetrics, cohortClasses] = result;
```
imatiach-msft, line 660: it looks like there is a breaking change here in the UI - how can we ensure that the new UI with the older saved dashboards/python dependencies still works here and doesn't fail?

Advitya17: Perhaps the env can be rebuilt which installs the latest version of `responsibleai_vision` and the DPv2 components can rerun. Have addressed this in more detail outside this PR. This optimization is almost a necessary improvement because of the way the torchmetrics API is invoked, and considering the computational cost of calculating object detection metrics. Let me know your suggestions, thanks a lot!

Advitya17: Update: Confirmed it's okay to go with this change.

imatiach-msft (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverview.tsx
```diff
@@ -657,27 +657,47 @@ export class ModelOverview extends React.Component<
...
+                const [mAP, aP, aR] = cohortMetric;
+
+                const key: [number[], string, string, number] = [
+                  selectionIndexes[cohortIndex],
+                  this.state.aggregateMethod,
+                  cohortClasses[i],
+                  this.state.iouThreshold
+                ];
+                if (!this.objectDetectionCache.has(key.toString())) {
+                  this.objectDetectionCache.set(key.toString(), [mAP, aP, aR]);
```
tongyu-microsoft, line 690: we set the cache here, where do we use it? when do we .get()?

tongyu-microsoft (approved):

PR #2171, opened 2023-07-07
AUC chart

hawestra:
## Description
first draft to add AUC chart to model overview; still needs PM input

## Checklist

image

- [ ] TODO: I have added e2e tests for all UI changes.
- [ ] TODO: Documentation was updated if it was needed.


libs/core-ui/src/lib/util/calculateAUCData.ts
```diff
@@ -0,0 +1,142 @@
...
+      }
+    }
+    addROCPoint(
+      truePositives,
+      falsePositives,
+      trueNegatives,
+      falseNegatives,
+      rocData
+    );
+    truePositives = falsePositives = trueNegatives = falseNegatives = 0;
```
romanlutz, line 77: Can we just set all of these inside the for-loop instead of reusing them?

libs/core-ui/src/lib/util/calculateAUCData.ts
```diff
@@ -0,0 +1,142 @@
...
+}
+
+export function calculatePerClassROCData(
+  probabilityY: number[],
+  binY: number[]
+): IROCData {
+  const rocData: IROCData = {
+    points: []
+  };
+  const thresholds = range(0, 1, 0.01);
```
romanlutz, line 47: That's not quite right, is it?
For example, if I have two points within a tick (let's say at 0.001 and 0.004) then none of my thresholds covers the case where they are on opposite sides of the decision boundary. This could very well be the best FPR/TPR combination....

So I would order the values first, and then always take thresholds as the middle between two values, e.g.,

values: [1,5,4,2,3]
ordered: [1,2,3,4,5]
thresholds: [-inf, 1.5, 2.5, 3.5, 4.5, inf]
so total of 6 points for ROC, but then you still need to calculate the convex hull of those points if I'm not mistaken. How are you avoiding hitting points below the convex hull btw?

libs/core-ui/src/lib/components/OverallMetricChartUtils.ts
```diff
@@ -220,6 +220,7 @@ export function generateMetricsList(
   if (!chartProps) {
     return [];
   }
+  console.log(yAxisProperty);
```
romanlutz, line 223: Remove?

libs/core-ui/src/lib/util/calculateAUCData.test.ts
```diff
@@ -0,0 +1,148 @@
...
+      [1, 0],
+      [0, 1],
+      [1, 0]
+    ]);
+  });
+});
+describe("Test calculatePerClassROCData", () => {
+  it("generate x,y data corresponding to fpr and tpr respectively", () => {
+    const result = calculatePerClassROCData(
+      [0.33, 0.32, 0.34, 0.29, 0.12, 0.41, 0.4, 0.39],
```
romanlutz, line 40: If you add zeros after the comma to get all the points closer together (0.032, 0.033, 0.034, for example), then it will not do it right, will it?

libs/core-ui/src/lib/util/calculateAUCData.ts
```diff
@@ -0,0 +1,160 @@
...
+// based on https://msdata.visualstudio.com/Vienna/_git/AzureMlCli?path=/src/azureml-metrics/azureml/metrics/_classification.py&version=GBmaster
+export function calculateAUCData(dataset: IDataset): SeriesOptionsType[] {
+  if (!dataset.probability_y || !dataset.class_names) {
+    // TODO: show warning message ?
+    return [...getStaticROCData()];
+  }
+
+  // TODO: for some reason in the adult census data, classnames doesn't match the true_y data
+  // need to confirm how this data aligns. temporarily hard code classnames to 0 and 1 for now
+  const cNames = [0, 1];
```
romanlutz, line 141: This should only run for binary classification, right? So we probably need a check somewhere to disable the component otherwise.

For multiclass one could do one vs all but I don't know if anyone wants that.

libs/localization/src/lib/en.json
```diff
@@ -1828,6 +1830,7 @@
       "regressionDistributionPivotItem": "Target distribution",
       "metricsVisualizationsPivotItem": "Metrics visualizations",
       "confusionMatrixPivotItem": "Confusion matrix",
+      "AUCPivotItem": "AUC Chart",
```
romanlutz, line 1833: This is fine, but we'll probably need a little explainer somewhere. Especially because it's an acronym

libs/core-ui/src/lib/util/calculateAUCData.ts
```diff
@@ -0,0 +1,160 @@
...
+  ];
+}
+
+export function calculatePerClassROCData(
+  probabilityY: number[],
+  binY: number[]
+): IROCData {
+  const thresholds = probabilityY.sort();
+  const points: IPoint[] = [];
+  for (const threshold of thresholds) {
```
romanlutz, line 46: 100*n effort, still linear

My suggestion would be to sort (nlogn) and then step through (n) and build the entire curve in one walkthrough because with ordering you know exactly how it changes the percentage when increasing thresholds 🙂 This still needs the convex hull cleanup, of course, because some points are below others and irrelevant.

I suspect since our n is quite small the log n factor won't really hurt.

The convex hull cleanup would have more points, though (about n instead of 101), so that could take longer.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverviewChartPivot.tsx
```diff
@@ -192,6 +194,25 @@ export class ModelOverviewChartPivot extends React.Component<
...
             </PivotItem>
           )}
+        {!ifEnableLargeData(this.context.dataset) &&
+          (IsBinary(this.context.modelMetadata.modelType) ||
+            IsMulticlass(this.context.modelMetadata.modelType)) && (
+            // isFlightActive(
+            //   aucChartExperienceFlight,
+            //   this.context.featureFlights
+            // ) && (
+            // TODO: support multiclass: IsMulticlass(this.context.modelMetadata.modelType)) && (
```
romanlutz, line 204: I guess you would need a drop-down to select the one vs all class and then generate the 0s and 1s based on that.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/getAUCChartOptions.ts
```diff
@@ -0,0 +1,51 @@
...
+import { ITheme } from "@fluentui/react";
+import { IHighchartsConfig } from "@responsible-ai/core-ui";
+import { SeriesOptionsType } from "highcharts";
+
+export function getAUCChartOptions(
+  data: SeriesOptionsType[],
+  theme?: ITheme
+): IHighchartsConfig {
+  // Open questions:
+  // 1. select cohort ?
```
romanlutz, line 13: Can you elaborate?

romanlutz (commented):
Perhaps on the wrong track here, but it seems like we might miss something by just looking at 100 thresholds vs all possible thresholds between values (after sorting).

libs/core-ui/src/lib/util/calculateAUCData.ts
hawestra: in the studio, there's an auc chart for multiclass so i assumed we'd want that (binarizeData is supposed to handle this case)

libs/localization/src/lib/en.json
hawestra: will discuss with Minsoo!

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverviewChartPivot.tsx
hawestra: i think the idea would be to show a line for each class; does that make sense ? the user can optionally disable the line's visibility (see the roc chart here [https://ml.azure.com/experiments/id/e01fdc58-9702-4fb8-add9-1a5d09b21058/runs/iris-automl-1_24?wsid=/subscriptions/e0fd569c-e34a-4249-8c24-e8d723c7f054/resourceGroups/hawestra-rg/providers/Microsoft.MachineLearningServices/workspaces/hawestra-ws&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#metrics](url)

hawestra:
> Perhaps on the wrong track here, but it seems like we might miss something by just looking at 100 thresholds vs all possible thresholds between values (after sorting).

originally, i was looking at 100 thresholds because this is how the metrics package does it but i changed that to look at the y probabilities themselves and go through each of these as the threshold

libs/core-ui/src/lib/util/calculateAUCData.ts
romanlutz: That's probably 1 vs all, so that makes sense.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ModelOverviewChartPivot.tsx
romanlutz: That will be tricky if you have one line up per cohort as well. Or are we showing just for current cohort?

PR #2181, opened 2023-07-14
Require pillow>=10.0.0, catch drise failure due to low mask_res or num_masks

romanlutz:
## Description

In telemetry, we found that two exceptions were thrown. Firstly, drise throws `ValueError` when `num_masks` or `mask_res` are too low. This is caught by the try-except block and the `_get_fail_str` method subsequently runs into an `AttributeError` because it doesn't have a `textsize` method. This is due to a breaking change in v10.0.0 of Pillow.

This PR therefore
- requires Pillow>=10.0.0 explicitly so that dependencies can't later accidentally downgrade us. We are already using Pillow==10.0.0 in our environments, so this isn't necessarily a change, just insurance.
- uses the new method `textbbox` instead of `textsize` from Pillow which avoids the second exception
- catches the `ValueError` from drise and provides a corresponding user error with informative message to the user.

A simple way to reproduce this is to take our dpv2 test cases and run `test_object_detection_pipeline_rai.yaml` with `num_masks=1` and `max_evals=10` (to make it faster).

## Checklist


imatiach-msft (approved):

responsibleai_vision/responsibleai_vision/managers/explainer_manager.py
```diff
@@ -577,11 +585,13 @@ def _is_object_detection_task(self):
         return self._task_type == ModelTask.OBJECT_DETECTION
 
     def _get_fail_str(self):
-        fail = Image.new('RGB', (100, 100))
+        fail = Image.new('RGB', (250, 250))
```
Advitya17, line 588: Why is this changed/needed?

Advitya17 (approved):

ziqi-ma (approved):

responsibleai_vision/responsibleai_vision/managers/explainer_manager.py
romanlutz: Because the text is way longer than 100 and it's just getting cut off. That makes this entire method pointless. With 250 I found it fits comfortably without wasting much space.

romanlutz: I wasn't sure if it's required to be square so I kept that.

PR #2182, opened 2023-07-17
Vision Data Analysis / Data Balance E2E Tests

Advitya17:
Added Data Analysis / Data Balance tests for the vision dashboard.

## Checklist

### DIFF
--- a/libs/e2e/src/lib/describer/modelAssessmentVision/dataBalance/describeAggregateBalanceMeasures.ts
+++ b/libs/e2e/src/lib/describer/modelAssessmentVision/dataBalance/describeAggregateBalanceMeasures.ts
@@ -1,0 +1,43 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { Locators } from "../Constants";
+import { IModelAssessmentData } from "../IModelAssessmentData";
+
+export function describeAggregateBalanceMeasures(
+  dataShape: IModelAssessmentData
+): void {
+  describe("Aggregate balance measures", () => {
+    if (!dataShape.dataBalanceData?.aggregateBalanceMeasuresComputed) {
+      it("should render no measures computed message", () => {
+        cy.get(
+          `${Locators.AggregateBalanceMeasures} #MissingParameterPlaceHolder`
+        ).contains(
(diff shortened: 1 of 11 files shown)


libs/e2e/src/lib/describer/modelAssessmentVision/dataBalance/describeAggregateBalanceMeasures.ts
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { Locators } from "../Constants";
+import { IModelAssessmentData } from "../IModelAssessmentData";
+
+export function describeAggregateBalanceMeasures(
```
imatiach-msft, line 7: this component actually works for the vision scenario? wow!  I've never actually tried it out with text or vision dashboards.

libs/e2e/src/lib/describer/modelAssessmentVision/dataBalance/describeFeatureBalanceMeasures.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    it("should have a header with clickable info button", () => {
+      cy.get(Locators.FeatureBalanceMeasuresHeader).contains(
+        "Feature balance measures"
+      );
+      cy.get(`${Locators.FeatureBalanceMeasuresHeader} button`).should("exist");
+    });
+
+    it("should have dropdowns for label, feature, and measure", () => {
```
imatiach-msft, line 29: I think a lot of this test code is duplicated - is there some way for us to re-use the common code instead of duplicating all of these files/test code everywhere?  The less code we have to maintain the easier it will be for us to update it.  Also we will have fewer issues.

imatiach-msft (approved):

RubyZ10 (approved):

romanlutz (approved):

libs/e2e/src/lib/describer/modelAssessmentVision/dataBalance/describeAggregateBalanceMeasures.ts
Advitya17: Yeah, and that seems to make sense as the image dashboard currently has 3 components, 2 of which (Model Overview & Data Analysis) share the same UI elements for the most part. The 1 component that's different is Vision Data Explorer for which new UI tests will be needed.

romanlutz:
Why did you close it? @Advitya17

PR #2187, opened 2023-07-24
add conversion to numpy array

hawestra:
## Description

If the user brings their own model, the result of predictions isn't necessarily a numpy array which is what we expect. The conversion to a numpy array was missing in text classficication, image classfication, and object detection which resulted in a RangeIndex error. This address this bug: [Bug 2532332](https://msdata.visualstudio.com/Vienna/_workitems/edit/2532332): RangeIndex Error when users bring their own model with text and vision

## Checklist

### DIFF
--- a/responsibleai_text/responsibleai_text/managers/error_analysis_manager.py
+++ b/responsibleai_text/responsibleai_text/managers/error_analysis_manager.py
@@ -72,7 +72,8 @@ class WrappedIndexPredictorModel:
         if self.task_type in classif_tasks:
             dataset = self.dataset.iloc[:, 0].tolist()
             self.predictions = self.model.predict(dataset)
-            self.predictions = np.array(self.predictions)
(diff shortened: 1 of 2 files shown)


responsibleai_text/responsibleai_text/managers/error_analysis_manager.py
```diff
@@ -72,6 +72,8 @@ def __init__(self, model, dataset, is_multilabel, task_type, classes=None):
         if self.task_type in classif_tasks:
             dataset = self.dataset.iloc[:, 0].tolist()
             self.predictions = self.model.predict(dataset)
+            if not isinstance(self.predictions, np.ndarray):
+                self.predictions = np.array(self.predictions)
```
imatiach-msft, line 76: this should instead be taken care of in ml_wrappers wrap_model function... similar to tabular case, we shouldn't be adding any extra conversion logic here

PR #2025, opened 2023-04-05
[Model Overview] Object Detection Metrics support

Advitya17:
This PR invokes the Flask endpoint and retrieves the Model Overview Metrics value from the backend.

While selecting a widget option, the values are set to 0 and a re-render is triggered which updates the values on the dashboard.

Currently, a dummy value (42) is being used in the backend to test the engineering changes. The actual metrics will be surfaced after a corresponding PR is merged in the backend.

## Description

image

image

## Checklist

- [x] I have added screenshots above for all UI changes.
- [ ] I have added e2e tests for all UI changes.
- [ ] Documentation was updated if it was needed.

PR #2201, opened 2023-08-02
Fix heatmap by removing erroneous color configuration

romanlutz:
## Description

The heatmap is currently broken and does not show colors even when the toggle is turned on.

image

This PR fixes that by removing erroneous color configuration which was introduced in #2025. Also somewhat related to #2081  and #2086.

image

## Checklist

### DIFF
--- a/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/StatsTableUtils.ts
+++ b/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/StatsTableUtils.ts
@@ -141,9 +141,7 @@ export function generateCohortsStatsTable(
             // only 1 unique value in the set, set color to 0
             colorValue = 0;
           }
-          const colorConfig = { color: "transparent" };
           items.push({
-            ...colorConfig,
             colorValue,
             value: Number(labeledStat.stat.toFixed(3)),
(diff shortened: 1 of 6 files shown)


libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/StatsTableUtils.ts
```diff
@@ -141,9 +141,7 @@ export function generateCohortsStatsTable(
             // only 1 unique value in the set, set color to 0
             colorValue = 0;
           }
-          const colorConfig = { color: "transparent" };
```
Advitya17, line 144: Could we conditionally keep this for MO metrics (in dataset cohorts)?

Advitya17 (changes requested):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/StatsTableUtils.ts
romanlutz: I've turned off the heatmap for OD. See updates :-)

Advitya17 (approved):

hawestra (approved):

PR #2225, opened 2023-08-09
Error Analysis backend support for Object Detection

Advitya17:
This PR adds EA support for the OD scenario, to achieve feature parity with the rest of the image dashboards. This baseline support with the error rate metric is equivalent in UX to that of the multilabel scenario.

This is also the first PR that utilizes the Error Labeling work from `vision_explanation_methods` repo to evaluate the prediction and consider OD-specific edge cases (such as localization, duplicate/missing detections etc). Further testing & improvements are separately being made to the Error Labeling work which EA can piggyback off of (and other parts of the dashboard in upcoming work).

Limitations:
1. The IoU threshold is not currently surfaced to the user. Support could be added such as an IoU slider for EA, or an universal IoU slider after other components also utilize the Error Labeling work.
2. The Error Rate metric is a naive metric for both OD & multilabel as it marks a prediction as incorrect even with 1 incorrect label/detection. Support for better metrics can be added based on customer feedback.

Snapshots after also making the change in https://github.com/microsoft/responsible-ai-toolbox/pull/2233 + utilizing editable installs of `responsibleai` as per `main`:

image

image

image

## Checklist


natalie-isak:
I didn't look too deeply into your build errors, but it looks like you need to add the OBJECT_DETECTION constant to ModelTask (which it looks like you did?). Maybe you need to build raiutils indepedently prior to building the remainder of this PR?

erroranalysis/tests/test_cohort_filter.py
```diff
@@ -346,6 +348,25 @@ def test_cohort_filter_regression_error(self):
                            model_task,
                            filters=filters)
 
+    def test_cohort_filter_object_detection(self):
+        model_task = ModelTask.OBJECT_DETECTION
+        model = get_object_detection_fridge_model()
+
+        dataset = load_fridge_object_detection_dataset()
+        X_test = dataset[["image"]]
```
imatiach-msft, line 356: nit: use image/label constants here?

PR #2227, opened 2023-08-09
OD model type support in raiutils for EA

Advitya17:
This PR adds OD enum in EA due to the hybrid use-case of OD models.

## Checklist

### DIFF
--- a/raiutils/raiutils/models/model_utils.py
+++ b/raiutils/raiutils/models/model_utils.py
@@ -7,7 +7,8 @@ from enum import Enum
 class ModelTask(str, Enum):
     """Provide model task constants.
 
-    Can be 'classification', 'regression', 'forecasting', or 'unknown'.
+    Can be 'classification', 'regression', 'object_detection',
+    'forecasting', or 'unknown'.
 
     Note: Keeping sentence case constants (Classification, Regression)
     for backwards compatibility, please use ALL_UPPER_CASE instead.
@@ -15,6 +16,7 @@ class ModelTask(str, Enum):
 
     CLASSIFICATION = 'classification'
     REGRESSION = 'regression'
+    OBJECT_DETECTION = 'object_detection'
     FORECASTING = 'forecasting'
     UNKNOWN = 'unknown'
     Classification = 'classification'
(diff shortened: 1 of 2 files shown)


raiutils/raiutils/models/model_utils.py
```diff
@@ -60,6 +62,19 @@ def is_forecaster(model):
...
+    """Check if the model is an object detection model.
+
+    :return: True if the model is a an object detection model,
+    False otherwise.
+    :rtype: bool
+    """
+    return (model is not None and
+            hasattr(model, SKLearn.PREDICT) and
+            hasattr(model, SKLearn.PREDICT_PROBA) and
+            model.predict_proba is not None)
```
hawestra, line 75: why does this function determine whether a model is object_detection ? doesn't it just ensure that predict and predict_proba are defined or am i missing something ?

raiutils/raiutils/models/model_utils.py
```diff
@@ -60,6 +62,19 @@ def is_forecaster(model):
...
 
 
+def is_object_detection_model(model):
+    """Check if the model is an object detection model.
+
+    :return: True if the model is a an object detection model,
+    False otherwise.
+    :rtype: bool
+    """
+    return (model is not None and
```
gaugup, line 72: Couldn't you leverage some code from is_classifier() and just check for hasattr(model, SKLearn.PREDICT) additionally?

raiutils/tests/test_model_utils.py
```diff
@@ -44,3 +53,9 @@ def test_forecaster(self):
         assert not is_classifier(forecaster)
         assert is_forecaster(forecaster)
         assert is_quantile_forecaster(forecaster)
+
+    def test_object_detection_model(self):
+        forecaster = ObjectDetectionModel()
```
gaugup, line 58: Maybe change the variable name here to object_detector here instead of forecaster?

gaugup (changes requested):

imatiach-msft (approved):

gaugup (approved):

PR #2228, opened 2023-08-09
Move missing value validations before unique value validations

gaugup:
## Description

The np.unique() call downstream will fail in case of missing values in data. So doing the missing value validation up front.

## Checklist

### DIFF
--- a/responsibleai/tests/rai_insights/test_rai_insights_validations.py
+++ b/responsibleai/tests/rai_insights/test_rai_insights_validations.py
@@ -398,7 +398,7 @@ class TestRAIInsightsValidations:
             str(ucve.value)
 
     def test_dirty_train_test_data(self):
-        X_train = pd.DataFrame(data=[['1', np.nan], ['2', '3']],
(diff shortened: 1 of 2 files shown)


romanlutz (approved):

hawestra (approved):

imatiach-msft (approved):

responsibleai/tests/rai_insights/test_rai_insights_validations.py
```diff
@@ -398,7 +398,7 @@ def test_mismatch_train_test_features(self):
             str(ucve.value)
 
     def test_dirty_train_test_data(self):
```
imatiach-msft, line 400: should we add a test with both missing and non-unique data and validate it fails correctly?  It doesn't look like the np.nan -> 1 test change here is validating this scenario?

gaugup: There are tests for missing data. The dirty data test was modified because it had missing value and was failing for missing value reason. the np.nan -> 1 actually validates the dirty data condition.

PR #2229, opened 2023-08-09
OD model type support in raiwidgets for EA

Advitya17:
This PR adds OD support in EA due to the hybrid use-case of OD models.

## Checklist

### DIFF
--- a/raiwidgets/raiwidgets/error_analysis_dashboard_input.py
+++ b/raiwidgets/raiwidgets/error_analysis_dashboard_input.py
@@ -346,6 +346,8 @@ class ErrorAnalysisDashboardInput:
                 metric = Metrics.ERROR_RATE
             else:
                 metric = self._error_analyzer.metric
+        elif self._error_analyzer.model_task == ModelTask.OBJECT_DETECTION:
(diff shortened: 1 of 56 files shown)


raiwidgets/raiwidgets/error_analysis_dashboard_input.py
```diff
@@ -346,6 +346,8 @@ def setup_local(self, explanation, model, dataset, true_y, classes,
                 metric = Metrics.ERROR_RATE
             else:
                 metric = self._error_analyzer.metric
+        elif self._error_analyzer.model_task == ModelTask.OBJECT_DETECTION:
```
Advitya17, line 349: Hi @imatiach-msft , I haven't added any unit test for the OD case here as that'll require the `torch` and `torchvision` modules to load the fridge model, and hence may increase dependencies on this module (+ pass on similar torch version issues to this module in the future with 1.13 v/s >2.0).

Curious if you have suggestions to navigate this for purposes of unit testing, or if I shouldn't worry about testing here. Thanks!

tongyu-microsoft (approved):

imatiach-msft (approved):

PR #2230, opened 2023-08-09
OD model type support in responsibleai for EA

Advitya17:
This PR adds OD support in EA due to the hybrid use-case of OD models.

## Checklist


responsibleai/responsibleai/managers/error_analysis_manager.py
Advitya17, line 1: Hi @imatiach-msft , for the ModelAnalyzer call on line 504, I see it doesn't have the class names, and I had also propagated the OD enum in the code change I made to line 270. Am I also supposed to put in the class names and OD enum on 504 as well?

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -474,7 +476,8 @@ def _validate_rai_insights_input_parameters(
         """
         valid_tasks = [
             ModelTask.CLASSIFICATION.value,
-            ModelTask.REGRESSION.value
+            ModelTask.REGRESSION.value,
```
imatiach-msft, line 479: we should not be modifying rai_insights - only rai vision insights?

responsibleai/tests/model_analysis/test_model_analysis_validations.py
```diff
@@ -30,7 +30,8 @@ def test_validate_unsupported_task_type(self):
         X_test['target'] = y_test
 
         message = ("Unsupported task type 'regre'. "
```
imatiach-msft, line 32: I don't think we should be seeing this error - this is for tabular data only - this validation should not be run for vision?

responsibleai/responsibleai/managers/error_analysis_manager.py
imatiach-msft: are you running into an issue with EA and automl OD models specifically?  I fixed that in this PR:
https://github.com/microsoft/responsible-ai-toolbox/pull/2245
otherwise not sure what you are referring to here - we shouldn't be making any changes here, as this package is for tabular data - we should be making these changes in rai vision insights I believe?

responsibleai/responsibleai/rai_insights/rai_insights.py
Advitya17: We're just modifying rai insights in terms of propagating the OD enum to EA manager and making sure the user input specified for classes is returned as is.

responsibleai/tests/model_analysis/test_model_analysis_validations.py
Advitya17: Update: tweaked the test case to reflect the original prod user error message

imatiach-msft (approved):

PR #2235, opened 2023-08-10
fix for aggregate parameter passed to torchmetrics

imatiach-msft:
## Description

The MeanAveragePrecision no longer takes in an aggregate parameter.  This PR fixes this issue by removing it and adding a test to validate the compute_object_detection_metrics call.  Note I do not remove the UI changes here since we may still want to keep this parameter when computing in aggregate metrics across classes - but we need to call a different method in this case anyway, and this is currently not yet supported.

## Checklist

### DIFF
--- a/responsibleai_vision/tests/test_rai_vision_insights.py
+++ b/responsibleai_vision/tests/test_rai_vision_insights.py
@@ -287,3 +287,13 @@ def run_rai_insights(model, test_data, target_column,
     validate_rai_vision_insights(
         rai_insights, test_data,
         target_column, task_type)
+    if task_type == ModelTask.OBJECT_DETECTION:
+        selection_indexes = [[0]]
+        aggregate_method = 'Macro'
+        class_name = classes[0]
+        iou_threshold = 70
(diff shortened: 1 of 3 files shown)


responsibleai_vision/tests/test_rai_vision_insights.py
```diff
@@ -287,3 +287,13 @@ def run_rai_insights(model, test_data, target_column,
...
+    if task_type == ModelTask.OBJECT_DETECTION:
+        selection_indexes = [[0, 1]]
+        aggregate_method = 'Macro'
+        class_name = classes[0]
+        iou_threshold = 70
+        object_detection_cache = {}
+        metrics = rai_insights.compute_object_detection_metrics(
+            selection_indexes, aggregate_method, class_name, iou_threshold,
+            object_detection_cache)
+        assert len(metrics) == 2
```
Advitya17, line 299: maybe also assert the number of metrics per cohort is 3?

Advitya17 (approved):

PR #2248, opened 2023-08-14
Raise usererror on prediction failure.

kicha0:
Use full test dataset for validation in rai text insight. 

## Checklist

### DIFF
--- a/responsibleai_text/responsibleai_text/managers/error_analysis_manager.py
+++ b/responsibleai_text/responsibleai_text/managers/error_analysis_manager.py
@@ -15,6 +15,7 @@ from erroranalysis._internal.error_analyzer import ModelAnalyzer
 from erroranalysis._internal.error_report import as_error_report
 from responsibleai._tools.shared.state_directory_management import \
     DirectoryManager
+from responsibleai.exceptions import UserErrorException
 from responsibleai.managers.error_analysis_manager import \
     ErrorAnalysisManager as BaseErrorAnalysisManager
 from responsibleai.managers.error_analysis_manager import as_error_config
@@ -71,8 +72,12 @@ class WrappedIndexPredictorModel:
                          ModelTask.MULTILABEL_TEXT_CLASSIFICATION]
         if self.task_type in classif_tasks:
             dataset = self.dataset.iloc[:, 0].tolist()
-            self.predictions = self.model.predict(dataset)
-            self.predict_proba = self.model.predict_proba(dataset)
+            self.predictions = self._raise_user_error_on_failure(
+                self.model.predict, dataset
+            )
+            self.predict_proba = self._raise_user_error_on_failure(
+                self.model.predict_proba, dataset
+            )
         elif self.task_type == ModelTask.QUESTION_ANSWERING:
             self.predictions = self.model.predict(
                 self.dataset.loc[:, ['context', 'questions']])
@@ -120,6 +125,15 @@ class WrappedIndexPredictorModel:
         index = X.index
         pred_proba = self.predict_proba[index]
         return pred_proba
+
+    def _raise_user_error_on_prediction_failure(func, *args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except Exception as ex:
+            raise UserErrorException(
+                "Unable to use user model to retrieve predictions"
+                f" from given dataset. Original exception: {ex}"
+            )
 
 
 class ErrorAnalysisManager(BaseErrorAnalysisManager):
--- a/responsibleai_text/responsibleai_text/managers/error_analysis_manager.py
+++ b/responsibleai_text/responsibleai_text/managers/error_analysis_manager.py
@@ -72,10 +72,10 @@ class WrappedIndexPredictorModel:
                          ModelTask.MULTILABEL_TEXT_CLASSIFICATION]
         if self.task_type in classif_tasks:
             dataset = self.dataset.iloc[:, 0].tolist()
-            self.predictions = self._raise_user_error_on_failure(
+            self.predictions = self._raise_user_error_on_prediction_failure(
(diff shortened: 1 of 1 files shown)


responsibleai_text/responsibleai_text/managers/error_analysis_manager.py
```diff
@@ -67,15 +70,23 @@ def __init__(self, model, dataset, is_multilabel, task_type, classes=None):
...
-                         ModelTask.MULTILABEL_TEXT_CLASSIFICATION]
+        classif_tasks = [
+            ModelTask.TEXT_CLASSIFICATION,
+            ModelTask.MULTILABEL_TEXT_CLASSIFICATION,
+        ]
         if self.task_type in classif_tasks:
             dataset = self.dataset.iloc[:, 0].tolist()
-            self.predictions = self.model.predict(dataset)
-            self.predict_proba = self.model.predict_proba(dataset)
+            self.predictions = self._raise_user_error_on_failure(
```
kicha0, line 79: everything else is from black formatter, let me know and I can revert the changes on formatting

gaugup: Shall we move the predict on the entire dataset in responsibleai-text here:- https://github.com/microsoft/responsible-ai-toolbox/blob/798e9d5eb12c35d72aedea0f604753543c26f702/responsibleai_text/responsibleai_text/rai_text_insights/rai_text_insights.py#L245? Moving predict validation in reponsibleai-text makes the validation same across all RAI verticals.

gaugup (approved):

responsibleai_text/responsibleai_text/rai_text_insights/rai_text_insights.py
```diff
@@ -264,16 +264,15 @@ def _validate_model(self, model: Any, test: pd.DataFrame,
...
-        small_test_data = get_text_columns(small_test_data, text_column)
-        small_test_data = small_test_data.iloc[0]
+        test_data = get_text_columns(test_data, text_column)
         if task_type != ModelTask.QUESTION_ANSWERING:
-            small_test_data = small_test_data.tolist()
+            test_data = test_data.tolist()
         # Call the model
         try:
-            model.predict(small_test_data)
+            model.predict(test_data)
```
imatiach-msft, line 275: For text this will be very slow - I don't suggest going down this route.  Text predictions are much slower than tabular.  Is there any reason why we need to do this just for validation?

responsibleai_text/responsibleai_text/rai_text_insights/rai_text_insights.py
```diff
@@ -264,16 +264,15 @@ def _validate_model(self, model: Any, test: pd.DataFrame,
...
-        small_test_data = test.iloc[0:1].drop(
+        # Use full test dataset for validation
+        test_data = test.drop(
             target_column, axis=1)
-        small_test_data = get_text_columns(small_test_data, text_column)
-        small_test_data = small_test_data.iloc[0]
+        test_data = get_text_columns(test_data, text_column)
         if task_type != ModelTask.QUESTION_ANSWERING:
-            small_test_data = small_test_data.tolist()
+            test_data = test_data.tolist()
```
imatiach-msft, line 272: it looks like tests are all failing with errors:

```
responsibleai_text/rai_text_insights/rai_text_insights.py:163: in __init__
    self._validate_rai_insights_input_parameters(
responsibleai_text/rai_text_insights/rai_text_insights.py:394: in _validate_rai_insights_input_parameters
    self._validate_model(model, test, target_column,
responsibleai_text/rai_text_insights/rai_text_insights.py:272: in _validate_model
    test_data = test_data.tolist()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self =                                                 text
0                            i didnt feel humiliated
1  i can go from feeling so hopeless to so damned...
2   im grabbing a minute to post i feel greedy wrong
name = 'tolist'

    def __getattr__(self, name: str):
        """
        After regular attribute access, try looking up the name
        This allows simpler access to columns for interactive use.
        """
        # Note: obj.x will always call obj.__getattribute__('x') prior to
        # calling obj.__getattr__('x').
        if (
            name not in self._internal_names_set
            and name not in self._metadata
            and name not in self._accessors
            and self._info_axis._can_hold_identifiers_and_holds_name(name)
        ):
            return self[name]
>       return object.__getattribute__(self, name)
E       AttributeError: 'DataFrame' object has no attribute 'tolist'
```

imatiach-msft (changes requested):
It looks like tests are failing with errors and this will make things much slower - any way to avoid this?  We only have one component for text so I wonder if we can just catch this where the error occurs instead of here?

responsibleai_text/responsibleai_text/rai_text_insights/rai_text_insights.py
kicha0: changed back to raising user error in error analysis manager predict call

responsibleai_text/responsibleai_text/rai_text_insights/rai_text_insights.py
kicha0: changed back to raising user error in error analysis manager predict call

responsibleai_text/responsibleai_text/managers/error_analysis_manager.py
kicha0: discussed on teams. full test data prediction is a very expensive call especially on text data. We can mitigate the issue with good user error exception message for now and think of caching later.

imatiach-msft (approved):

PR #2253, opened 2023-08-16
refactor notebook examples into tabular, text and vision folders

imatiach-msft:
## Description

Refactor notebook examples into tabular, text and vision folders.

New folder structure:

image

## Checklist


imatiach-msft:
"Shall we put getting started and tour also in tabular since these are tabular only."

done!

gaugup (approved):

PR #2258, opened 2023-08-18
RAIInsights: prevent failures on optional methods & fix feature range formatting for timestamps

romanlutz:
## Description

With optional methods such as `forecast_quantiles` we need to be able to check if the model has the method at all without getting a failure. This PR adjusts the handling slightly to prevent unintended issues.

Secondly, the feature ranges need to be specified in a JSON-serializable format, so we're storing them as strings now.

## Checklist


gaugup (changes requested):
Could add unit tests for these changes?

tongyu-microsoft (approved):

romanlutz:
> Could add unit tests for these changes?

Added tests as requested.

PR #2263, opened 2023-08-18
Tabular Only: Address missing values as per each manager in responsibleai

gaugup:
## Description

The PR looks to address the issue of missing data values by pushing the missing value checks inside each manager and customizing the checks according to the manager. The check introduced under `_validate_data_is_not_missing` fails explanations and error analysis which could be computed if the model handles the missing values correctly (that is the predictions work fine for missing data).

Below are the set of changes introduced in this PR for handling missing data.
1. Remove the check `_validate_data_is_not_missing` and it's corresponding unit tests.
2. Performing predict()/predict_proba() on the entire dataset and not just the first row of the dataset.
3. Handling missing value logic per manager:-
  - The explainer manager computes explanations if only the numerical columns have missing values. If the categorical columns have missing values, then it asks the users to impute those missing values.
  - The eror manager computes error report but not feature importances for data having missing values
  - The counterfactuals and causal effects are not computed for data having missing values. They ask users to modify their data to fill missing values. We cross checked with econml team that we shouldn't impute values as this could lead to false causal effects.
 4. The logic to check if we can find unique values in data using `np.unique()` has been located under `_data_validations.py` so that it only applies to causal and counterfactual managers.
 5. Added tests for testing the missing value matrix according to each manager, missing value type (like numerical/categorical) and train/test datasets. 

 ## Checklist


responsibleai/responsibleai/managers/explainer_manager.py
```diff
@@ -97,6 +98,22 @@ def add(self):
             raise UserConfigValidationException(
                 'Model is required for model explanations')
 
+        if len(set(_find_features_having_missing_values(
```
romanlutz, line 101: nit: `any(...)` is preferable here because we don't actually care about the length, but rather that there are any elements (not zero).

responsibleai/responsibleai/managers/explainer_manager.py
```diff
@@ -97,6 +98,22 @@ def add(self):
...
+            self._initialization_examples)) &
+                set(self._categorical_features)) > 0:
+            raise UserConfigValidationException(
+                "Categorical features cannot have missing "
+                "values for computing explanations. "
+                "Please check your training data.")
+
+        if len(set(_find_features_having_missing_values(
```
romanlutz, line 109: same here with `any`

responsibleai/responsibleai/managers/causal_manager.py
```diff
@@ -158,6 +159,14 @@ def add(
         :param random_state: Controls the randomness of the estimator.
         :type random_state: int or RandomState or None
         """
+        if len(_find_features_having_missing_values(self._train)) > 0:
```
romanlutz, line 162: same here with `any`, and right below

responsibleai/responsibleai/managers/explainer_manager.py
```diff
@@ -97,6 +98,22 @@ def add(self):
...
+            self._evaluation_examples)) &
+                set(self._categorical_features)) > 0:
+            raise UserConfigValidationException(
```
romanlutz, line 112: Alternatively, we could tell users which features have missing values. As a user, this sort of thing is very annoying to me since the library already did the work and still chose not to tell me the answer 😆 Same applies to all those occurrences, of course.

[In that case, you would not be able to use `any`, of course, since you actually need to say which features were found...]

responsibleai/responsibleai/rai_insights/rai_insights.py
```diff
@@ -643,8 +621,8 @@ def _validate_rai_insights_input_parameters(
             # Ensure that the model has the required methods and that they
             # do not change the input data.
             if task_type != ModelTask.FORECASTING:
-                self._ensure_model_outputs(input_data=small_train_data)
-            self._ensure_model_outputs(input_data=small_test_data)
+                self._ensure_model_outputs(input_data=actual_train_data)
```
imatiach-msft, line 624: In some cases calling predict can be expensive on some models - by using all data we can really slow down our dashboard computations much more.  Perhaps we will need to look into some caching logic for such scenarios, since we seem to be calling predict on all data many times but not re-using the results anywhere.

responsibleai/responsibleai/utils.py
```diff
@@ -19,3 +23,18 @@ def compute_wrapper(*args, **kwargs):
 
 def _separator(max_len):
     print('=' * max_len)
+
+
+def _find_features_having_missing_values(
+        data: pd.DataFrame) -> List[str]:
+    """Return list of features which have missing values.
+
+    parame data: The dataset to check.
```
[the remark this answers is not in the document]
imatiach-msft: also needs to have colon in front, like :param data:
same below

imatiach-msft (approved):

responsibleai/responsibleai/utils.py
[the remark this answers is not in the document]
gaugup: done

responsibleai/responsibleai/managers/explainer_manager.py
gaugup: done

responsibleai/responsibleai/managers/causal_manager.py
gaugup: done

responsibleai/responsibleai/managers/explainer_manager.py
gaugup: done

responsibleai/responsibleai/rai_insights/rai_insights.py
gaugup: I agree we seem to be doing quite a few prediction calls. Caching these predictions maybe is out of scope of this PR. Maybe I can follow up on this later.

PR #2266, opened 2023-08-21
GPU support for OD metrics

Advitya17:
Added device support for calculating Model Overview metrics for OD

## Checklist


responsibleai_vision/responsibleai_vision/rai_vision_insights/rai_vision_insights.py
```diff
@@ -1091,6 +1091,7 @@ def compute_object_detection_metrics(
         dashboard_dataset = self.get_data().dataset
         true_y = dashboard_dataset.object_detection_true_y
         predicted_y = dashboard_dataset.object_detection_predicted_y
+        device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
```
imatiach-msft, line 1094: We should be using the device parameter passed in the constructor instead - sometimes torch.cuda.is_available may return something unexpected to the user, or the user may encounter errors on one device and they will have to use another.  Moreover, we should be able to specify a device int in case there are multiple GPUs using the constructor parameter.

imatiach-msft (changes requested):
see comment to use device parameter from constructor

imatiach-msft (approved):

PR #2268, opened 2023-08-22
Endpoint Interruption Logic for Model Overview with OD

Advitya17:
https://github.com/microsoft/responsible-ai-toolbox/assets/37190647/36c28b06-a641-45c4-b08c-4fe7dc537e15

## Description

For scenarios such as OD & QA that use an endpoint to calculate Model Overview metrics, even if the metrics are drawn from the cache, there's still an execution invoked for the endpoint to correctly trigger the state change. 

If there're any user interaction related to changing cohorts or quick user input changes while a previous endpoint instance is running, the dashboard would fall into an error. This PR adds abort logic to the endpoint such that a change in user inputs aborts the previous execution to avoid errors in the state or cohort values.

A future PR would add corresponding support to QA as well.

## Checklist


romanlutz (approved):
Nice, also thank you for adding the video!

Have we considered adding a spinner while it's loading?

Advitya17:
> Nice, also thank you for adding the video!
> 
> Have we considered adding a spinner while it's loading?

Thanks for the review, @romanlutz! It has come up in discussions before, but we haven't put that in the backlog just yet. Currently we're making do with `N/A` until values are rendered, but could see in a future PR if spinners inside the table values is (a) feasible to put in (instead of a string), and (b) preferred PM decision.

Let me know your thoughts and can bring it to discussion accordingly.

PR #2286, opened 2023-08-28
Add torch installation support for rai test utils release

Advitya17:
## Description

PR raised to fix this Test pipeline failure with new torch dependencies - https://github.com/microsoft/responsible-ai-toolbox/actions/runs/5988756217/job/16244250255

## Checklist

### DIFF
--- a/.github/workflows/CI-notebook-text.yml
+++ b/.github/workflows/CI-notebook-text.yml
@@ -9,6 +9,9 @@ on:
       - "raiwidgets/**"
       - "responsibleai_text/**"
       - ".github/workflows/CI-notebook-text.yml"
+      - "libs/e2e/src/lib/describer/modelAssessment/**"
+      - "libs/interpret-text/**"
+      - "notebooks/**"
 
 jobs:
   ci-notebook-text:
--- a/.github/workflows/CI-responsibleai-text-vision-pytest.yml
+++ b/.github/workflows/CI-responsibleai-text-vision-pytest.yml
@@ -97,10 +97,18 @@ jobs:
         run: |
           pip install -r ${{ matrix.packageDirectory }}/requirements-automl.txt
 
-      - name: Install package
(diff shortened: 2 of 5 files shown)


imatiach-msft (approved):

.github/workflows/release-rai-test-utils.yml
```diff
@@ -26,31 +26,55 @@ jobs:
...
+          auto-update-conda: true
+          python-version: ${{ matrix.pythonVersion }}
+
+      - if: ${{ matrix.operatingSystem != 'macos-latest' }}
+        name: Install pytorch on non-MacOS
+        shell: bash -l {0}
+        run: |
+          conda install --yes --quiet pytorch torchvision captum cpuonly -c pytorch
+
+      - if: ${{ matrix.operatingSystem == 'macos-latest' }}
```
gaugup, line 40: Same as above.

.github/workflows/release-rai-test-utils.yml
```diff
@@ -26,31 +26,55 @@ jobs:
...
+        name: Install Anaconda packages on MacOS, which should not include cpuonly according to official docs
+        shell: bash -l {0}
+        run: |
+          conda install --yes --quiet pytorch torchvision captum -c pytorch
+
       - name: update and upgrade pip, setuptools, wheel, and twine
+        shell: bash -l {0}
```
gaugup, line 47: Why do you need bash -1 {0} in all the workflow steps?

gaugup (changes requested):

.github/workflows/release-rai-test-utils.yml
Advitya17: As per @imatiach-msft's suggestion and prior testing that this bash command will be needed in places where both python & conda (because of torch) are setup. SImilar to `CI-Python.yml`.

gaugup (approved):

PR #2289, opened 2023-08-29
Updated release file for `erroranalysis`

Advitya17:
## Description

Updated release file with torch-conda support due to `vision_explanation_methods` dependency with the OD scenario, which contains the Error Labeling work and utilizes torch & torchvision.

## Checklist


.github/workflows/release-erroranalysis.yml
```diff
@@ -27,31 +27,47 @@ jobs:
         with:
           python-version: 3.8
 
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          auto-update-conda: true
+          python-version: ${{ matrix.pythonVersion }}
```
gaugup, line 33: Looks like matrix is not defined. Could you validate the yml file?

gaugup (changes requested):

.github/workflows/CI-python-vision.yml
```diff
@@ -0,0 +1,146 @@
...
+          pip install -r requirements-object-detection.txt
+        working-directory: ${{ matrix.packageDirectory }}
+
+      - name: Install package
+        shell: bash -l {0}
+        run: |
+          pip install -v -e .
+        working-directory: ${{ matrix.packageDirectory }}
+
+      - name: Pip freeze
```
gaugup, line 83: Why do you need to install raiwidgets here?

gaugup: I think we can remove this step.

.github/workflows/CI-python-vision.yml
```diff
@@ -0,0 +1,146 @@
...
+          name: requirements-dev.txt
+          path: raiwidgets/installed-requirements-dev.txt
+
+      - name: Run tests
+        shell: bash -l {0}
+        run: |
+          pytest --durations=10 --doctest-modules --junitxml=junit/test-results.xml --cov=${{ matrix.packageDirectory }} --cov-report=xml --cov-report=html
+        working-directory: ${{ matrix.packageDirectory }}
+
+      - name: Upload code coverage results
```
gaugup, line 102: The code coverage is already uploaded as a part of CI-python.yml. Do we need to upload this report again? The coverage will anyways be very less.

.github/workflows/CI-python-vision.yml
```diff
@@ -0,0 +1,146 @@
...
+      - ".github/workflows/CI-python-vision.yml"
+
+jobs:
+  ci-python:
+    strategy:
+      matrix:
+        packageDirectory:
+          ["erroranalysis", "rai_test_utils"]
+        operatingSystem: [ubuntu-latest, macos-latest, windows-latest]
+        pythonVersion: ["3.8", "3.9", "3.10"]
```
romanlutz, line 20: erroranalysis supports 3.6 onwards. Are we willing to drop support for them? If so, please update the setup.py file. Not testing and hoping that it works sounds dangerous to me.

@imatiach-msft FYI

.github/workflows/CI-python-vision.yml
```diff
@@ -0,0 +1,146 @@
...
+        name: Install Anaconda packages on MacOS, which should not include cpuonly according to official docs
+        shell: bash -l {0}
+        run: |
+          conda install --yes --quiet pytorch torchvision captum -c pytorch
+
+      - if: ${{ matrix.operatingSystem == 'macos-latest' }}
+        name: Install latest lightgbm from conda-forge for MacOS
+        shell: bash -l {0}
+        run: |
+          conda install --yes --quiet lightgbm -c conda-forge
```
romanlutz, line 56: Is lightgbm needed for vision?

.github/workflows/CI-python-vision.yml
```diff
@@ -0,0 +1,146 @@
...
+      - name: Install package extras
+        shell: bash -l {0}
+        run: |
+          pip install -r requirements-object-detection.txt
+        working-directory: ${{ matrix.packageDirectory }}
+
+      - name: Install package
+        shell: bash -l {0}
+        run: |
+          pip install -v -e .
```
romanlutz, line 80: So this is without the extras (which would be pip install .[object-detection]). What about with extras? I'm a bit confused honestly, because  you installed the requirements for the extras above. So this is neither really (?)

romanlutz (changes requested):

.github/workflows/CI-python-vision.yml
Advitya17: @romanlutz I installed `requirements-object-detection.txt` under the Install extras name. `pip install .[object_detection]` would just install this file under the hood; let me know if I should replace directly installing from test file to using the extras term?

PR #2297, opened 2023-08-30
Split release testing into two parallel branches

romanlutz:
## Description

The release pipeline is currently broken because the unittests don't mesh well with pytorch (seg faults, memory leaks), so this PR splits unittests (without pytorch) and notebook e2e tests (with pytorch) into two parts that are executed in parallel. It already succeeded in Test PyPI https://github.com/microsoft/responsible-ai-toolbox/actions/runs/6020548015

## Checklist


imatiach-msft (approved):

.github/workflows/release-rai-tabular.yml
```diff
@@ -238,7 +329,8 @@ jobs:
             interpret-vision,
             localization,
             error-analysis,
-            model-assessment
+            model-assessment,
+            e2e
```
gaugup, line 333: Do we need to publish e2e?

PR #2303, opened 2023-08-30
Heatmap Support for Object Detection

Advitya17:
This PR adds heatmap support for OD scenario to achieve corresponding feature parity with other dashboards.

## Description

image

image

The main drawback of this support is I had to end up disabling the magenta color during `N/A` status while metrics are being computed, because I could not override the metrics to have transparent color anywhere in StatsTableUtils.ts as that ended up overriding the heatmap colors.

image

Considering the magenta color with `N/A` display is only valid for OD & QA scenarios, and is not a breaking feature change but a nice-to-have UI (not to mention potentially switching this with async loaders), this support can be enabled in a future PR.

## Checklist

### DIFF
--- a/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/DatasetCohortStatsTable.tsx
+++ b/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/DatasetCohortStatsTable.tsx
@@ -19,6 +19,7 @@ interface IDatasetCohortStatsTableProps {
   selectableMetrics: IDropdownOption[];
   selectedMetrics: string[];
   showHeatmapColors: boolean;
+  modelType: string;
 }
 
 class IDatasetCohortStatsTableState {}
@@ -37,7 +38,8 @@ export class DatasetCohortStatsTable extends React.Component<
       this.props.selectableMetrics,
       this.props.labeledStatistics,
       this.props.selectedMetrics,
-      this.props.showHeatmapColors
+      this.props.showHeatmapColors,
+      this.props.modelType
     ).items;
 
     const showColors =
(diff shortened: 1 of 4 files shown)


romanlutz (commented):
This passes information (`modelType`) quite deep into the logic that doesn't really need this information. Instead, would it be possible to set this at the top level (constructor of ModelOverview, for example) based on the model type (technically, task type, but I guess it's fine...) and factor it into useTexturedBackgroundForNaN?

In general, we want to keep the components and their associated logic as decoupled as possible, and model type is rather the opposite. It is entirely possible that I'm missing something here that makes this necessary, so feel free to point that out!

Advitya17:
> This passes information (`modelType`) quite deep into the logic that doesn't really need this information. Instead, would it be possible to set this at the top level (constructor of ModelOverview, for example) based on the model type (technically, task type, but I guess it's fine...) and factor it into useTexturedBackgroundForNaN?
> 
> In general, we want to keep the components and their associated logic as decoupled as possible, and model type is rather the opposite. It is entirely possible that I'm missing something here that makes this necessary, so feel free to point that out!

@romanlutz Thanks for the question. Unfortunately the way the code is setup, if I just pass the ModelType to the `useTexturedBackgroundForNaN` which is the same input as `this.state.showHeatmapColors`, it totally removes the blue heatmap colors.

image

With the way the code flow works with N/A magenta colorValue and the blue color rendering for the heatmap (and the same input that both take in for assigning colors - `this.state.showHeatmapColors`), it only seemed possible to disable the N/A magenta color assignment inside the function, so that the blue color rendering isn't disabled as well.

romanlutz:
Oh. I guess I had assumed it's straightforward, sorry. If you see a simple way to refactor it go ahead, otherwise ignore.

romanlutz (approved):

PR #2304, opened 2023-08-31
Release erroranalysis 0.5.1

Advitya17:
## Checklist


gaugup (approved):

.github/workflows/CI-python.yml
```diff
@@ -40,57 +40,28 @@ jobs:
...
       - name: Install package dependencies
-        shell: bash -l {0}
         run: |
           pip install -r requirements-dev.txt
         working-directory: ${{ matrix.packageDirectory }}
 
-      - if: ${{ matrix.packageDirectory == 'rai_test_utils' }}
-        name: Install package extras
-        shell: bash -l {0}
+      - name: Install test extras
```
imatiach-msft, line 54: shouldn't this object detection section be removed altogether here, since we are not installing torch?

Advitya17: Good question, these dependencies only contain `requests` and `Pillow`. None of the `requirements-object-detection.txt` files contain torch, torch is always installed via conda. I added this step for tests to work across `rai_test_utils` and `erroranalysis` that utilizes methods from rai test utils in its tests. 

Considering rai test utils has common code to be used by other modules for testing + isn't part of any other builds, this step ensures that any non-torch tests (eg - loading image datasets) can run across all the other python modules.

Still, trying out another way (perhaps more elegant) that doesn't require this step section.

PR #2319, opened 2023-09-06
Forecasting: consume locally served model

romanlutz:
## Description

This PR adds functionality to consume a locally served forecasting model provided that an environment variable `RAI_MODEL_SERVING_PORT` with the corresponding port is set. Unittests are also part of the PR.

Note: this can later be extended to work for other types of models (not just forecasting), but that is currently out of scope.

## Checklist


responsibleai/responsibleai/_internal/_served_model_wrapper.py
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+from raiutils.data_processing import serialize_json_safe
+
+import json
+import requests
+
+
+class ServedModelWrapper:
```
imatiach-msft, line 10: I feel like this would more naturally live in ml-wrappers instead of the responsibleai package?

responsibleai/responsibleai/_internal/_served_model_wrapper.py
```diff
@@ -0,0 +1,32 @@
...
+# Licensed under the MIT License.
+
+from raiutils.data_processing import serialize_json_safe
+
+import json
+import requests
+
+
+class ServedModelWrapper:
+    def __init__(self, port):
```
imatiach-msft, line 11: missing docs, description, etc on all methods here

responsibleai/tests/rai_insights/test_served_model.py
```diff
@@ -0,0 +1,82 @@
...
+        model=model,
+        train=train,
+        test=test,
+        target_column="target",
+        task_type='forecasting',
+        feature_metadata=FeatureMetadata(
+            datetime_features=['time'],
+            time_series_id_features=['id']
+        ),
+        forecasting_enabled=True)
```
imatiach-msft, line 38: why is this extra forecasting_enabled flag needed if there is already a forecasting task type specified?

responsibleai/responsibleai/_internal/_served_model_wrapper.py
imatiach-msft: perhaps this is similar to the existing EndpointWrapperModel:

https://github.com/microsoft/ml-wrappers/blob/main/python/ml_wrappers/model/endpoint_wrapper.py#L38

which calls a model deployed as an AzureML endpoint?

romanlutz: Hmm, a few key differences, but for the moment I'd really like to retain the flexibility of making quick changes if needed. As we generalize for all tabular tasks I think it makes a ton of sense to see if we can move it to ml-wrappers (or add to an existing one).

My case is a lot simpler since it's just localhost with http, and I don't want predict or predict_proba

raiutils/raiutils/data_processing/data_processing_utils.py
```diff
@@ -143,7 +143,7 @@ def serialize_json_safe(o: Any):
         if isinstance(o, str):
             return json.dumps(o)[1:-1]
         return o
-    elif isinstance(o, datetime.datetime):
+    elif isinstance(o, datetime.datetime) or isinstance(o, pd.Timestamp):
```
gaugup, line 146: Could you have a separate PR for this in raiutils? Also add a test case for this is raiutils.

gaugup (changes requested):

responsibleai/tests/rai_insights/test_served_model.py
romanlutz: Because it's still experimental. If you don't set the feature flag it won't work. The feature flag is not documented.

raiutils/raiutils/data_processing/data_processing_utils.py
romanlutz: #2320

responsibleai/responsibleai/serialization_utilities.py
```diff
@@ -31,7 +32,7 @@ def serialize_json_safe(o: Any):
         if isinstance(o, str):
             return json.dumps(o)[1:-1]
         return o
-    elif isinstance(o, datetime.datetime):
+    elif isinstance(o, datetime.datetime) or isinstance(o, pd.Timestamp):
```
gaugup, line 35: Could you add a test for this change?

gaugup (approved):
Will need to go over related design doc before approving.

responsibleai/responsibleai/serialization_utilities.py
romanlutz: added

responsibleai/responsibleai/_internal/_served_model_wrapper.py
```diff
@@ -0,0 +1,54 @@
...
+            url=f"http://localhost:{self.port}/invocations",
+            headers={"Content-Type": "application/json"},
+            data=json.dumps(
+                {"dataframe_split": X.to_dict(orient='split')},
+                default=serialize_json_safe))
+        if response.status_code < 300:
+            # json.loads decodes byte string response.
+            # Response is a dictionary with a single entry 'predictions'
+            return json.loads(response.content)['predictions']
+        else:
```
xuke444, line 50: we might want some logging here.
Is there anything we can do to log some error deatils without violate pii?

responsibleai/responsibleai/_internal/_served_model_wrapper.py
```diff
@@ -0,0 +1,54 @@
...
+        # request formatting according to mlflow docs
+        # https://mlflow.org/docs/latest/cli.html#mlflow-models-serve
+        # JSON safe serialization takes care of datetime columns
+        response = requests.post(
+            url=f"http://localhost:{self.port}/invocations",
+            headers={"Content-Type": "application/json"},
+            data=json.dumps(
+                {"dataframe_split": X.to_dict(orient='split')},
+                default=serialize_json_safe))
+        if response.status_code < 300:
```
xuke444, line 46: this check may be a bit risky
how about 204, 202, 206?

romanlutz: Used raise_for_status instead as it handles all of these.

responsibleai/responsibleai/_internal/_served_model_wrapper.py
romanlutz: Caught the error case and raising with a specific error message that we can check for in Studio to show in telemetry & logs.

xuke444 (approved):

responsibleai/responsibleai/_internal/_served_model_wrapper.py
romanlutz: Changed to RuntimeError

PR #2325, opened 2023-09-07
fix error analysis for RAITextInsights displaying all incorrect instances for text labels

imatiach-msft:
## Description

On the DBPedia dataset, fix error analysis for RAITextInsights displaying all incorrect instances for text labels:
image
Note this is only happening for a particular multiclass text classification case where the predicted label is a numpy int type.

## Checklist

### DIFF
--- a/responsibleai_text/tests/common_text_utils.py
+++ b/responsibleai_text/tests/common_text_utils.py
@@ -29,8 +29,12 @@ COVID19_EVENTS_LABELS = ["event1", "event2", "event3", "event4",
                          "event5", "event6", "event7", "event8"]
 COVID19_EVENTS_MODEL_NAME = "covid19_events_model"
 BLBOOKSGENRE_MODEL_NAME = "blbooksgenre_model"
+DBPEDIA_MODEL_NAME = "dbpedia_model"
 NUM_BLBOOKSGENRE_LABELS = 2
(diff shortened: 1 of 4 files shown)


responsibleai_text/tests/common_text_utils.py
```diff
@@ -123,7 +127,15 @@ def load_blbooks_genre_dataset():
...
     dataset = dataset[dataset.label != 1].reset_index(drop=True)
     dataset = dataset.drop(columns=grouping_col)
-    print(dataset)
+    return dataset
+
+
+def load_dbpedia_dataset():
+    dataset = datasets.load_dataset(
+        "DeveloperOats/DBPedia_Classes", split="test")
+    dataset = pd.DataFrame({"text": dataset["text"], L1: dataset[L1]})
```
Advitya17, line 136: Instead of lines 136 to 138, I'd recommend simply renaming the column from L1 to DBPEDIA_LABEL, like

`
dataset.rename(columns={L1 : DBPEDIA_LABEL}, inplace=True)
`

Advitya17 (approved):

responsibleai_text/tests/common_text_utils.py
imatiach-msft: good idea, done!

PR #2330, opened 2023-09-09
Single Container in Image Data Explorer for Object Detection

Advitya17:
This PR replaces the container view with a single container for OD scenario, which is the first step towards displaying correct/incorrect objects per image instead of classifying them as a binary correct/incorrect prediction.

image

Intended behavior (example from Figma):

image

## Checklist


imatiach-msft (approved):

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/TabsView.tsx
```diff
@@ -104,44 +109,76 @@ export class TabsView extends React.Component<ITabsViewProps, ITabViewState> {
...
-                <Stack.Item id="errorInstances">
-                  <TitleBar
-                    count={this.props.errorInstances.length}
-                    type={TitleBarOptions.Error}
-                  />
-                </Stack.Item>
-                <Stack.Item
-                  className={classNames.imageListContainer}
-                  id="errorImageContainer"
+              {this.props.taskType === DatasetTaskType.ObjectDetection ? (
```
imatiach-msft, line 112: did you make sure that the non-object detection cases are still working exactly the same as before?  asking because we had several cases in the past where object detection UI changes broke the multiclass/multilabel scenarios.

PR #2339, opened 2023-09-15
Detection info inside Object Detection Flyout

Advitya17:
This PR expands on the Error Labeling integration to provide precise object detection labels in terms of correct & incorrect detections inside the Flyout.

## Description

image

## Checklist


libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
```diff
@@ -76,8 +76,8 @@ export class FlyoutObjectDetection extends React.Component<
       return <div />;
     }
     const classNames = flyoutStyles();
-    const predictedY = getJoinedLabelString(item?.predictedY);
-    const trueY = getJoinedLabelString(item?.trueY);
+    const correctDetections = getJoinedLabelString(item?.odCorrect);
+    const incorrectDetections = getJoinedLabelString(item?.odIncorrect);
 
     return (
       <FluentUI.FocusZone>
```
imatiach-msft, line 83: I see some warnings like:
image
any idea what is causing these?

responsibleai/responsibleai/_interfaces.py
```diff
@@ -38,7 +38,7 @@ class Dataset:
     object_detection_true_y: Optional[List]
     object_detection_predicted_y: Optional[List]
     imageDimensions: Optional[List[List[int]]]
-    object_detection_labels: Optional[List[str]]
+    object_detection_labels: Optional[List[Dict[str, str]]]
```
imatiach-msft, line 41: This worries me a bit because we keep making breaking changes for object detection.  Now the old dashboards won't work at all anymore, if I understand correctly?
Might it be better at this point to just introduce a new "optional" variable to ensure that the old dashboards will still somehow render, or should we just go ahead and make this breaking change?

imatiach-msft (approved):

libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
imatiach-msft: seems like builds are failing on these as well

responsibleai/responsibleai/_interfaces.py
Advitya17: Discussed offline, we wouldn't want the users to use the old dashboard as those Object Detection labels are incorrect.

PR #2342, opened 2023-09-15
causal_manager.py: Move user config validation checks from compute() to add()

gaugup:
## Description

It looks like these checks should be part to add() so stop the unnecessary compute() execution.

## Checklist

### DIFF
--- a/responsibleai/responsibleai/managers/causal_manager.py
+++ b/responsibleai/responsibleai/managers/causal_manager.py
@@ -219,6 +219,36 @@ class CausalManager(BaseManager):
                        f"got {heterogeneity_model}")
             raise UserConfigValidationException(message)
 
+        # Check treatment_cost is valid
+        if isinstance(treatment_cost, int) and \
+                treatment_cost == 0:
+            revised_treatment_cost = [0] * len(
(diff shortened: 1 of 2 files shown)


responsibleai/responsibleai/managers/causal_manager.py
```diff
@@ -219,6 +219,36 @@ def add(
                        f"got {heterogeneity_model}")
             raise UserConfigValidationException(message)
 
+        # Check treatment_cost is valid
+        if isinstance(treatment_cost, int) and \
```
Advitya17, line 223: Is the first part of the condition needed if the next condition already checks with `and` if the variable is 0? Trying to check if a list variable == 0 may already return False without needing to check its dtype .. ?

Advitya17 (approved):

responsibleai/responsibleai/managers/causal_manager.py
gaugup: I guess it is needed as it means that all non-zero treatment_cost will result in user error.

imatiach-msft (approved):

PR #2343, opened 2023-09-15
Remove unreachable branch in `test_rai_insights_save_and_load_scenarios.py`

gaugup:
## Description

The pytest parameter list does not contain any other manager name.

## Checklist


responsibleai/tests/rai_insights/test_rai_insights_save_and_load_scenarios.py
```diff
@@ -207,11 +207,8 @@ def test_rai_insights_add_save_load_save(self, manager_type, adult_data):
             )
         elif manager_type == ManagerNames.ERROR_ANALYSIS:
             rai_insights.error_analysis.add()
-        elif manager_type == ManagerNames.EXPLAINER:
```
Advitya17, line 210: Nit: Why is this change needed? Won't the value error double check that an undesired enum doesn't go through, and won't we be assuming with the `else` that we have to add in the explainer? Perhaps not a big deal coz this is a testing file.

Advitya17 (approved):

imatiach-msft (approved):

responsibleai/tests/rai_insights/test_rai_insights_save_and_load_scenarios.py
romanlutz: Would also like to know :-)

gaugup: The new manager would be added anyways to the parameter list and the code to handle it would be added in the same PR, isn't it? It is dead code anyways.

PR #2355, opened 2023-09-22
Forecasting: generate time series as cohorts

romanlutz:
## Description

With raiwidgets we can inject time series as prebuilt cohorts, but in Studio we don't have that option, so this PR adds the functionality to generate time series as cohorts if they weren't provided. This only happens if time series ID features were provided, otherwise we assume that there is only a single time series in the dataset. 

While working on this I also uncovered a bug in the function `IsTimeOrTimeSeriesIdColumn`. Checking for datetime features doesn't quite make sense since we populate the `index` of the dataset with the time column that we get from Python. I removed that check since it actually broke the single time series case.

Overall, I also added two new test scenarios that capture the time series generation cases.

There are no UI changes.

## Checklist


libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/GenerateTimeSeriesCohorts.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export function generateTimeSeriesCohorts(
+  defaultErrorCohort: ErrorCohort,
+  preBuiltErrorCohortList: ErrorCohort[],
+  dataset: IDataset,
+  jointDataset: JointDataset
+): ErrorCohort[] {
+  if (
+    dataset.feature_metadata?.time_series_id_features &&
+    dataset.feature_metadata?.time_series_id_features?.length > 0
```
hawestra, line 24: mega nit: second dataset.feature_metatdata.time_series_id_features?.length (doesn't need the first ?)

libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/GenerateTimeSeriesCohorts.ts
```diff
@@ -0,0 +1,89 @@
...
+  ) {
+    // Need to generate time series as cohorts for forecasting.
+    const timeSeriesIdFeatureIndices =
+      dataset.feature_metadata?.time_series_id_features.map((featureName) =>
```
hawestra, line 28: remove ? again

hawestra (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/GenerateTimeSeriesCohorts.ts
```diff
@@ -0,0 +1,89 @@
...
+      );
+      distinctTimeSeriesIds.add(JSON.stringify(timeSeriesId));
+    });
+
+    preBuiltErrorCohortList = [];
+    distinctTimeSeriesIds.forEach((timeSeriesId) => {
+      const timeSeriesIdArray = JSON.parse(timeSeriesId);
+      const filterList: IFilter[] = [];
+      const cohortNameParts: string[] = [];
+      dataset.feature_metadata?.time_series_id_features?.forEach(
```
RubyZ10, line 44: remove the "?" ?

libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/GenerateTimeSeriesCohorts.ts
```diff
@@ -0,0 +1,89 @@
...
+  Cohort,
+  CohortSource,
+  IDataset
+} from "@responsible-ai/core-ui";
+
+import { translatePreBuiltCohortFilterForDataset } from "./ProcessPreBuiltCohort";
+
+export function generateTimeSeriesCohorts(
+  defaultErrorCohort: ErrorCohort,
+  preBuiltErrorCohortList: ErrorCohort[],
```
RubyZ10, line 18: Is this needed?

RubyZ10 (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Cohort/GenerateTimeSeriesCohorts.ts
romanlutz: The compiler expects the ? operator in this line (both occurrences). I suspect it's because we're inside the forEach (?)

PR #2358, opened 2023-09-25
Numerical Exif Metadata support for Image Dashboards

Advitya17:
This PR adds support for numerical exif metadata for all image dashboards.

## Description

image

image

image

image

image

image

image

## Checklist


libs/interpret-vision/src/lib/VisionExplanationDashboard/Controls/FlyoutObjectDetection.tsx
```diff
@@ -91,7 +91,7 @@ export class FlyoutObjectDetection extends React.Component<
           className={classNames.odFlyoutContainer}
         >
           <FluentUI.Stack tokens={FlyoutODUtils.stackTokens.medium}>
-            <FluentUI.Stack>
+            <FluentUI.Stack tokens={FlyoutODUtils.stackTokens.medium}>
```
imatiach-msft, line 94: nice use of constants!

responsibleai_vision/responsibleai_vision/utils/image_reader.py
```diff
@@ -90,9 +90,9 @@ def get_all_exif_feature_names(image_dataset):
                 for tag_id in exifdata:
                     # get the tag name, instead of human unreadable tag id
                     tag = TAGS.get(tag_id, tag_id)
-                    # TODO: add support for other data types
-                    if tag not in image_dataset.columns \
-                       and isinstance(exifdata.get(tag_id), str):
+                    if tag not in image_dataset.columns:
+                       data = exifdata.get(tag_id)
+                       if isinstance(data, str) or isinstance(data, int) or isinstance(data, float):
                         exif_feature_names.add(tag)
```
imatiach-msft, line 96: Can you fix the spacing here? it looks like there is a single space on line 96.  Does this actually work in python?
Oh, I guess python accepts 1, 2, or 4 spaces according to this:
https://stackoverflow.com/questions/2034517/pythons-interpretation-of-tabs-and-spaces-to-indent
Still I think we need the same number of spaces for indent.  Maybe we need to have some linting for this.

responsibleai_vision/responsibleai_vision/utils/feature_extractors.py
```diff
@@ -81,10 +81,11 @@ def extract_features(image_dataset: pd.DataFrame,
                     # decode bytes
                     if isinstance(data, bytes):
                         data = data.decode()
-                    # TODO: add support for other data types
                     if isinstance(data, str):
                         feature_metadata.categorical_features.append(str(tag))
                         row_feature_values[feature_names.index(tag)] = data
+                    elif isinstance(data, int) or isinstance(data, float):
+                        row_feature_values[feature_names.index(tag)] = data
```
imatiach-msft, line 88: how were you able to fix the json serialization error you were running into with the circular references?  or was that due to some other data type?

imatiach-msft (approved):

responsibleai_vision/responsibleai_vision/utils/image_reader.py
Advitya17: Yes plan to fix linting at the end after all technical comments are resolved :)

responsibleai_vision/responsibleai_vision/utils/feature_extractors.py
Advitya17: I'm so glad you asked, I honestly dk how this suddenly got working loolll

I could try to guess rebuilding, but believe I had already done that before and ran into the same error.

Another hypothesis can be, I perhaps removed unwanted data / references by strictly pegging metadata to str, int, or float.

PR #2362, opened 2023-09-26
Pytorch dataset

jamesbchao:
## Description
Jupyter notebook example of using a standard pytorch model and dataset with the RAI vision dashboard.

## Checklist


imatiach-msft:
looks like there are some test failures:
```Error: Some of the notebooks don't have tests. If a new notebook is added, Please add tests.```
I think you need to either add the notebook to the test automation or blacklist it - I'm guessing you could add it to regular notebook python tests but blacklist the e2e UI tests

PR #2363, opened 2023-09-26
Aggregate metrics

jamesbchao:
## Description
One-liner change to add back aggregate torchmetrics as per: https://msdata.visualstudio.com/Vienna/_queries/edit/2646440/?queryId=7b8f40a5-dea1-4516-a248-f93f2cc1ce66

## Checklist

### DIFF
--- a/notebooks/responsibleaidashboard/vision/responsibleaidashboard-standard-pytorch-model-dataset.ipynb
+++ b/notebooks/responsibleaidashboard/vision/responsibleaidashboard-standard-pytorch-model-dataset.ipynb
@@ -1,0 +1,104 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "c758ccdb",
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Files already downloaded and verified\n"
+     ]
+    },
+    {
+     "name": "stderr",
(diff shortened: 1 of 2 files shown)


imatiach-msft:
@jamesbchao we are waiting for next torchmetrics release before we can add that argument back, there's some story around this issue

jamesbchao:
> @jamesbchao we are waiting for next torchmetrics release before we can add that argument back, there's some story around this issue

Hi Ilya, are you referring to [torchmetrics v 1.2](https://lightning.ai/pages/community/announcements/torchmetrics-v1.2/)? I tried running the app with the new line and it seems to work.


Screenshot (14)

imatiach-msft:
@jameschao oh I see, indeed it was just released:
https://github.com/Lightning-AI/torchmetrics/releases

From release notes:

```
Update to Mean Average Precision
MeanAveragePrecision, the most widely used metric for object detection in computer vision, now supports two new arguments: average and backend.

The average argument controls averaging over multiple classes. By the core definition, the default way is macro averaging, where the metric is calculated for each class separately and then averaged together. This will continue to be the default in Torchmetrics, but now we also support the setting average="micro". Every object under this setting is essentially considered to be the same class, and the returned value is, therefore, calculated simultaneously over all objects.
```

imatiach-msft (approved):

imatiach-msft:
@jamesbchao seems it should be average, not aggregate, according to:

https://github.com/Lightning-AI/torchmetrics/issues/2004
https://github.com/Lightning-AI/torchmetrics/pull/2018

PR #2366, opened 2023-09-27
fix choose cohorts close or cancel updating feature cohorts selection in flyout

imatiach-msft:
## Description

Fixing bug from bug bash discovered by @vinuthakaranth 
The "Choose cohorts" flyout was saving the user's selection even when they clicked cancel or close buttons.
This issue was found during text/vision dashboard bugbash but it looks like a general issue that is in the tabular dashboard as well.

Video of functionality prior to fix:

https://github.com/microsoft/responsible-ai-toolbox/assets/24683184/5afe7832-69fe-43b6-aa03-da0ed53e7bc3

Video of functionality after fix:

https://github.com/microsoft/responsible-ai-toolbox/assets/24683184/f0f9b577-dedd-4ad7-ad54-71f0270a79d2

## Checklist

### DIFF
--- a/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyoutProps.tsx
+++ b/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyoutProps.tsx
@@ -1,0 +1,19 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { ErrorCohort } from "@responsible-ai/core-ui";
+
+export interface IChartConfigurationFlyoutProps {
+  isOpen: boolean;
+  onDismissFlyout: () => void;
+  datasetCohorts: ErrorCohort[];
+  featureBasedCohorts: ErrorCohort[];
+  selectedDatasetCohorts?: number[];
+  selectedFeatureBasedCohorts?: number[];
+  updateCohortSelection: (
+    selectedDatasetCohorts: number[],
+    selectedFeatureBasedCohorts: number[],
+    datasetCohortChartIsSelected: boolean
+  ) => void;
+  datasetCohortViewIsSelected: boolean;
+}
--- a/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyoutUtils.tsx
+++ b/libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyoutUtils.tsx
@@ -1,0 +1,88 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { IDropdownOption } from "@fluentui/react";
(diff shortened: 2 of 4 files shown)


libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyoutProps.tsx
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { ErrorCohort } from "@responsible-ai/core-ui";
+
+export interface IChartConfigurationFlyoutProps {
```
vinuthakaranth, line 6: nit: both props and state interface can be in single file

vinuthakaranth: or can be moved to utils file too

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyoutUtils.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+import { IDropdownOption } from "@fluentui/react";
+import { ErrorCohort } from "@responsible-ai/core-ui";
+
+import { IChartConfigurationFlyoutProps } from "./ChartConfigurationFlyoutProps";
+
+export const selectAllOptionKey = "selectAll";
+
+export function getInitialNewlySelectedDatasetCohorts(
+  props: IChartConfigurationFlyoutProps
```
vinuthakaranth, line 12: can we just pass the props that are needed to this function and the following function? this avoids the overhead of passing all unwanted props.

vinuthakaranth (approved):

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyoutProps.tsx
imatiach-msft: we have 24 files that are props only:
https://github.com/search?q=repo%3Amicrosoft%2Fresponsible-ai-toolbox+Props.ts&type=code
and we have 12 files that are state only:
https://github.com/search?q=repo%3Amicrosoft%2Fresponsible-ai-toolbox+State.ts&type=code
I hope it's ok to keep to this convention?  I just wanted to bring the line count down and this was the only thing I could think of.  Seems it needs deeper refactoring to bring it below the suggested limit of 300 lines.

libs/model-assessment/src/lib/ModelAssessmentDashboard/Controls/ModelOverview/ChartConfigurationFlyoutUtils.tsx
imatiach-msft: good point, done!

PR #2368, opened 2023-09-27
fix data analysis description and tooltips hidden for large screens

imatiach-msft:
## Description

Two fixes for tooltips:

1.) Tooltips seem to be getting hidden in larger screens and shown in smaller screens.  Removing "hideXxlUp" to allow tooltips to be shown.
Before:
image

After:
image

2.) Data analysis info tooltip description incorrect:

Before:
image

After:
image

## Checklist


libs/localization/src/lib/en.json
```diff
@@ -186,6 +186,12 @@
     "showTop": "Datapoints with the largest estimated causal responses to treatment feature: {0}",
     "whatifDescription": "What-if allows you to perturb features for any input and observe how the model's prediction changes. You can perturb features manually or specify the desired prediction (e.g., class label for a classifier) to see a list of closest data points to the original input that would lead to the desired prediction. Also known as prediction counterfactuals, you can use them for exploring the relationships learnt by the model; understanding important, necessary features for the model's predictions; or debug edge-cases for the model. To start, choose input points from the data table or scatter plot."
   },
+  "DataAnalysis": {
+    "TableView": {
+      "description": "View the dataset in a table representation with the ground truth and predicted labels.  The table is separated into correct and incorrect predictions.",
```
gaugup, line 191: The sentence "The table is separated into correct and incorrect predictions." could be reworded as "The table is separated into sub-tables of correct and incorrect predictions."

imatiach-msft: done!

gaugup (approved):

PR #2382, opened 2023-10-02
Unit test for Exif support in the Image Dashboards

Advitya17:
## Description

Added unit test for exif support on the Image Dashboards.

## Checklist

### DIFF
--- a/responsibleai_vision/tests/common_vision_utils.py
+++ b/responsibleai_vision/tests/common_vision_utils.py
@@ -289,7 +289,7 @@ def load_fridge_object_detection_dataset(automl_format=False):
     else:
         columns = [ImageColumns.IMAGE.value,
                    ImageColumns.LABEL.value]
-    feats = []
+    features = []
     for i, file in enumerate(os.listdir("./data/odFridgeObjects/" + "images")):
         image_path = "./data/odFridgeObjects/" + "images" + "/" + file
         if automl_format:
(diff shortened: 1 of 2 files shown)


responsibleai_vision/tests/common_vision_utils.py
```diff
@@ -283,12 +283,13 @@ def load_fridge_object_detection_dataset(automl_format=False):
...
-                                     ImageColumns.LABEL.value])
+        columns = [ImageColumns.IMAGE.value,
+                   ImageColumns.IMAGE_DETAILS.value,
+                   ImageColumns.LABEL.value]
     else:
-        data = pd.DataFrame(columns=[ImageColumns.IMAGE.value,
-                                     ImageColumns.LABEL.value])
+        columns = [ImageColumns.IMAGE.value,
+                   ImageColumns.LABEL.value]
+    feats = []
```
romanlutz, line 292: nit: can you spell it out? 😄

romanlutz (approved):

responsibleai_vision/tests/test_image_utils.py
```diff
@@ -86,3 +87,12 @@ def test_retry_sessions_retries_on_conn_failure(self, request_mock):
         session = image_reader_get_retry_session(url)
 
         assert session.get(url).status_code == 200
+
+    def test_get_all_exif_feature_names(self):
+        image_dataset = load_fridge_object_detection_dataset().head(2)
+        exif_feature_names = get_all_exif_feature_names(image_dataset)
+        assert len(exif_feature_names) == 11
```
gaugup, line 94: You could change this line to compare this with the length of the set in lines 95-98.

Advitya17: Length of the set or the list would be the same. Just casted the elements to a set for the assertion that the objects have the same elements (coz the order of `exif_feature_names` can change).

PR #2483, opened 2024-01-04
fix duplicate categorical features added from image exif metadata in RAI Vision Dashboard

imatiach-msft:
## Description

fix duplicate categorical features added from image exif metadata in RAI Vision Dashboard

Seeing duplicate categorical features in RAI VIsion Dashboard:
image

It looks like we are adding them for every row but we should only add them once.  Interestingly, this does not cause any errors currently as we are just checking if some feature name is in the list of categorical features in code, but it's definitely not desirable and can slow performance for larger image data.

This PR only adds the categorical feature names once to the list during exif feature metadata extraction and adds a test to validate this.

## Checklist


responsibleai_vision/responsibleai_vision/utils/feature_extractors.py
```diff
@@ -115,3 +85,37 @@ def extract_features(image_dataset: pd.DataFrame,
...
+                # get the tag name, instead of human unreadable tag id
+                tag = str(TAGS.get(tag_id, tag_id))
+                data = exifdata.get(tag_id)
+                # decode bytes
+                if isinstance(data, bytes):
+                    data = data.decode()
+                    if len(data) > MAX_CUSTOM_LEN:
+                        data = data[:MAX_CUSTOM_LEN] + '...'
+                if isinstance(data, str):
+                    if tag in feature_names:
```
Advitya17, line 106: For all conditions involving `in` check, I wonder if the collections objects on the right should be initially casted to sets instead of being lists to ensure O(1) lookup instead of O(n)?

Advitya17 (approved):

responsibleai_vision/responsibleai_vision/utils/feature_extractors.py
imatiach-msft: I was thinking of that as well - I guess we would need to do some performance testing with a change like that to validate it

PR #2514, opened 2024-01-26
Added metrics for genai text

kartik727:
Added AI-assisted metrics like coherence, fluency, etc. for generative text task types.


responsibleai_text/responsibleai_text/utils/genai_metrics/metrics.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+"""Compute AI-assisted metrics for generative text models."""
+
+from pathlib import Path
+
+import evaluate
```
imatiach-msft, line 8: can you wrap this in a try/catch similar to this one:
```
try:
    import evaluate
except ImportError:
    module_logger.debug(
        'Could not import evaluate, required if using a QA model')
```

imatiach-msft: https://github.com/microsoft/responsible-ai-toolbox/blob/main/responsibleai_text/responsibleai_text/rai_text_insights/rai_text_insights.py#L38

imatiach-msft: otherwise it can cause errors for other tasks which don't need it, specifically if __init__ files re-import functions for shorter namespaces

responsibleai_text/responsibleai_text/utils/genai_metrics/scripts/coherence.py
```diff
@@ -0,0 +1,103 @@
+# Copyright (c) Microsoft Corporation
+# Licensed under the MIT License.
+
+"""Coherence metric."""
+
+import datasets
```
imatiach-msft, line 6: similar thing for dataset here, I think we might want to wrap it in a try/catch

responsibleai_text/responsibleai_text/utils/genai_metrics/scripts/coherence.py
```diff
@@ -0,0 +1,103 @@
...
+import evaluate
+import pandas as pd
+
+logger = evaluate.logging.get_logger(__name__)
+
+
+_CITATION = """
```
imatiach-msft, line 13: can we move _CITATION and _KWARGS_DESCRIPTION  to a common constants file and re-use for all metrics?

responsibleai_text/responsibleai_text/utils/genai_metrics/scripts/coherence.py
```diff
@@ -0,0 +1,103 @@
...
+            inputs_description=_KWARGS_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "predictions": datasets.Value("string", id="sequence"),
+                    "references": datasets.Value("string", id="sequence")
+                }
+            ),
+        )
+
+    def _compute(self, *, predictions=None, references=None, **kwargs):
```
imatiach-msft, line 82: it seems that a lot of the code in _compute is repeated across metrics - maybe we can refactor it out into a common helper method?

responsibleai_text/responsibleai_text/utils/genai_metrics/scripts/coherence.py
```diff
@@ -0,0 +1,103 @@
...
+RATING:
+""".strip()
+
+
+@evaluate.utils.file_utils.add_start_docstrings(
+    _DESCRIPTION, _KWARGS_DESCRIPTION)
+class Coherence(evaluate.Metric):
+    def _info(self):
+
+        return evaluate.MetricInfo(
```
imatiach-msft, line 70: maybe this can also be implemented and re-used from some helper method in all metrics files?

imatiach-msft (approved):

responsibleai_text/responsibleai_text/utils/genai_metrics/scripts/coherence.py
kartik727: This is the standard interface that the `evaluate.Metric` class expects.

PR #2516, opened 2024-01-29
added generative text explainer

mohsinposts:
Added generative text explainer and it's respective in the reponsibleai_text setup.py. 

Note to be able to use this explainer fully kartik727's pull request "Added metrics for genai text #2514" needs to be approved since it includes code that extends the RAI toolbox to handle the model task type "generative_text"

### DIFF
--- a/responsibleai_text/responsibleai_text/managers/explainer_manager.py
+++ b/responsibleai_text/responsibleai_text/managers/explainer_manager.py
@@ -29,6 +29,24 @@ from responsibleai_text.common.constants import (ModelTask,
                                                  Tokens)
 from responsibleai_text.utils.question_answering import QAPredictor
 
+try:
+    from interpret_text.generative.lime_tools.explainers import (
+        LocalExplanationSentenceEmbedder
+    )
+except ImportError as e:
+    print("Could not import LocalExplanationSentenceEmbedder: ", e)
+
+try:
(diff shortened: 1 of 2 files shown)


responsibleai_text/responsibleai_text/managers/explainer_manager.py
```diff
@@ -29,6 +29,23 @@
...
+try:
+    from interpret_text.generative.lime_tools.explainers import \
+        LocalExplanationSentenceEmbedder
+except ImportError as e:
+    print("Could not import LocalExplanationSentenceEmbedder: ", e)
+
+try:
+    from interpret_text.generative.model_lib.openai_tooling import ChatOpenAI
+except ImportError as e:
+    print("Could not import ChatOpenAI: ", e)
```
imatiach-msft, line 41: maybe we can instead set chat_openai_installed = False and then throw before it is used in line 175, and similar for other imports?
Like the import here:
https://github.com/microsoft/ml-wrappers/blob/47babf7185664cb25b47409f912aa7e601213444/python/ml_wrappers/model/openai_wrapper.py#L12
And exception here:
https://github.com/microsoft/ml-wrappers/blob/47babf7185664cb25b47409f912aa7e601213444/python/ml_wrappers/model/openai_wrapper.py#L178

PR #2524, opened 2024-02-01
Add methods and constants for genai metrics 

kartik727:
Adding all the new methods and constants for the new generative text task types.

Full support for the new task type would require more changes but this PR is meant to add all the new methods without breaking the working of any of the existing task types.


raiwidgets/raiwidgets/responsibleai_dashboard_input.py
```diff
@@ -171,7 +171,7 @@ def _prepare_filtered_error_analysis_data(self, features, filters,
 
     def debug_ml(self, data):
         try:
-            features = data[0]
+            features = data[0]  # TODO: Remove prompt feature
```
imatiach-msft, line 174: I think comments need to be in-line to pass linting - but in this case I think we can even just remove this comment?

responsibleai_text/responsibleai_text/managers/error_analysis_manager.py
```diff
@@ -193,9 +203,17 @@ def __init__(self, model: Any, dataset: pd.DataFrame,
             task_type, index_classes)
         if categorical_features is None:
             categorical_features = []
+        if task_type == ModelTask.GENERATIVE_TEXT:
+            sup_task_type = ErrorAnalysisTask.REGRESSION
+            ext_dataset = ext_dataset.copy()
+            del ext_dataset['prompt']
+            ext_dataset['target_score'] = 5
```
imatiach-msft, line 210: nit: maybe update the order to be:
```
target_column = 'target_score'
ext_dataset[target_column] = 5
```

responsibleai_text/responsibleai_text/rai_text_insights/rai_text_insights.py
```diff
@@ -269,7 +280,9 @@ def _validate_model(self, model: Any, test: pd.DataFrame,
             target_column, axis=1)
         small_test_data = get_text_columns(small_test_data, text_column)
         small_test_data = small_test_data.iloc[0]
-        if task_type != ModelTask.QUESTION_ANSWERING:
+        if task_type not in [
```
imatiach-msft, line 283: nit: I think this might look a bit cleaner
```
list_task_outputs = [ModelTask.QUESTION_ANSWERING,
                                ModelTask.GENERATIVE_TEXT]
if task_type not in list_task_outputs:
  ...
```

responsibleai_text/responsibleai_text/utils/genai_metrics/scripts/equivalence.py
```diff
@@ -48,6 +48,15 @@
 
 This rating value should always be an integer between 1 and 5. So the rating \
 produced should be 1 or 2 or 3 or 4 or 5.
+Some examples of valid responses are:
```
imatiach-msft, line 51: this looks like the same repeating text in all of the prompts - maybe we can unify them by moving to a common constant in the common file and then importing and specifying that in all of these files?

imatiach-msft (approved):

raiwidgets/raiwidgets/responsibleai_dashboard_input.py
kartik727: There are more changes to be made in this method that are not a part of this PR. This is just a marker for making sure those are accounted for in the next PR. Will remove it then. Looks like this comment did pass linting so don't need to worry about that.

PR #2565, opened 2024-05-15
remove pins on werkzeug, itsdangerous and markupsafe since they are currently not needed

imatiach-msft:
## Description

Removing pins on werkzeug, itsdangerous and markupsafe since they are currently not needed.
Some of these were added about a year ago to fix build breaks:
https://github.com/microsoft/responsible-ai-toolbox/pull/1238
However these issues have been resolved and we should not be pinning them but instead rely on flask to pin to the correct versions unless we encounter issues again.
Otherwise, downstream customers are seeing problems with our pins being too restrictive.

## Checklist

### DIFF
--- a/rai_core_flask/requirements.txt
+++ b/rai_core_flask/requirements.txt
@@ -3,8 +3,5 @@ Flask-Cors
 flask-socketio
 ipython<=7.16.3; python_version <= '3.6'
 ipython>=7.31.1; python_version > '3.6'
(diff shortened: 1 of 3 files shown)


jeff-shepherd (approved):

tongyu-microsoft (approved):

imatiach-msft:
merging this for now as builds are broken due to unrelated reason, will send separate PR for broken builds (macos)

PR #2577, opened 2024-08-01
Upgrade flask to resolve notebook error " ImportError: cannot import name 'url_quote' from 'werkzeug.urls'"

tongyu-microsoft:
For ImportError: cannot import name 'url_quote' from 'werkzeug.urls', since we pinned werkzeug>=3.0.3 to fix the vulnerability issue on wekzeug, and the url_quote function was removed in Werkzeug version 3.0.0, we need to upgrade the Flask version to be compatible with the Werkzeug version 3.0.3.

## Checklist

### DIFF
--- a/rai_core_flask/requirements.txt
+++ b/rai_core_flask/requirements.txt
@@ -1,4 +1,4 @@
-Flask
+Flask>=3.0.0
 Flask-Cors
 flask-socketio
 ipython<=7.16.3; python_version <= '3.6'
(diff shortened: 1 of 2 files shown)


rai_core_flask/requirements.txt
```diff
@@ -1,4 +1,4 @@
-Flask
+Flask>=3.0.0
```
imatiach-msft, line 1: we should already be getting the latest flask version since there is no pin here - so the question is what is downgrading flask instead of getting the latest version?

imatiach-msft (approved):

qusongms (approved):

PR #2623, opened 2026-04-27
docs(notebooks): add encrypted fairness-metrics example using fairlearn-fhe

BAder82t:
Adds one notebook under `notebooks/individual-dashboards/fairness-dashboard/` showing how to compute Fairlearn metrics on encrypted predictions using [`fairlearn-fhe`](https://pypi.org/project/fairlearn-fhe/).

No changes to RAI core or the dashboards.

## What it does

1. Trains a logistic-regression baseline on UCI Adult.
2. Computes `demographic_parity_difference`, `equalized_odds_difference`, and `equal_opportunity_difference` in plaintext.
3. Computes the same three metrics on `encrypt(ctx, y_pred)` via `fairlearn_fhe.metrics`.
4. Asserts the two agree within `1e-3`.

## Verified locally

`jupyter nbconvert --to notebook --execute` runs end-to-end. Plaintext vs encrypted verdicts agree at `~1e-7` abs error.

## Notes

- New file only.
- `fairlearn-fhe` is Apache-2.0, requires `fairlearn>=0.10`, installable via `pip install fairlearn-fhe`.

### DIFF
--- a/notebooks/individual-dashboards/fairness-dashboard/fairness-dashboard-encrypted-metrics.ipynb
+++ b/notebooks/individual-dashboards/fairness-dashboard/fairness-dashboard-encrypted-metrics.ipynb
@@ -1,0 +1,274 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Encrypted fairness metrics with `fairlearn-fhe`\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


imatiach-msft (approved):

imatiach-msft:
this looks great, let me just run all of the tests and make sure it passes. I think you might want to improve the docs maybe somehow as well, like have an extra readme, but that could be a separate PR.

BAder82t:
Thanks for the review! 🙏

